PR #25, opened 2025-06-03
Setup nix

alexmatson-da:
Greatly simplified Nix setup from the CNU repo. We can extend this as needed when necessary (CI support, overlays for custom packages, etc...)

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,2 @@
+watch_dir nix
+use flake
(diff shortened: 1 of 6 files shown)


mjuchli-da:
We decided to postpone NIX for as long as differences in the local setup are not causing issues.

mjuchli-da (changes requested):
We decided to postpone NIX for as long as differences in the local setup are not causing issues.

mjuchli-da:
Closed for now, subject to reopening if needed in the future.

PR #27, opened 2025-06-03
boiler plate express server

rukmini-basu-da:


clients/remote/server/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "scripts": {
+    "build": "tsc",
+    "dev": "ts-node src/index.ts",
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "start": "node dist/index.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "type": "commonjs",
```
alexmatson-da, line 15: we should try to aim to make everything work with ES Modules (the new ecosystem standard) rather than CommonJS. I think if you change this to `"type": "module"` (and in the tsconfig), you should hopefully be able to build and run it successfully

alexmatson-da (approved):

PR #29, opened 2025-06-03
Workspaces

alexmatson-da:
This PR

- **Sets up NPM workspaces**

We now have a root `package.json` for handling the workspaces. There are a few installed dependencies and scripts, but these are intended to be used for repo management, not specific to any package. 

I decided to not rearrange the directory layout (yet). So each package is added to the workspace explicitly, in the `workspaces` key

----

- **Sets up Prettier for code formatting** 

Run `npm run format:fix` to apply the formatting to all files. The formatting rules are specified in the root-level `.prettierrc`

---

- **Installs Husky for integrating NPM scripts with git hooks** 

In order to run a Prettier check on every commit, I set up [Husky](https://typicode.github.io/husky/). The metamask repos use simple-git-hooks, but I found that it was overriding the `prepare-commit-msg` hook, and I couldn't get it to use a bash script. This means that new contributors shouldn't have to add the hook manually, hopefully


rukmini-basu-da (approved):
nice!

PR #30, opened 2025-06-03
setup example dapp

alexmatson-da:
The goal of this PR was to figure out how we might test the SDK on an actual dApp UI. In the long run we will begin integrating the SDK into the Utility, but it will be useful to have something in-repo for local dev purposes.

To that end:

- I added a example dApp project in `example/`, which uses a Vite+React+TS setup
- The example project installs `canton-wallet-sdk` as a dependency via the workspace 
- To demonstrate a working end-to-end bundle, the example dApp outputs a hello world message from the library
- I also started a root-level `tsconfig.base.json` file to reuse the same TS Configs across packages in the monorepo

Screenshot 2025-06-03 at 6 22 12 PM


mjuchli-da (approved):
Thanks! The example dApp certainly makes sense (as is planned within the Phase 1). 
However, I wonder whether or not it's worth demonstrating this via React or if VanillaJS + Web components would suffice here. That said, approved for now and subject to be discussed later.

PR #42, opened 2025-06-05
Implement a simple RPC generator

alexmatson-da:
The approach here:

- The `core/rpc-generator` package defines a custom template to be used by `open-rpc-generator` CLI
   - It generates two zero-dependency TS files that can be included directly in existing packages 🎉 
- The configuration for the build is defined in `dapp-rpc-generator-config.json` at the root, and executed with `yarn generate:dapp`
- The output is written directly into `clients/remote/src/dapp-api/rpc-gen` 
- The file `clients/remote/src/dapp-api/controller.ts` is a handwritten file implementing the RPC handlers
- The `POST /rpc` endpoint was added to the Express app to handle all JSON-RPC requests. I looked at libraries like [express-json-rpc-router](https://www.npmjs.com/package/express-json-rpc-router) but couldn't get them to work, so I wrote my own

Screenshot 2025-06-05 at 7 25 43 PM


alexgraham-da (approved):
Very nice! We should be able to extend this to handle the internal APIs (e.g. WK <-> keystore) as well.

mjuchli-da:
Nice one!

PR #48, opened 2025-06-10
add keystore api

alexgraham-da:
Adds an OpenRPC spec for the keystore connector API.

An addition I made to what was currently in the design doc is an `internalTxId` field for `signTransaction`. This is the ID that the Wallet Kernel will be tracking the transaction as - with multiple KeyStores, we can't be sure that the ID assigned by the KeyStore itself will be unique enough to be used as the WK's storage index.

The reason I have it passed in here is it may be useful for the KeyStore - Fireblocks for example has a field for `externalTxId`, and we may use the Wallet Kernel's ID in the Fireblocks `notes` field as a query parameter in a link to the visualizer.

I wasn't sure what to use for an event 'endpoint' (e.g. emits on a websocket), so I went with this:
```json
        {
            "name": "subscribeTransactions",
            "params": [
                {
                    "name": "params",
                    "title": "subscribeTransactionsParams",
                    "schema": {
                        "type": "object",
                        "properties": {
                            "txIds": {
                                "type": "array",
                                "description": "Unique identifiers of the transactions to subscribe to",
                                "items": {
                                    "type": "string"
                                }
                            }
                        },
                        "required": ["txIds"]
                    }
                }
            ],
            "result": {
                "name": "result",
                "title": "subscribeTransactionsResult",
                "schema": {
                    "ref": "#/components/schemas/Transaction"
                }
            }
        }
```

The `subscribe` indicates it is a subscription, and the `result` schema is the format the events will come in as.

### DIFF
--- a/api-specs/openrpc-keystore-connector-api.json
+++ b/api-specs/openrpc-keystore-connector-api.json
@@ -1,17 +1,276 @@
 {
     "openrpc": "1.2.6",
     "info": {
-        "title": "Wallet JSON-RPC Keystore API",
+        "title": "Canton JSON-RPC Wallet API",
         "version": "1.0.0",
-        "description": "An OpenRPC specification for the Keystore connector which allows the wallet kernel to interact with a Keystore."
+        "description": "An OpenRPC specification for interacting with a Canton wallet over JSON-RPC."
     },
-    "methods": [],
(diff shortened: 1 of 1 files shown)


mjuchli-da (approved):
Thanks!

PR #50, opened 2025-06-10
add dapp-client into core

rukmini-basu-da:
This generates the dapp client api into the core-modules. I'll add a follow up pr to use the sdk in the example dApp


core/wallet-dapp-rpc-client/build/index.d.ts
alexmatson-da, line 1: we can gitignore the `build` directory

alexmatson-da (approved):

PR #80, opened 2025-06-12
add client rpc generator

rukmini-basu-da:
Adds a custom rpc generator so the correct tsconfig.json is created when generating the clients.


alexmatson-da (approved):
nice!

PR #82, opened 2025-06-12
Wallet store

mjuchli-da:


core/wallet-store/src/Store.ts
```diff
@@ -0,0 +1,94 @@
...
+    auth: Auth
+}
+
+// Store interface for managing wallets, sessions, and networks
+
+export interface Store {
+    // Wallet methods
+    getWallets(): Promise<Array<Wallet>>
+    getPrimaryWallet(): Promise<Wallet | undefined>
+    setPrimaryWallet(partyId: PartyId): Promise<void>
```
PHOL-DA, line 75: we probably also need an `setPrimaryWallet(partyId: PartyId)`

core/wallet-store/src/Store.ts
```diff
@@ -0,0 +1,94 @@
...
+    addWallet(wallet: Wallet): Promise<void>
+
+    // Session methods
+    getSession(): Promise<Session | undefined>
+    setSession(session: Session): Promise<void>
+    removeSession(): Promise<void>
+
+    // Network methods
+    getNetwork(name: string): Promise<NetworkConfig>
+    getCurrentNetwork(): Promise<NetworkConfig>
```
PHOL-DA, line 85:
```suggestion
    getCurrentNetwork(): Promise<NetworkConfig>
    listNetworks(): Promise<Array<NetworkConfig>>
```

PHOL-DA (approved):
LGTM to me as an early draft, but would probably make sense to take a small session later on to validate if everything looks good compared to our expected cases. 

(We also have to figure out how to do with database migration)

core/wallet-store/src/Store.ts
```diff
@@ -0,0 +1,94 @@
...
+    hint: string
+    fingerprint: string
+    address: Address
+    chainId: string
+}
+
+// Session management
+
+export interface Session {
+    network: string
```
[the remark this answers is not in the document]
mjuchli-da: Yeah that was exactly the thinking that `NetworkConfig` would come with the whole tree... I think name makes more sense here.

core/wallet-store/src/Store.ts
mjuchli-da: Not sure I understand...

core/wallet-store/src/Store.ts
```diff
@@ -0,0 +1,94 @@
...
+    removeSession(): Promise<void>
+
+    // Network methods
+    getNetwork(name: string): Promise<NetworkConfig>
+    getCurrentNetwork(): Promise<NetworkConfig>
+    updateNetwork(network: NetworkConfig): Promise<void>
+    removeNetwork(name: string): Promise<void>
+}
+
+export interface AccountStore {
```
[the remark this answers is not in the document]
mjuchli-da: I was debating with myself about this yesterday – and I think `AccountStore` is probably unnecessary. The thing is that we shouldn't make the `Store` user-specific either, but keep it flexible, so instead of
```
...
getWallets(account: Account)
...
```
we rather have
```
...
getWallets()
...
```
Then it's down to the implementation to support multi-tenancy (by for example using an authService) or not.

core/wallet-store/src/Store.ts
PHOL-DA: If you want to change your Primary Wallet ? Or do we just default to the first one you create ?

mjuchli-da: Yeah but the exact same method is already there that you mentioned :)

PHOL-DA: I must be going blind :laughing:

PR #83, opened 2025-06-12
add templates folder client

rukmini-basu-da:
Add templates folder for the custom rpc client. This was being ignored by .gitignore and should've been added [here](https://github.com/hyperledger-labs/splice-wallet-kernel/pull/80)


alexmatson-da (approved):
thanks!

PR #84, opened 2025-06-12
 use react for discovery popup window 

alexmatson-da:


PHOL-DA (approved):

mjuchli-da:
I'm not sure if we should use React for the SDK components or go with a less opinionated Web Components approach. If the community wants to extend or build their SDK from scratch by reusing some of the components we provide, this might become less attractive.

alexmatson-da:
Closing in favor of #89

PR #88, opened 2025-06-16
clients/remote: add zod schemas for wallet store interfaces

rukmini-basu-da:
This PR adds the zod library to `clients/remote` so we can read configs from a file that map to the `wallet-store` interfaces. The initial use case is for the `listNetworks` call in the `connect` flow where the client will supply a config file with the networks/idps.


clients/remote/src/config/StoreConfig.ts
```diff
@@ -0,0 +1,76 @@
+import * as store from '../../../../core/wallet-store/dist/Store.js'
+// import * as store from 'core-wallet-store'
+import { z, ZodType, ZodObject } from 'zod'
+
+//TODO: fix the core-wallet-store import
```
alexmatson-da, line 5: what was the import issue?

alexmatson-da (approved):
nice!

PR #92, opened 2025-06-17
Internal Wallet Store

mjuchli-da:


core/wallet-store/src/StoreInternal.ts
```diff
@@ -0,0 +1,158 @@
+import { Store, Wallet, PartyId, Session, NetworkConfig } from './Store.js'
+
+type UserId = string
+
+interface UserService {
```
mjuchli-da, line 5: To be extracted into separate package, e.g. `wallet-auth`

alexmatson-da (approved):

PR #102, opened 2025-06-22
User RPC Client

mjuchli-da:
To be used in remote clients web project


core/wallet-user-rpc-client/.gitignore
```diff
@@ -0,0 +1 @@
+docs
```
PHOL-DA, line 1: what is this doing ? there is no "docs" in at this place in the folder structure

core/wallet-user-rpc-client/eslint.config.mjs
```diff
@@ -0,0 +1,67 @@
...
+import prettierConfig from 'eslint-config-prettier'
+import prettierPlugin from 'eslint-plugin-prettier'
+import { fileURLToPath } from 'url'
+import { dirname } from 'path'
+import globals from 'globals'
+
+const __filename = fileURLToPath(import.meta.url)
+const __dirname = dirname(__filename)
+
+export default [
```
PHOL-DA, line 13: now this is almost the exact same as in `wallet-dapp-rpc-client`, could we maybe combine it into one instead?

ecosystem.config.js
```diff
@@ -1,7 +1,7 @@
 export const apps = [
     {
         name: 'remote',
-        script: 'yarn workspace client-remote dev',
+        script: 'yarn workspace clients-remote start',
```
PHOL-DA, line 4: why do we want to change this from `dev` to `start`? And why dont't we want to change this for all the other workspaces?

PHOL-DA (approved):
A little cleanup requested but otherwise LGTM

ecosystem.config.js
mjuchli-da: I'll revert to `dev` (I had to use `start` as dev kept restarting and I think we need to exclude something from the watch -- but that's a separate issue)

PR #108, opened 2025-06-24
redirect to discovered wk

rukmini-basu-da:
In the sdk.connect() function, this will open up a new walletKernelWindow with the DiscoverResult, which currently is localhost:3002

https://github.com/user-attachments/assets/ca9acd3b-77fa-4ec2-bee4-c4a8d3700417

### DIFF
--- a/sdk/src/connect/index.ts
+++ b/sdk/src/connect/index.ts
@@ -21,11 +21,20 @@ export type ConnectError = {
 }
 
 export async function connect(): Promise<ConnectResult> {
+    const walletKernelWindow = window.open('', '_blank')
+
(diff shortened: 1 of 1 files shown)


mjuchli-da (changes requested):
I have a few additions and I think it makes sense to work this into this PR:
- We should reuse the `popup.js`
- We should use the dApp API to make the actual `connect` request against the dApp API with `baseUrl = result.url`
- Implement a handler to extract the `userUrl` returned by the dApp API response
- This `userUrl` is what we should show in the popup

I will try to add these ideas throughout my morning and so we can sync tomorrow.

PR #116, opened 2025-06-26
Add a common types package w/ Zod parsers

alexmatson-da:
Changes:

- `core-types` is a package that contains common utility types used across multiple packages. This includes wallet messages, window events, discover results, etc. It also includes zod parsers, as typically the receiving handler only receives an `unknown` object, and the format must be validated/parsed 

- the browser extension backend now fully acts as a JSON-RPC service, verifying that request messages adhere to the JSON-RPC spec, and it responds with JSON-RPC responses

- a while back I used `arktype` in the Remote WK backend to parse unknown JSON objects. Since we ended up using `Zod` for the config schemas, I aligned on Zod for those as well (and removed `arktype`)


rukmini-basu-da (approved):
nice!

PR #117, opened 2025-06-26
login flow

rukmini-basu-da:
This adds the network login page and the login callback, which extracts the access_token. In a follow up PR, I'll use the user-api to listNetworks rather than serving a static config, but wanted to get the basic functionality merged first. 

I tested by running a local dex instance  as the IDP provider with the following config:

```
# The base path of dex and the external name of the OpenID Connect service.
# This is the canonical URL that all clients MUST use to refer to dex. If a
# path is provided, dex's HTTP service will listen at a non-root URL.
issuer: http://127.0.0.1:5556/dex

# The storage configuration determines where dex stores its state. Supported
# options include SQL flavors and Kubernetes third party resources.
#
# See the documentation (https://dexidp.io/docs/storage/) for further information.
storage:
  type: memory

# Configuration for the HTTP endpoints.
web:
  http: 0.0.0.0:5556
  allowedOrigins: ["*"]


# Default values shown below
oauth2:
  passwordConnector: local
  grantTypes:
    - "password"
    - "authorization_code"
    - "refresh_token"
    - "urn:ietf:params:oauth:grant-type:device_code"
    - "urn:ietf:params:oauth:grant-type:token-exchange"
  responseTypes: ["code"]
  skipApprovalScreen: true

# Instead of reading from an external storage, use this list of clients.
# If this option isn't chosen clients may be added through the gRPC API.
staticClients:
  - id: "test-client"
    redirectURIs:
      - "http://localhost:3002/callback"
      - "http://localhost:8002"
    name: "WK UI"
    public: true
    secret: "service-account-secret"

# Let dex keep a list of passwords which can be used to login to dex.
enablePasswordDB: true

# to get password hash:
# echo "password" | htpasswd -BinC 10 admin | cut -d: -f2
staticPasswords:
  - email: "ellis@dex.com"
    hash: "$2y$10$earQUmjoX7cqzo.QSBaqXeavW8Y8dpzHbt34h4odEsycBQAhZP4sy"
    username: testuser
    userID: "ellis"
    # sub: CgVlbGxpcxIFbG9jYWw

oauth2:
  responseTypes: ["token", "code", "id_token"]
```

https://github.com/user-attachments/assets/4284cff3-8dab-48a6-a61d-855e90d14cc5


alexmatson-da (approved):
nice!

Maybe time to start thinking if we want repo tooling around Dex or maybe find some NodeJS based alternative like https://www.npmjs.com/package/oauth2-mock-server?activeTab=readme

PR #121, opened 2025-06-30
Add scripts to download and run a Canton participant

alexmatson-da:


rukmini-basu-da (approved):

scripts/src/fetch-canton.ts
```diff
@@ -0,0 +1,75 @@
...
+import fs from 'fs'
+import path from 'path'
+import zlib from 'zlib'
+import tar from 'tar-fs'
+import { pipeline } from 'stream/promises'
+
+const SHA256 =
+    '5f1bf64d5d3bf50c4dd379bca44d46069e6ece43377177a6e09b4ff0979f640d'
+
+async function downloadCantonTarGz(tarfile: string): Promise<void> {
```
PHOL-DA, line 24: i have generally tried to keep the scripts synchronically by default

scripts/src/fetch-canton.ts
```diff
@@ -0,0 +1,75 @@
...
+    success,
+} from './utils.js'
+import crypto from 'crypto'
+import fs from 'fs'
+import path from 'path'
+import zlib from 'zlib'
+import tar from 'tar-fs'
+import { pipeline } from 'stream/promises'
+
+const SHA256 =
```
PHOL-DA, line 21: i would put this one next to the `CANTON_VERSION`

scripts/src/utils.ts
```diff
@@ -1,10 +1,33 @@
...
+export const CANTON_CONF = path.join(getRepoRoot(), 'canton.conf')
+
+// Canton versions
+export const DAML_RELEASE_VERSION = '3.4.0-snapshot.20250625.0'
+export const CANTON_VERSION = '3.4.0-snapshot.20250617.16217.0.vbdf62919'
 
 // Get the root of the current repository
 export function getRepoRoot(): string {
-    return process.cwd()
+    const cwd = process.cwd()
```
PHOL-DA, line 16: i think the  changes made to this method returns a folder path 1 layer higher than before (which might be a problem for the clean coding script as it derives based on folder and subfolder pathing).

scripts/src/start-canton.ts
```diff
@@ -0,0 +1,45 @@
+/**
+ * Start a local Canton participant using the config file in the canton/ directory.
+ */
+
+import { CANTON_BIN, CANTON_CONF, error, info } from './utils.js'
+import { existsSync } from 'fs'
+import pm2 from 'pm2'
+
+async function main() {
```
PHOL-DA, line 9: are we sure we want to do it like this instead of using something like docker ?

mjuchli-da: It would actually be my preferred option to run it as a system process, as suggested above—CN is doing the same, and it works very smoothly. Running it in a Docker container could just be an extra option, no?

scripts/src/fetch-canton.ts
alexmatson-da: once you start dealing with network requests and such you're in async land whether you like it or not... only difference without the explicit `async` keyword is using `.then()` Promise syntax instead of `await ...`

scripts/src/start-canton.ts
PHOL-DA: if it requires artifcatory then it is a no go for now :)

scripts/src/fetch-canton.ts
alexmatson-da: yep! did a test and it seems to wait for all pending async ops to complete before exiting

PHOL-DA (approved):

PR #122, opened 2025-07-01
Fireblocks single user implementation

alexgraham-da:
Implements the "single API user" version of the Fireblocks signing driver.

There's a few TODOs in here:
1. Dynamic reconfiguration
2. Event streaming
3. Ensure the driver is signing a valid Canton transaction
4. Add server version of controller (if we are doing this)

Retrieving the API Key and secret key is left to the consumer of the driver/controller (it just takes a configuration that has the keys as strings).

There is a bit of further optimization that can be done (mainly caching) since the Fireblocks API isn't exactly fast and always returns all transactions.


core/signing/fireblocks/README.md
```diff
@@ -0,0 +1,3 @@
+# rpc-generator
```
mjuchli-da, line 1: I'd change the title here to "Fireblocks Signing Driver" or similar. 
(Ideally, it would come with a bit of description in a format whose structure we reuse for future signing connectors.)

core/signing/fireblocks/src/controller.ts
```diff
@@ -0,0 +1,153 @@
...
+} from 'core-signing-lib'
+import { FireblocksHandler } from './fireblocks'
+
+interface FireblocksConfig {
+    apiKey: string
+    apiSecret: string
+    apiPath?: string
+}
+
+export class FireblocksDriver implements SignerInterface {
```
mjuchli-da, line 30: I think the name `FireblocksDriver` is most accurate. However, some discussion led to changing the core package from `driver` to `signing`, and I presume that resulted in `SignerInterface` (I can't seem to find out what the reason was). If we stick to `FireblocksDriver`, can we change `SignerInterface` to `Driver` (or `SigningDriver`) just?

core/signing/fireblocks/src/fireblocks.ts
```diff
@@ -0,0 +1,268 @@
+import {
```
mjuchli-da, line 1: Could this be just `index.ts` so that imports against this package would provide access to the types here?

core/signing/fireblocks/src/fireblocks.ts
```diff
@@ -0,0 +1,268 @@
...
+                        )
+                        this.keyCacheByDerivationPath.set(
+                            JSON.stringify(key.data.derivationPath),
+                            storedKey
+                        )
+                    }
+                }
+            }
+        } catch (error) {
+            console.error('Error fetching vault accounts:', error)
```
mjuchli-da, line 107: It might be better to use `pino` logging here

core/signing/fireblocks/src/fireblocks.ts
```diff
@@ -0,0 +1,268 @@
...
+     * @param tx - The transaction to sign, as a string
+     * @param publicKey - The public key to use for signing
+     * @return The transaction object from Fireblocks
+     */
+    public async signTransaction(
+        tx: string,
+        publicKey: string
+    ): Promise<FireblocksTransaction> {
+        try {
+            if (!this.keyCacheByPublicKey.has(publicKey)) {
```
mjuchli-da, line 215: Not an issue for now but we should think of what happens if a user removes a given pubkey in fireblocks and we still have it cached...

core/signing/fireblocks/src/controller.test.ts
```diff
@@ -0,0 +1,112 @@
...
+        name: keyName,
+        publicKey: TEST_FIREBLOCKS_PUBLIC_KEY,
+    }
+    return {
+        signer,
+        key,
+    }
+}
+
+describe.skip('fireblocks controller (skip due to API secret requirement)', () => {
```
mjuchli-da, line 63: I'm inclined to have a mocked `FireblocksHandler` injected into the `FireblocksDriver`. This way, we would at least be able to locally test end-to-end by assuming certain rpc behaviour. The fireblocks.test.ts would catch rpc changes. Wdyt?

mjuchli-da (dismissed):
# Naming

Sorry for being pedantic with naming: there seems to be a bit of a mix between `signer` and `driver` wording. I personally prefer `driver` since this generalises the concept of a store of some kind (signing keys) and the interface would then express it's capabilities to conduct `signing`. 

# Package/Folder structure

I'd move `signing/internal` and `signing/fireblocks` to core top-level: e.g. 
- `core/signing-lib`
- `core/signing-internal`
- `core/signing-fireblocks`

This enforces clear exposure and usage of lib methods and seems more flexible when it comes to adding or removing drivers.

core/signing/fireblocks/src/controller.test.ts
alexgraham-da: Added in this PR

PHOL-DA (approved):
LGTM, added a few suggestion or comments along the way

core/signing-fireblocks/src/fireblocks.ts
```diff
@@ -0,0 +1,271 @@
...
+
+            for (const vault of vaultAccounts) {
+                if (vault.id) {
+                    const derivationPath = `[44, ${CC_COIN_TYPE}, ${vault.id}, 0, 0]`
+                    const key = await this.client.vaults.getPublicKeyInfo({
+                        algorithm: 'MPC_EDDSA_ED25519',
+                        derivationPath,
+                    })
+                    if (
+                        key.data.publicKey &&
```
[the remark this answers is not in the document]
alexgraham-da: It will always be defined

core/signing-fireblocks/src/fireblocks.ts
```diff
@@ -0,0 +1,271 @@
...
+     * refresh the key cache.
+     * @returns AsyncGenerator of FireblocksTransactions
+     */
+    public async *getTransactions(): AsyncGenerator<FireblocksTransaction> {
+        let refreshedCache = false
+        try {
+            const transactions = await this.client.transactions.getTransactions(
+                {
+                    sourceType: 'VAULT_ACCOUNT',
+                    limit: 500,
```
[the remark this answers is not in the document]
alexgraham-da: They don't have any regular pagination implemented unfortunately. I don't think they generally want you using this endpoint (they would prefer you to use a webhook).

The endpoint is ordered descending by date, so you will always be getting the latest transactions. Depending on the volume of transactions, if you don't do 500 more transactions before you sign and submit, you would be fine.

They do provide `before` and `after` filters, so we could implement pagination w/ those. Since it's a generator we wouldn't have to worry about it fetching all transactions from the beginning of time as well. I'd prefer it be a followup to this PR though.

PR #124, opened 2025-07-01
Implement allocateParty RPC method for user API

alexmatson-da:
Showcase of how we can implement an RPC method that talks to the Canton ledger API, by implementing `allocateParty` for the remote WK

https://github.com/user-attachments/assets/9b3221e1-8008-4ace-8b6a-44828fbc269e


rukmini-basu-da (approved):

mjuchli-da (approved):
Thx! 
We obviously have to change the implementation (internal vs external party) with the integration of the signing providers. Besides, the endpoint will change according to what was described here: https://github.com/hyperledger-labs/splice-wallet-kernel/issues/125. I'll do the latter in a follow-up PR.

Issue #131, opened 2025-07-02
Refactor SpliceProvider


alexmatson-da:
With the changes to the client codegen, it is now feasible to reuse the dApp API client directly within the SpliceProvider, with its improved type checking for request methods & params. The way this might look:

- We define a package `rpc-transport` which exports various classes implementing the `RpcTransport` interface for sending RPC requests
    - `HttpTransport`
    - `WindowTransport`
    - ... 
- The generated clients accept an instance of a transport in their constructor (alternatively, arguments to instantiate a transport internally?)
- The SpliceProvider instantiates a dApp API client internally based on configured transport option (to support provider injection from either the extension or the remote service)

This cleanly separates the logic between protocol transports, RPC requests, and the dApp SDK experience

PR #132, opened 2025-07-02
Tweak client codegen for improved request typing

alexmatson-da:
Also removes the implicit dependencies on Node/EventEmitter so that the clients are usable in a browser environment 🎉 

Eventually we can use the clients within the SpliceProvider and unify our transport logic, but that is captured in #131


mjuchli-da:
Nice! I'd like to review later today or tomorrow.

mjuchli-da (approved):
Cool!

PR #136, opened 2025-07-03
add signing provider id to wallet store object

rukmini-basu-da:
Updates to the wallet store interface to introduce the `SigningProvider` and `SigningDriver` (fields still tbd), adds `NetworkId` to the `NetworkConfig` and a `WalletFilter` so we can filter the wallets by `networkIds` or `signingProviderIds`


mjuchli-da (approved):
Thanks!

PR #137, opened 2025-07-03
implement prepareReturn method

alexmatson-da:
current status:

- A new button on the example dApp can be clicked to submit a Ledger command that creates a `Ping` contract (part of the built-in `AdminWorkflows` DAR uploaded to all participants)
- The request is routed through the SpliceProvider to the dApp API
- improve some RPC error handling & display
- authenticates SpliceProviderHttp with the token received after login


mjuchli-da (approved):
Nice!!

PR #138, opened 2025-07-07
add pagination to fireblocks getTransactions

alexgraham-da:
Adds pagination to the Fireblocks Driver's `getTransaction` function using `createdAt` timestamps.


core/signing-fireblocks/src/fireblocks.test.ts
```diff
@@ -17,23 +17,29 @@ describe('fireblocks handler', () => {
         // in index.test.ts
         test.skip('FIREBLOCKS_API_KEY environment variable is not set, skipping test.', () => {})
     } else {
-        test('key creation', async () => {
```
PHOL-DA, line 20: you dont want to keep the test "key creation" or is it because it is only setup ?

PHOL-DA (approved):
LGTM

PR #141, opened 2025-07-07
Signing Participant Driver

mjuchli-da:


core/signing-participant/src/controller.ts
```diff
@@ -0,0 +1,53 @@
...
+    SignTransactionParams,
+    SignTransactionResult,
+    SubscribeTransactionsResult,
+} from 'core-signing-lib'
+import { randomUUID } from 'node:crypto'
+
+export class ParticipantSigningDriver implements SigningDriverInterface {
+    public partyMode = PartyMode.INTERNAL
+    public signingProvider = SigningProvider.PARTICIPANT
+    public controller = buildController({
```
mjuchli-da, line 21: I'm unsure about the controller. Given the properties above, a client will know that if a command is to be submitted via the wallet whose driver is `PartyMode.INTERNAL`, it does not require signing. 

Should we leave all methods unimplemented?

PHOL-DA (approved):
LGTM

PR #142, opened 2025-07-08
Open RPC Schema Titles

PHOL-DA:
In order to have nicely code-generated classes based on the open-rpc specs all fields, properties, methods etc. should have a title defined (this changes exported type variable generation from `String3T7JhIFf` to `Url`)


mjuchli-da (approved):
Nice one!

PR #145, opened 2025-07-08
Add socket.io connection to Splice HTTP Provider for SDK

alexmatson-da:
This enables event updates over the SpliceProvider by adding support for a "Notifier" on the store. The store implementation is responsible for emitting notifications over the this interface, allowing WK implementations to subscribe to various events. 

Splitting it up this way was motivated by a few factors:

- The store is the single source of truth; we want to be notified on store updates
- Updates can be initiated by one API server/controller (the User API), but consumed by another (the dApp API)
- Store implementations should remain generic w.r.t to the WK runtime. For the Remote WK, we use a NodeJS EventEmitter. For the browser WK, we'd use browser event listeners. 
- Socket.io servers and Express servers can be combined to run on a single HTTP server on the same port. However, accessing the socket IO interface from within the RPC middleware is non-trivial -- the Notifier approach solves this problem. 

https://github.com/user-attachments/assets/8470ca8f-36ba-4e7b-a08f-05f5de11654f


core/splice-provider/src/SpliceProviderHttp.ts
```diff
@@ -1,12 +1,18 @@
 import { isSpliceMessageEvent, RequestPayload, WalletEvent } from 'core-types'
 import { SpliceProviderBase } from './SpliceProvider'
+import { io, Socket } from 'socket.io-client'
 
 export class SpliceProviderHttp extends SpliceProviderBase {
     private sessionToken?: string
+    private socket: Socket
```
PHOL-DA, line 7: Maybe we should rename the class and file to `SpliceProviderWS` (since it is WebSocket and not Http anymore?)

PHOL-DA (approved):

mjuchli-da (approved):

mjuchli-da (commented):
As discussed, there should be a separate `NotificationService` which allows to `emit` events per user.

clients/remote/src/middleware/jwtAuth.ts
```diff
@@ -7,10 +7,6 @@ export function jwtAuth(authService: AuthService, logger: Logger) {
     return async (req: Request, res: Response, next: NextFunction) => {
         const authHeader = req.headers.authorization
 
-        if (!authHeader || !authHeader.startsWith('Bearer ')) {
```
mjuchli-da, line 10: @alexmatson-da what is the reasoning behind the removal of this?
If I remember correctly, I specifically made auth optional this way since some of the endpoints could be "public"

alexmatson-da: not removed, but moved it inside the auth service to reuse it (for socket auth)

PR #146, opened 2025-07-08
network management ui

rukmini-basu-da:
Related to https://github.com/hyperledger-labs/splice-wallet-kernel/issues/149

In a following PR, I'll add the update network, but this takes care of listing, adding, and deleting networks. Additionally, added the rpc addNetworks call to the wallet-store.

https://github.com/user-attachments/assets/559ae0d4-c45a-4d40-ae19-2d79ec6adb4f


clients/remote/src/web/frontend/networks/index.ts
```diff
@@ -0,0 +1,392 @@
...
+import { customElement, state } from 'lit/decorators.js'
+import { userClient } from '../rpc-client'
+import {
+    AddNetworkParams,
+    Network,
+    RemoveNetworkParams,
+} from 'core-wallet-user-rpc-client'
+
+@customElement('user-ui-networks')
+export class UserUiNetworks extends LitElement {
```
mjuchli-da, line 12: I'd move it to `wallet-ui-components` windows as `AppNetworks`

rukmini-basu-da: Sounds good, I've moved the `NetworkForm` and `NetworkTable` to the components. However, I left some logic in the `networks` component as we use the `userClient` for the handleSubmit/handleDelete functions

mjuchli-da (approved):
Thanks!
Layout-wise we should make improvements in a subsequent PR as this is unfortunately not compatible with the popup :( 
image

PR #147, opened 2025-07-09
canton bootstrapping and console output

PHOL-DA:
this pr:

- Changes naming convention for synchronizer from `da` to `wallet`
- Attaches a bus to pm2 when running `yarn start:canton` that logs console output (so potential debugging becomes easier)


alexmatson-da (approved):
very nice!

PR #153, opened 2025-07-10
Implement basic ledger write flow w/ internal party

alexmatson-da:
(apologies for the large diff, I moved the codegen file for the Ledger API which counted as +-4k lines) 

This implements a basic interactive flow for submitting internal ledger party commands. 

workflow:

1. the dApp sends `prepareExecute` request with a set of ledger `JsCommands` 
2. the backend generates a UUID to use for the `commandId`. This is used to identify the same transaction across the flow (at this stage, we do not have an actual `transactionId` to use)
3. the backend prepares the transaction (for external signing), then stores the pending transaction in the store
4. the backend returns a URL to the WK user UI for user approval
5. the `SpliceProvider` library automatically generates a popup window to the approval page, with `commandId` supplied as a query param
6. the approval page reads the commandId off the query param, and (since this belongs to an internal party), directly submits the transaction to the ledger API

# Demo!

https://github.com/user-attachments/assets/a7485074-6086-4c91-8122-6e8ac1c9c6da

And we can see the contract actually is created in the participant's ACS:

Screenshot 2025-07-10 at 6 44 19 PM


mjuchli-da (approved):
Thanks. There is a bit of error handling missing but we can tackle it in a follow-up

PR #156, opened 2025-07-15
Store synchronizer ID alongside network model

alexmatson-da:


core/splice-provider/src/SpliceProviderHttp.ts
```diff
@@ -76,31 +81,38 @@ export class SpliceProviderHttp extends SpliceProviderBase {
             }),
         })
 
-        const body = await res.json()
+        const body = await res.json().then(JsonRpcResponse.parse)
+
+        if ('error' in body) throw new Error(body.error.message)
 
         if (method === 'prepareExecute') {
```
mjuchli-da, line 88: Note for a future task: I think we should make the fact that some of the endpoints are "async" (i.e. resulting in a popup) a generic concept at some point.

mjuchli-da (approved):
Thanks!

PR #162, opened 2025-07-16
modify the canton bootstrap script to reuse crypto keys

rukmini-basu-da:
This modifies the canton bootstrap script by reusing the crypto keys and manually initializing the participant. The motivation of this is that when we enable auth on the participant, we need to add the participantId in audience claim of our jwt token, so this will enable us to have the same participantId each time we start canton.


alexmatson-da (approved):
👏

PR #166, opened 2025-07-17
add token to ledger client

rukmini-basu-da:
This adds a jwt token to the `LedgerClient`. In the pr where I enable auth on the participant, I'll clean up some of the hardcoded values (probably will be config driven)


alexmatson-da (approved):
thanks!

PR #171, opened 2025-07-22
fix jwks server url

rukmini-basu-da:
When creating the authenticated participant, we updated the jwks url for the mock-oauth2 server to `http://127.0.0.1:8889/jwks`. This updates the auth service to use the new url.


alexmatson-da (approved):
thx!

PR #176, opened 2025-07-23
refactor ledger client instantiation  

rukmini-basu-da:
Refactors the ledger client to be instantiated per request and to use the access token from the authContext of the request and the baseUrl from the network configuration based off of the currentSession.

### DIFF
--- a/clients/remote/src/index.ts
+++ b/clients/remote/src/index.ts
@@ -7,42 +7,15 @@ import { AuthService } from 'core-wallet-auth'
 import { StoreInternal } from 'core-wallet-store'
 import { ConfigUtils } from './config/ConfigUtils.js'
 import { configSchema } from './config/Config.js'
-import { LedgerClient } from 'core-ledger-client'
 import { Notifier } from './notification/NotificationService.js'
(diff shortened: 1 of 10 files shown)


clients/remote/src/index.ts
```diff
@@ -7,43 +7,16 @@ import { AuthService } from 'core-wallet-auth'
 import { StoreInternal } from 'core-wallet-store'
 import { ConfigUtils } from './config/ConfigUtils.js'
 import { configSchema } from './config/Config.js'
-import { LedgerClient } from 'core-ledger-client'
 import { Notifier } from './notification/NotificationService.js'
 import EventEmitter from 'events'
 import { createRemoteJWKSet, jwtVerify } from 'jose'
-import axios from 'axios'
-import qs from 'qs'
```
alexmatson-da, line 15: lets remove from package.json too and keep the dependency list clean

alexmatson-da (approved):

clients/remote/src/index.ts
rukmini-basu-da: sounds good, removed!

PR #181, opened 2025-07-28
Multi user fireblocks

alexgraham-da:
Implements API Key per user for the `FireblocksSigningDriver`. The driver also has the option of including a default, so this can be used both for single-API Key setups or optionally to allow a default key to be used if a user does not have their own.  

Note that I do allow a `SigningDriver` to be used without an `AuthContext` (though this would only work for Fireblocks if a default key is included). This would allow the wallet kernel to do automation such as internally polling for transactions to submit. We can change it to always require an `AuthContext` if that's not something that will be necessary.

The `FireblocksSigningDriver` and other signing drivers are singletons, so to get a controller for the particular user, you pass it an `AuthContext`:

```typescript
const driverController = signingDriver.controller(authContext)
const keys = await driverController.getKeys()
```

You can see usage of this in the Jest tests, but I can add some documentation in the READMEs as well.


core/signing-fireblocks/package.json
```diff
@@ -10,7 +10,8 @@
     "dependencies": {
```
PHOL-DA, line 10: even though it is not part of the PR can you add `"type": "module"` to this pr. (as the check confirms is required) ?

core/signing-fireblocks/src/index.test.ts
```diff
@@ -77,43 +121,71 @@ export function throwWhenRpcError<T>(value: T | RpcError): void {
...
+        noDefaultSigningDriver,
         key,
+        controller: signingDriver.controller(TEST_AUTH_CONTEXT),
     }
 }
 
-test('key creation', async () => {
-    const { signingDriver } = await setupTest()
-    const err = await signingDriver.controller.createKey({ name: 'test' })
+test.skip('key creation', async () => {
```
PHOL-DA, line 166: any reason we skip this test ?

core/signing-fireblocks/src/index.ts
```diff
@@ -24,178 +24,222 @@ import {
...
+                    }
+                }
+            },
+
+            createKey: async (
+                _params: CreateKeyParams
+            ): Promise<CreateKeyResult> => {
                 return {
-                    error: 'bad_arguments',
+                    error: 'not_implemented',
```
PHOL-DA, line 189:
```suggestion
                    error: 'not_allowed',
```

core/signing-fireblocks/src/index.ts
```diff
@@ -24,178 +24,222 @@ import {
...
                     error_description:
-                        'either public key or txIds must be supplied',
+                        'Create key is not allowed in Fireblocks driver.',
```
PHOL-DA, line 191:
```suggestion
                        'Create fireblocks key through the Wallet Kernel is not allowed, please create new keys directly in fireblocks.',
```

core/signing-fireblocks/src/index.test.ts
PHOL-DA: nvm, saw the creating keys are not allowed can we instead remove this or catch the exception/error that indicates it is not allowed ?

PHOL-DA (approved):
Overall LGTM, i see a lot of TODO's, would be lovely if those could be created into tracked issues so we can easily identify them later on.

core/signing-fireblocks/src/index.test.ts
alexgraham-da: It actually already was that...I think the skip was added accidentally, will remove.

PR #182, opened 2025-07-29
requestAccounts

rukmini-basu-da:
There's an edge case where if the user creates a wallet first, then connects the dapp after, then the dapp won't have the list of existing wallets available because it wasn't connected at the time the event was fired. This adds a `requestAccounts` function to the dapp api that will return a list of wallets that the user has authorized the dapp to access.


example/src/App.tsx
```diff
@@ -90,6 +93,37 @@ function App() {
...
 
+    function requestAccounts() {
+        setError('')
+        const provider = window.splice
+
+        if (provider !== undefined) {
+            provider
+                .request({
+                    method: 'requestAccounts',
+                })
```
alexmatson-da, line 104: I was thinking it makes more sense to put this inside the `useEffect` above (like how it calls the `status` method). 

That'll call the method automatically when the component first renders, so we can update the primary party immediately (if we have existing wallets). The provider (& auth token) is cached, so this also refetches the data if the user refreshes the page

alexmatson-da (approved):
nice!

PR #189, opened 2025-08-01
manually initialize sequencer/mediator

rukmini-basu-da:
Manually initialize the sequencer/mediator so we can re-use the same synchronizerId. The motivation for this is because we add the synchronizerId to the network configuration, so it's useful to know this value upfront and will speed up the testing flow.


alexmatson-da (approved):
nice! thanks for figuring this out

PR #195, opened 2025-08-06
Integration Guide Example Scripts

PHOL-DA:
Created two example script to showcase how to integrate with a wallet kernel programmable (Intended for the Wallet Integration Guide)


rukmini-basu-da (approved):
Might be nice to have a readme for the example-scripts, but otherwise looks good!

example-scripts/00-utils/httpTransport.ts
```diff
@@ -0,0 +1,47 @@
+import {
+    JsonRpcRequest,
+    RequestPayload,
+    ResponsePayload,
+    RpcTransport,
+} from 'core-types'
+import { v4 as uuidv4 } from 'uuid'
+
+export class HttpTransport implements RpcTransport {
```
alexmatson-da, line 9: this is duplicated from the implementation in the remote client. I think we should publish our RpcTransport implementations (Marc is also working on one for the browser Window postMessage) to a separate package for wallet kit consumers to use

PR #222, opened 2025-08-13
Skeleton work for the wallet-sdk

PHOL-DA:
this PR includes:
- Clean up of all outstanding cleancoding script output (was annoying me).
- new folder SDK containing:
  - dapp-sdk (previous sdk)
  - wallet-sdk (new sdk for wallet providers against the ledger)
- ledger-core folder (this is a very generic component that will probable be re-used between clients and sdk for common commands and procedures).  
- script for fetching splice repo (and specifically the token standards)


core/ledger/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "core-ledger",
```
mjuchli-da, line 2:
```suggestion
    "name": "core-ledger-lib",
```

Implies that the folder name changes accordingly

mjuchli-da (commented):
I would consider enhancing the `core-ledger-client` with the transaction building blocks, rather than creating a dedicated library `core-ledger-lib|builder`:
- a dependency on `core-ledger-client` is required for the types
- updates on the openapi spec will have a cascading effect (spec -> ledger-client -> ledger-lib|builder)

E.g. start off by putting everything related into a `builder.ts` file and see if a dedicated module is necessary along the way.

mjuchli-da (approved):

PR #227, opened 2025-08-13
Sign with kernel provider keypair

### LINKED ISSUE #66: RPC Wallet Kernel for external parties
Extends Key Store of standalone RPC based Wallet Kernel application with external parties. Key storage is “on paper”. Demonstrate signing flow and successful topology transactions.

### LINKED ISSUE #220: Grant user rights on party allocation
The user initiating a party allocation should get the `actAs` and `readAs` right for the newly allocated party.

alexmatson-da:
Fixes #66 
Fixes #220


clients/remote/src/user-api/controller.ts
```diff
@@ -38,7 +38,64 @@ type AvailableSigningDrivers = Partial<
     Record<SigningProvider, SigningDriverInterface>
 >
 
+async function grantUserRights(
```
PHOL-DA, line 41: this is fine for now, but can we maybe add a TODO and issue to:
"instead of polling (and having a max-retry) we should instead listen for an event broadcast (new party created) and the grant user rights based on that".

PHOL-DA (approved):

clients/remote/src/user-api/controller.ts
mjuchli-da: Hm i can't reproduce this behaviour with internal parties (e.g. grant rights works immediately after allocating)

mjuchli-da (changes requested):
Wait for https://github.com/hyperledger-labs/splice-wallet-kernel/pull/225

mjuchli-da (approved):

mjuchli-da:
FYI: once rebased, the `operator` user is no longer a `participant_admin` and as a result I think you'll need to use the `ledgerClientAdmin` in some spots

core/wallet-user-rpc-client/src/openrpc.json
```diff
@@ -287,18 +300,7 @@
...
-                        "correlationId": {
-                            "title": "correlationId",
-                            "type": "string"
-                        },
-                        "traceId": {
-                            "title": "traceId",
-                            "type": "string"
-                        }
-                    },
-                    "required": ["correlationId", "traceId"]
```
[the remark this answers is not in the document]
mjuchli-da: > For an external party submission, I found that it was actually an empty object (even though it succeeded).

That's because the current execute is async. @thibault-da is working on a sync version I think – we should check.

clients/remote/src/user-api/controller.ts
```diff
@@ -34,16 +34,74 @@ import {
...
+    )
+
+    if (!result.newlyGrantedRights) {
+        throw new Error('Failed to grant user rights')
+    }
+
+    return
+}
+
 async function signingDriverCreate(
```
mjuchli-da, line 99: I think this is getting too complex. As far as I can see we just allocate a party in either of the possible ways. I therefore suggest we split it out into a `PartyAllocationService` for now, integration-test each way of allocating, and use this in the controller for now. This way, it also serves as a solid starting point to be integrated into the `wallet-sdk` (cc: @PHOL-DA).

clients/remote/src/user-api/controller.ts
```diff
@@ -85,13 +143,15 @@ async function signingDriverCreate(
                 signingProviderId,
             }
 
+            grantUserRights(wallet.partyId, userId, ledgerClientAdmin)
```
mjuchli-da, line 146: Given that we set the `userId` in the post request, this is not necessary AFAIK

PR #240, opened 2025-08-18
Removed auth context from signing-lib interface

PHOL-DA:
This removes the authContext from the signing libs.

Reasoning:
The signing library should not need to know the complete authentication context of the user simply to find an unique identifier for that user, this is only used for the fireblocks integration. Since an offline signing is needed for the wallet SDK, then an authcontext can not be provided.


mjuchli-da (commented):
Should we stick to the typed `UserId` (from `core-wallet-auth`)?

PR #243, opened 2025-08-18
enable offline signing mode

PHOL-DA:
a basic way to enable offline signing using the signing-lib


mjuchli-da (commented):
Is it worth also creating a method:

`createKeyPair` in `signing-lib`?

So that in the internal driver controller we will have:
```
createKey: async (
                params: CreateKeyParams
            ): Promise<CreateKeyResult> => {
                const key = base.createKeyPair()
...
```

mjuchli-da (approved):
Thx! 

Small thing: I think methods should be lowercase

PR #253, opened 2025-08-20
disable playwright install (and introduce cache)

PHOL-DA:
playwright is currently very flaky (see #254)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -44,6 +44,14 @@ jobs:
             - name: Run unit tests
               run: yarn test:all
 
+            - name: Cache Playwright Browsers
+              uses: actions/cache@v4
+              with:
+                  path: example/node_modules/.cache/ms-playwright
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -43,28 +43,36 @@ jobs:
...
+#                  path: example/node_modules/.cache/ms-playwright
+#                  key: playwright-browsers-${{ runner.os }}-${{ hashFiles('example/package-lock.json', 'example/yarn.lock') }}
+#                  restore-keys: |
+#                      playwright-browsers-${{ runner.os }}-
+#
+#            - name: Install Playwright Browsers
+#              run: yarn workspace example playwright install --with-deps
+#
+#            - name: Download Canton
+#              run: yarn script:fetch:canton
```
alexmatson-da, line 59: We can also cache the canton download, probably by `hashFiles('scritps/src/utils.ts') where we control the Canton version

PHOL-DA: yes, but did not seem like an expensive part of the run (but the playwright install sometimes took towards of 16 min)

alexmatson-da (approved):

PR #256, opened 2025-08-20
wallet sdk allocate internal/external party

### LINKED ISSUE #203: Party Allocation
This is the first step that any new wallet would potentially be looking at using.

should be based around extracting: https://github.com/hyperledger-labs/splice-wallet-kernel/blob/525d90277159feb7c7c9aa02a30a5b115c84aa17/clients/remote/src/user-api/controller.ts#L89C13-L156C18

rukmini-basu-da:
Closes #203


sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -17,20 +34,52 @@ export interface ledgerController {
...
+        token: string
+    ) {
         this.client = new LedgerClient(baseUrl, token, this.logger)
         this.userId = userId
         this.partyId = ''
         this.synchronizerId = ''
+        this.topologyClient = new TopologyWriteService(
+            this.synchronizerId,
+            adminApiUrl,
+            token,
```
PHOL-DA, line 79: the token provided is most likely not an admin token, so this would not work.

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -44,6 +93,101 @@ export class LedgerController implements ledgerController {
         return this
     }
 
+    async prepareExternalPartyTopology(
+        userId: string,
+        partyHint: string,
```
PHOL-DA, line 97: make partyHint optional

if partyHint is not defined then partyId should be: `${publicKey}::${namespace}`

this makes it deterministic for wallet providers and exchanges

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -44,6 +93,101 @@ export class LedgerController implements ledgerController {
         return this
     }
 
+    async prepareExternalPartyTopology(
+        userId: string,
```
PHOL-DA, line 96: use the userId that is defined on the object

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -1,8 +1,25 @@
...
+    PostResponse,
+    GetResponse,
+    TopologyWriteService,
+} from 'core-ledger-client'
 import { signTransactionHash } from 'core-signing-lib'
 import { v4 } from 'uuid'
 import { pino } from 'pino'
 
+export type PreparedParty = {
+    txs: Uint8Array<ArrayBufferLike>[]
```
PHOL-DA, line 12: instead of having `txs` & `txHashes` can we then have a `partyTransaction` containing a `tXRaw` & `txHash` and then have the preparedParty like:
`export type PreparedParty = {
    partyTransactions: partyTransaction[]
    combinedHash: string
}`

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -17,20 +34,52 @@ export interface ledgerController {
...
+
+    prepareExternalPartyTopology(
+        userId: string,
+        partyHint: string,
+        publicKey: string
+    ): Promise<PreparedParty>
+
+    submitExternalPartyTopology(
+        signedHash: string,
+        txs: Uint8Array<ArrayBufferLike>[],
```
PHOL-DA, line 50: instead of having these again, it could just use the `PreparedParty` that you used as a return type earlier

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -44,6 +93,101 @@ export class LedgerController implements ledgerController {
...
+        txHashes: Buffer<ArrayBuffer>[],
+        publicKey: string,
+        partyHint: string,
+        userId: string
+    ): Promise<AllocatedParty> {
+        const namespace =
+            TopologyWriteService.createFingerprintFromKey(publicKey)
+        const partyId = `${partyHint}::${namespace}`
```
PHOL-DA, line 135: lets just take a partyId as input parameter instead (and maybe have the prepare return the partyId)

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -44,6 +93,101 @@ export class LedgerController implements ledgerController {
...
+
+        await this.topologyClient.addTransactions(signedTopologyTxs)
+        await this.topologyClient.authorizePartyToParticipant(partyId)
+        await this.client.grantUserRights(userId, partyId)
+
+        return { partyHint, partyId, namespace }
+    }
+
+    async allocateInternalParty(
+        userId: string,
```
PHOL-DA: use `this.userId` instead

PHOL-DA (commented):
Left a few comments,

set the focus on making this as simple and easy as possible to use (each new parameter is a way a user can make a mistake)

once the comments are resolved i think it looks quite good

alexmatson-da (approved):

PR #265, opened 2025-08-21
docs: add party allocation example

rukmini-basu-da:
This adds an example for how to use the wallet-sdk to allocate an external party and command submission.

### DIFF
--- a/docs/wallet-integration-guide/examples/package.json
+++ b/docs/wallet-integration-guide/examples/package.json
@@ -13,6 +13,7 @@
         "run-01": "tsx ./01-auth.ts"
     },
     "devDependencies": {
+        "@types/node": "^24.3.0",
         "tsx": "^4.20.3",
         "typescript": "^5.8.3"
     },
--- a/docs/wallet-integration-guide/examples/package.json
+++ b/docs/wallet-integration-guide/examples/package.json
@@ -15,7 +15,8 @@
     "devDependencies": {
         "@types/node": "^24.3.0",
         "tsx": "^4.20.3",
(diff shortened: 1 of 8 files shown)


docs/wallet-integration-guide/examples/package.json
```diff
@@ -13,10 +13,13 @@
...
     },
     "devDependencies": {
+        "@types/node": "^24.3.0",
         "tsx": "^4.20.3",
-        "typescript": "^5.8.3"
+        "typescript": "^5.8.3",
+        "uuid": "^11.1.0"
     },
     "dependencies": {
+        "@splice/core-signing-lib": "workspace:^",
```
PHOL-DA, line 22: wallet integration guide should only depend on sdk-wallet.

if you need something from other libraries, then it needs to be exposed through the sdk-wallet

sdk/wallet-sdk/src/topologyController.ts
```diff
@@ -56,6 +57,7 @@ export class TopologyController {
         const namespace =
             TopologyWriteService.createFingerprintFromKey(publicKey)
 
+        //can't use base64 encoded public key because it has invalid charcters
```
PHOL-DA, line 60: is this a TODO or something to fix ?

and two lines below we use the publickey, can we either:
1. generate a valid partyHint based on the publicKey
2. use a generic hardcoded like 'wallet' of option 1 is not possible

PHOL-DA (changes requested):

PHOL-DA (approved):

PR #283, opened 2025-08-26
feat: wallet sdk token standard

mjuchli-da:


core/ledger-client/src/ledger-api-utils.ts
```diff
@@ -11,6 +11,7 @@ import {
 } from './constants'
 import { components } from './generated-clients/openapi-3.3.0-SNAPSHOT.js'
 import { LedgerClient } from './LedgerClient'
+import { Holding, TransferInstructionView } from './txparse/types'
```
mjuchli-da, line 14: @rukmini-basu-da  FYI: from what I can see, we should try to avoid these duplicated types from `txparse/types` and where possible work with the daml codegenned types.

PR #290, opened 2025-08-26
docs: adding an intro page

davidrichards-da:


PHOL-DA (approved):

docs/wallet-integration-guide/src/canton-network-overview/index.rst
```diff
@@ -10,54 +10,85 @@ Key Features:
...
+* **Validate** specific transactions, such as a transfer of assets.
+* **Control** certain actions, like initiating transfers.
+* **See** specific state and transactions, such as a record of their holdings.
+
+Privacy is maintained at the party level, meaning transaction and state data is only shared with the parties who need to see it, 
+ensuring a high degree of confidentiality.
+
+**Onboarding and Format**
+
+Parties are formatted as `name::fingerprint`, where the **fingerprint** is a unique identifier which can be generated from a public key (and thus also from a private key).
```
meiersi-da, line 48:
```suggestion
Parties are formatted as ``name::fingerprint``, where the **fingerprint** is a unique identifier which can be generated from a public key (and thus also from a private key).
```

rst uses double-backticks for code.

PR #294, opened 2025-08-27
feat(core-token-standard, core-ledger-client): type issues

pawelstepien-da:
Resolve unknowns from ledger client using token standard codegen

### DIFF
--- a/core/ledger-client/src/txparse/parser.ts
+++ b/core/ledger-client/src/txparse/parser.ts
@@ -31,10 +31,7 @@ import {
     TokenStandardChoice,
     TransferInstructionView,
 } from './types'
-import {
-    Transfer,
-    TransferInstructionResult_Output,
-} from '@splice/core-token-standard/src/token-standard-models-1.0.0/lib/Splice/Api/Token/TransferInstructionV1/index'
+import { TransferInstructionV1 } from '@splice/core-token-standard'
 
 import { components } from '../generated-clients/openapi-3.3.0-SNAPSHOT'
 import { LedgerClient } from '../ledger-client'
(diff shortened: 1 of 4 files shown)


core/ledger-client/src/txparse/parser.ts
```diff
@@ -31,6 +31,10 @@ import {
     TokenStandardChoice,
     TransferInstructionView,
 } from './types'
+import {
+    Transfer,
+    TransferInstructionResult_Output,
+} from '@splice/core-token-standard/src/token-standard-models-1.0.0/lib/Splice/Api/Token/TransferInstructionV1/index'
```
mjuchli-da, line 37: Can we export this in a nicer way from within `@splice/core-token-standard`?

core/token-standard/package.json
```diff
@@ -18,6 +18,7 @@
         "zod": "^3.25.67"
     },
     "devDependencies": {
+        "@daml/types": "3.4.0-snapshot.20250616.13911.0.v482d9ec0",
```
mjuchli-da, line 21: Out of curiosity, why is this required?

pawelstepien-da: In outputs from daml codegen, in respective package.jsons there is an entry
```
    "peer-dependencies": {
        "@daml/ledger": "3.4.0-snapshot.20250616.13911.0.v482d9ec0",
        "@daml/types": "3.4.0-snapshot.20250616.13911.0.v482d9ec0"
    },
```
This means that this package expects consuming package to provide those dependencies for it.

core/ledger-client/src/txparse/parser.ts
pawelstepien-da: Good call.
I exported contents of /token-standard-models-1.0.0/lib/Splice/Api/Token from core/token-standard
This gives us types wrapped in respective daml module namespace, like `TransferInstructionV1.Transfer`

mjuchli-da (approved):

PR #298, opened 2025-08-27
ci: add docs to the build

da-tanabe:
Make sure that docs run as part of CI.

Right now, the docs have a lot of warnings, and we treat warnings as errors in the centralized docs build. In a subsequent PR, I'll fix the associated errors/warnings and flip that flag on.


alexmatson-da (approved):
To quote Indiana Jones: "Why did it have to be snakes?" 🐍

PR #305, opened 2025-08-28
docs: document canton coin mvp workflows

meiersi-da:


docs/wallet-integration-guide/src/exchange-integration/workflows.rst
```diff
@@ -0,0 +1,245 @@
...
+      Exchange, and Admin validator nodes. It is assigned an
+      update-id ``upd567`` and a record time ``t1`` by the Global
+      Synchronizer. It is assigned offset ``off1`` by the Exchange
+      Validator Node.
+
+3. Tx History Ingestion observes ``upd567`` at ``t1`` with offset
+   ``off1`` and updates the Canton Integration DB as follows.
+
+   a. Tx History Ingestion parses ``upd567`` using the token standard tx
+      history parser from the Wallet SDK to determine:
```
PHOL-DA, line 79: the Wallet SDK does not currently have a tx history parser, is this something we should add?

docs/wallet-integration-guide/src/exchange-integration/workflows.rst
```diff
@@ -0,0 +1,245 @@
...
+6. Customer observes the completion of the withdrawal at ``t1`` in the
+   Exchange UI and the receipt of the expected funds in their Customer Wallet.
+
+
+.. _utxo-management:
+
+UTXO Selection and Management
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: write this
```
PHOL-DA, line 198:
```suggestion
.. TODO: write this
```

docs/wallet-integration-guide/src/exchange-integration/workflows.rst
```diff
@@ -0,0 +1,245 @@
...
+.. _mvp-for-cn-tokens:
+
+MVP for all Canton Network Tokens
+---------------------------------
+
+
+Multi-Step Deposit Flow
+^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: add diagram, and flow description
```
PHOL-DA, line 235:
```suggestion
.. TODO: add diagram, and flow description
```

docs/wallet-integration-guide/src/exchange-integration/workflows.rst
```diff
@@ -0,0 +1,245 @@
...
+
+
+Multi-Step Withdrawal Flow
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+TODO: add diagram, polish write-up
```
PHOL-DA, line 241:
```suggestion
.. TODO: add diagram, polish write-up
```

PHOL-DA (approved):

docs/wallet-integration-guide/src/exchange-integration/workflows.rst
meiersi-da: Yes. I suggest to copy the one from the the token standard CLI together with its test files. Going forward I'd expect it to be maintained as part of the wallet SDK. It is a core piece of using the token standard.

This will mean that we ideally consume the parser from the wallet SDK repo in CN for the integration testing we do there. @moritzkiefer-da @OriolMunoz-da would that be an option? 

As an increased hardening step, we might want to talk to @matteolimberto-da whether he could provide similar test-data files for the upgraded utility that adds metadata for their legacy workflows; and whether there would be an option to setup an integration test for tx history parsing using the wallet SDK as part of the utilities repo as well.

meiersi-da: Hmm indeed. Not great. So I'd probably do the same we do for the Canton fork. Copy the sources into CN as a fork and regularly update them to minimize the diff.

This way diffs can be handled as part of a change, and upstreamed afterwards.

meiersi-da: (Resolving so merging is not blocked :shrug: -- will unresolve once merged.)

PR #317, opened 2025-08-28
fix(wallet-sdk,dapp-sdk): fill out package readmes for the SDKs

alexmatson-da:


mjuchli-da (approved):
As per comment, i recommend adding an "install" section

PR #323, opened 2025-08-29
test(core-ledger-client): parser unit tests

pawelstepien-da:
Added unit tests for tx parser


core/ledger-client/src/txparse/test-data/mock/txs.ts
```diff
@@ -0,0 +1,7081 @@
+export default [
```
meiersi-da, line 1: Please add a README.md to `test-data/` that explains where it comes from; and how to update the data.

I assume it was copied from https://github.com/hyperledger-labs/splice/tree/main/token-standard/cli/__tests__, correct?

Please also add a note that we expect the behavior of the parser in the wallet-sdk to stay in sync with the reference parser in the `splice/token-standard`.

FYI: @PHOL-DA

pawelstepien-da: Okay I will add README.md

Yes I copied test data from cli tests. 
But I modified expected output a bit, because in the CLI what is tested is output of the whole list-holding-txs command, which after processing mock data with `TransactionParser.parseTransaction`, post processes it by filtering out txs with empty events and mapping events objects shape
https://github.com/hyperledger-labs/splice/blob/main/token-standard/cli/src/commands/listHoldingTransactions.ts#L93

I tested input/output of `TransactionParser.parseTransaction` in separation of the command. 
Is that okay?

meiersi-da: Yes, that makes sense. Please resolve once you've added the README.

mjuchli-da (approved):

PR #334, opened 2025-08-31
docs: adding the curl examples to the preparing and signing a transac…

davidrichards-da:
…tion page


PHOL-DA (approved):

docs/wallet-integration-guide/examples/bash/sign-transaction-hash.sh
```diff
@@ -0,0 +1,5 @@
+#  The hash is signed using openssl.
+
+TRANSACTION_HASH=$(cat create_ping_prepare_response.json | jq -r .prepared_transaction_hash)
+PREPARED_TRANSACTION=$(cat create_ping_prepare_response.json | jq -r .prepared_transaction)
+SIGNATURE=$(echo -n "$TRANSACTION_HASH" | base64 --decode | openssl pkeyutl -rawin -inkey "PRIVATE_KEY_FILE" -keyform DER -sign | openssl base64 -e -A)
```
thibault-da, line 5: This is out of context of the docs it was taken from. It's not clear what `create_ping_prepare_response.json` is here.
Maybe just say it's the JSON output of the `curl -X POST http://YOUR_NODE_JSON_API/v2/interactive-submission/prepare` command

thibault-da: Also there's no information about where they should get "PRIVATE_KEY_FILE" from or what key it is. It should be the private key of the namespace of the external party. Don't know how deep you want to go there but I'd at least mention that. You can copy these openssl commands if you want to show how we expect this key to have been generated: https://docs.digitalasset.com/build/3.3/tutorials/app-dev/external_signing_topology_transaction.html#signing-keys

PR #341, opened 2025-09-01
feat(core-ledger-client, dapp-sdk): listing contracts by interface

pawelstepien-da:
Added output processing in TokenStandardService.listContractsByInterface so the return value is in same shape as in CLI.
Added corresponding method to TokenStandardController.
Refactored InterfaceId usage in TokenStandardService to be simple strings using external function for matching, instead of passing around object with methods, so it's easier to use from outside of module (like in controller), or by importing templateIds from token standard codegen.


core/ledger-client/src/txparse/types.ts
```diff
@@ -219,3 +223,8 @@ export interface PrettyTransactions {
     transactions: Transaction[]
     nextOffset: number
 }
+
+export interface PrettyContract<T = ViewValue> {
+    contractId: string
+    payload: T
```
mjuchli-da, line 229: I think we'll lose some information here. Wasn't it the case (and I'm not sure) that previously we potentially had the `JsActiveContract` as part of the `JsGetActiveContractsResponse.JsContractEntry`?
It seems that we filter by `JsActiveContract` anyway, and so could we just add this as an additional property to this `PrettyContract` (and thereby drop `contractId` as it's part of `JsActiveContract`)?

pawelstepien-da: Yes we lose some information along the way, which I think is purpose of `toPrettyContracts` to make output lightweight without info that was deemed unnecessary. Currently `payload` is only containing interface view fields. `toPrettyContracts` was commented in CLI as below:
`// Make them nicer to show by excluding stuff useless to users such as the createdEventBlob`

We can replace `contractId` with `activeContract` containing whole `JsActiveContract` object, but this would include `createdEventBlob`, so going against the intention in CLI. 
Or perhap we can alter the shape of this object, to contain some of those fields of `JsActiveContract`, but without the blob.
If so, I'm not sure which info we want to omit and which to include.

mjuchli-da: I lack context here and it is likely that I oversee something. My assumption was that this particular bit you're working on allows filtering by interface. Therefore, the result will contain the `View` (check) but it could also contain the `Contract` instance. It is then up to the CLI to pick what to display. @meiersi-da @OriolMunoz-da what is your take on this?

pawelstepien-da: @mjuchli-da and I agreed to add there a property like `activeContract` which would hold contents of `JsActiveContract` from acs response entries. Then SDK would provide all available info.
I will rename `payload` to something like `interfaceViewValue` and leave it along with `contractId` for convenience.

core/ledger-client/src/token-standard-service.ts
```diff
@@ -96,12 +111,30 @@ export class TokenStandardService {
...
-                }
+                })
+
+            /*  This filters out responses with entries of:
+                - JsEmpty
+                - JsIncompleteAssigned
+                - JsIncompleteUnassigned
+                while leaving JsActiveContract
+                Those removed entries should not affect returned contracts output, nor trigger error
+             */
```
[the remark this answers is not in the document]
pawelstepien-da: I based my assumption that only that event should be used on this line in CLI:
https://github.com/hyperledger-labs/splice/blob/main/token-standard/cli/src/commands/listContractsByInterface.ts#L48
If there was different event, it would fail at runtime. 

I will create an issue to add support for multiple synchronizers and update the comment.

mjuchli-da (approved):
Thanks. We agreed to add follow-up PRs with the improvements suggested here.

core/ledger-client/src/token-standard-service.ts
```diff
@@ -356,4 +389,18 @@ export class TokenStandardService {
                 .map(renderTransaction),
         }
     }
+
+    // Make them nicer to show by excluding stuff useless to users such as the createdEventBlob
```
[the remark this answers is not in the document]
pawelstepien-da: Yes, that's the only interface caller gets. But will be extended with JsActiveContract, createdEventBlob included. I will adjust the comment accordingly.

PR #376, opened 2025-09-05
docs: fix warnings

meiersi-da:
@mjuchli-da : did you ever discuss requiring trailing as part of a pre-commit hook? and potentially using `.editorconfig` like CN to do so by default on save in most editors?

- https://editorconfig.org/
- https://github.com/hyperledger-labs/splice/blob/main/.editorconfig


da-tanabe:
> Also @da-tanabe : I remember that you wanted to check warnings in CI. Did I do something wrong in my checkout or is that not yet enforced?

It's not yet enforced because I couldn't keep up with the volume of docs being written last week. I'll take a look at flipping this next week when the wallet kernel team is in planning.

mjuchli-da (approved):
Thanks for the suggestion, I will add the `.editorconfig`

PR #377, opened 2025-09-05
chore: generate openapi clients directly from urls

alexmatson-da:
so apparently the openapi generator can easily take a spec file directly from a remote url... this removes all the checked-in spec files and the tooling scripts we've developed for manually downloading, extracting, copying(, etc) various specs from various upstream sources.

the URLs are pinned to specific git revisions, to ensure the contents don't change out from under us (and that we are targeting specific upstream versions). For example:

- (ledger JSON api): https://raw.githubusercontent.com/digital-asset/daml/refs/tags/v3.3.0-snapshot.20250603.0/sdk/canton/community/ledger/ledger-json-api/src/test/resources/json-api-docs/openapi.yaml
- (token standard): https://raw.githubusercontent.com/hyperledger-labs/splice/refs/tags/0.4.11/token-standard/splice-api-token-allocation-instruction-v1/openapi/allocation-instruction-v1.yaml


mjuchli-da (changes requested):
I'm afraid the scan won't work. We need to rely on the packaged tarball which contains the normalized api specs (without relative paths in it) found here: https://github.com/digital-asset/decentralized-canton-sync/releases. In fact, this tarball contains all splice and token standard related api specs. I suggest:
- We adjust the token-standard fetching (to the same way we fetch scan)
- Leave ledger-api specs as is (it personally helps me having it in the repo and being able to open it besides the code...)

PR #384, opened 2025-09-08
feat: allow defining ledger end offset

PHOL-DA:


meiersi-da (approved):
Thanks. Please rename the PR title though to not talk about "time", but "offset".

PR #387, opened 2025-09-08
feat: added endpoint to fetch registry admin party

PHOL-DA:


core/ledger-client/src/token-standard-service.ts
```diff
@@ -93,6 +93,16 @@ export class TokenStandardService {
         }
     }
 
+    async getInstrumentAdmin(
+        transferFactoryRegistryUrl: string
+    ): Promise<string | undefined> {
+        const client = this.tokenStandardClient(transferFactoryRegistryUrl)
+
+        const info = await client.get('/registry/metadata/v1/info')
```
meiersi-da, line 101: Note that doing a single read of the admin party-id is not safe for a decentralized token. 
The problem is that the one node being contacted, can choose the registry party. 

It can be made safe for CC by running it against the Scan Proxy on a validator node.

So, I'm not sure whether exposing this function is a good idea for the SDK.
The mapping from admin-party to endpoint is security sensitive. Until we have the [long-term solution based on CNS in place](https://github.com/global-synchronizer-foundation/cips/blob/main/cip-0056/cip-0056.md#off-ledger-api-discovery-and-access), is best managed client-side as described here: https://docs.digitalasset.com/integrate/devnet/exchange-integration/workflows.html#canton-network-token-onboarding

PHOL-DA: i don't assume the scan proxy works for all token standards? this is based on the registry factory url provided (so this can be used for other tokens as well)

we can upgrade the local defaults to point to scan proxy instead of scan though

mjuchli-da (approved):

PR #390, opened 2025-09-08
chore: aggregate openapi generation

alexmatson-da:
This is kind of the best of both worlds:

1. The new script `generate-openapi-clients.ts` makes a centralized spot to add new OpenAPI specs and generate TS clients
2. It supports both URL (remote) and file-based YAML specs
3. We still fetch the Splice specs from the published tarball, but now we dump them directly to `api-specs/...` instead of `.splice-specs/...`
4. I switched to the URL approach for the ledger JSON API specs, but also added the option to write the file to `api-specs/...` so that we have a repo-local mirror (helpful for development/debugging) 

There is also more cleanup / fixes around the `fetch-canton` and `fetch-splice` scripts, so that they reuse the same tarball downloader and file integrity checker (now moved to utils)


mjuchli-da (approved):
thx!

PR #422, opened 2025-09-10
docs: exchange integration - Fix reference for treasury party setup

moritzkiefer-da:
probably wanna setup some CI checks but let's focus on the content first


meiersi-da (approved):
Yeah @da-tanabe has been promising multiple times to hook it up to CI... we should probably just do it.

PR #440, opened 2025-09-10
docs: improve sectioning of exchange integration

meiersi-da:


docs/wallet-integration-guide/src/exchange-integration/workflows.rst
```diff
@@ -3,6 +3,12 @@
 Integration Workflows
 =====================
 
+.. contents::
+   :local:
+   :depth: 2
+   :backlinks: none
```
meiersi-da, line 9: added as https://docs.digitalasset.com/integrate/devnet/exchange-integration/workflows.html only shows the top-level sections in the right side-bar

moritzkiefer-da (approved):
thanks!

PR #448, opened 2025-09-10
docs: exchange integration - Move transaction history parsing to its own section

moritzkiefer-da:
No real content change, will rephrase it to handle withdrawals and deposits together next.

### DIFF
--- a/docs/wallet-integration-guide/src/exchange-integration/index.rst
+++ b/docs/wallet-integration-guide/src/exchange-integration/index.rst
@@ -21,11 +21,9 @@ than exchanges that want to support deposits and withdrawals of Canton Network t
    overview
    architecture
    workflows
+   txingestion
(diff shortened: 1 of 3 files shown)


docs/wallet-integration-guide/src/exchange-integration/txparsing.rst
```diff
@@ -0,0 +1,180 @@
+.. _transaction-parsing:
+
```
meiersi-da: Is there a top-level title missing?

meiersi-da: e.g., "Tx History Ingestion Implementation" or "Tx History Ingestion Details"

meiersi-da (approved):
:pray: thanks

PR #456, opened 2025-09-11
docs: exchange integration - kms

meiersi-da:


moritzkiefer-da (approved):
Thanks! Is it worth mentioning that migration is painful so you should very much consider doing that from the start instead of as a later improvement?

PR #474, opened 2025-09-12
ci: parallelize test tasks

alexmatson-da:
This PR:

- Splits up the large `build` job in CI workflow to distinctly separate jobs for different tests (static tests, unit tests, and heftier e2e tests)
- Switches to NX for running repo builds & tests, which allows us to make use of the NX build cache for CI


mjuchli-da (approved):
Nice!

PR #476, opened 2025-09-13
chore: add snapshot testing for WIG docs snippets

### LINKED ISSUE #330: Test docs code snippets in CI
We check that the code examples can compile, but we should run the examples like unit tests to make sure we don't break any behavior

alexmatson-da:
Sets up `jest` to run all snippets defined in `examples/snippets` in the WIG. Uses snapshot testing to just ensure the same snippet generates the same output every time. Requires localnet to be running

Snippets can be excluded from the test runner on a case-by-case, by putting a custom comment anywhere in the snippet file: `// @disable-snapshot-test` 

See #450 for CI testing the example scripts


PHOL-DA (approved):
can we run this in the CI aswell ?

PR #496, opened 2025-09-17
feat(core-ledger-client, wallet-sdk): getting tx by id

pawelstepien-da:


sdk/wallet-sdk/src/tokenStandardController.ts
```diff
@@ -143,6 +143,16 @@ export class TokenStandardController {
             beforeOffset
         )
     }
+    /** Lists all holdings for the current party.
+     * @param updateId id of queried transaction
+     * @returns A promise that resolves to a transaction
+     */
+    async getTransactionById(updateId: string): Promise<Transaction> {
+        if (this.partyId === undefined) {
```
PHOL-DA, line 151: you can use 'this.getPartyId()' instead, that will throw an error if the party is not defined

PHOL-DA (approved):

sdk/wallet-sdk/src/tokenStandardController.ts
pawelstepien-da: Good catch! No reason to repeat that check.

PR #497, opened 2025-09-17
feat(wallet-sdk): observe command completion

### LINKED ISSUE #482: [Wallet SDK] follow Command to completion
currently in the SDK you can submit a command, however you have no way to figure out if it has actually completed.

this can be done by extending:
`/v2/commands/completions`
to figure out the status (call ledger-end just before and use that for querying)

this also allow to extend or create new functions for:
`executeSubmission` & `prepareSignExecuteSubmission` on the userLedger to allow waiting for the command to complete (or fail)

mjuchli-da:
Fixes #482


sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -108,7 +108,62 @@ export class LedgerController {
...
+                    userId: this.userId,
+                    parties: [this.getPartyId()],
+                    beginExclusive: offset,
+                }
+            )
+
+            // Find the completion for our commandId
+            const found = completions.find(
+                (completion) =>
+                    completion.completionResponse?.Completion?.value
```
PHOL-DA, line 144: does this also work in case that the completion has an error ?

PHOL-DA (approved):

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -108,7 +108,62 @@ export class LedgerController {
...
+                return found.completionResponse.Completion.value
+            }
+
+            // Optionally update offset if completions returns a new offset
+            const offsets = completions.map(
+                (c) =>
+                    c.completionResponse?.OffsetCheckpoint?.value.offset ||
+                    offset
+            )
+            offset = Math.max(...offsets)
```
pawelstepien-da, line 158: I think this is what made script in CI fail. If completions is an empty array then `Math.max(...[]) === -Infinity`

sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -121,48 +124,24 @@ export class LedgerController {
...
-            // Wait a bit before polling again
-            await new Promise((res) => setTimeout(res, 500))
-        }
-
-        throw new Error(
-            `Timeout waiting for completion of commandId: ${commandId}`
+        const completionPromise = awaitCompletion(
+            this.client,
+            ledgerEnd,
+            this.partyId!,
```
pawelstepien-da, line 135: Let's use this.getPartyId(). It's either string or throws error.

core/ledger-client/src/ledger-api-utils.ts
```diff
@@ -388,11 +394,26 @@ async function awaitCompletion(
...
+                completion.value.commandId === commandId
+            )
+        } else if (submissionId) {
+            return (
+                completion.value.userId === userId &&
+                completion.value.submissionId === submissionId
+            )
+        } else {
+            return false
+        }
```
pawelstepien-da, line 416: What do you think about flattening those conditions as:
```
if (completion.value.userId !== userId) return false;
if (commandId && completion.value.commandId !== commandId) return false
if (submissionId && completion.value.submissionId !== submissionId) return false

return true
```
And the else part would be already handled by the throw at first lines of function body.

pawelstepien-da (approved):

core/ledger-client/src/ledger-api-utils.ts
mjuchli-da: much better, thank you

PR #511, opened 2025-09-19
fix: update default dapp port

mjuchli-da:


core/wallet-ui-components/src/components/Discovery.ts
```diff
@@ -82,7 +82,7 @@ export class Discovery extends HTMLElement {
     }
 
     verifiedKernels(): DiscoverResult[] {
-        return [{ url: 'http://localhost:3000/rpc', walletType: 'remote' }]
+        return [{ url: 'http://localhost:3008/rpc', walletType: 'remote' }]
```
alexmatson-da, line 85: maybe we should also change the default port to `3008` in https://github.com/hyperledger-labs/splice-wallet-kernel/blob/b791198f97481a26cca794d91032a33b6dae91aa/clients/remote/src/index.ts#L19

since if someone starts the kernel without the env var override then this will be broken in the other direction

pawelstepien-da (approved):

PR #514, opened 2025-09-19
feat(wallet-sdk): check transfer pre approval by party and dso and expiration time

### LINKED ISSUE #499: [Wallet SDK] identify if receiver have set up transfer pre-approval
Create new functionality that identifies if a receiver have set up transfer pre-approval (and are eligible for 1-step transfer).

this can be done through the scanProxy using `v0/scan-proxy/transfer-preapprovals/by-party/{party}`

ideally this should return an object with includes the expiry time (so they can verify themselves if it has already expired)

rukmini-basu-da:
closes [#499](https://github.com/hyperledger-labs/splice-wallet-kernel/issues/499)

### DIFF
--- a/sdk/wallet-sdk/src/validatorController.ts
+++ b/sdk/wallet-sdk/src/validatorController.ts
@@ -14,6 +14,12 @@ import {
 } from '@canton-network/core-signing-lib'
 import { PartyId } from '@canton-network/core-types'
 
+export type TransferPreApprovalLookup = {
+    receiverParty: string
+    dso: string
+    expiresAt: string
+}
+
 /**
  * TokenStandardController handles token standard management tasks.
  * This controller requires a userId and token.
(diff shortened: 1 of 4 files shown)


sdk/wallet-sdk/src/validatorController.ts
```diff
@@ -193,20 +193,29 @@ export class ValidatorController {
...
 
     /**  Lookup a TransferPreapproval by the receiver party
-     * @param partyId receiver party id
-     * @returns A promise that resolves to an array of
-     * transfer preapprovals by party.
+     * @param receiverId receiver party id
+     * @param dsoParty the instrument partyId that has transfer preapproval
+     * @returns A promise that returns a boolean of whether an instrument can be transferred directly to the receiver
      */
+    async getTransferPreApprovalByParty(receiverId: string, dsoParty: string) {
```
PHOL-DA, line 200: i would use the `PartyId` type instead for these two.

`dsoParty` is the correct party to use for canton coin, however for any other instrument that would not be the case in the future (transfer pre-approval will become part of the token standard).

while not needed right now we should also have a check on the instrument name, just to make sure it will be backwards compatable once it becomes part of the token standard.

lastly i would change the return type from `boolean` to an object and give it the expiry date, the reason being that clients might want to evaluate based on that (i.e. not create the transfer if the expiry is less than 1 minute away).

alexmatson-da (approved):

PR #515, opened 2025-09-19
chore: add script to fix release tags automatically

alexmatson-da:
does what it says: 

Screenshot 2025-09-19 at 9 58 40 AM

Shows a preview before applying, and automatically skips tags that are already fixed


mjuchli-da (approved):
🚂🚂🚂

PR #516, opened 2025-09-19
chore: get wallet sdk deps dynamically

### LINKED ISSUE #437: Improve package validator test
We have a script (`yarn script:validate:package`) which generates a publishable NPM artifact for the wallet-sdk, and practices installing it in a tmp directory, in order to detect any possible faulty dependencies or package.json misconfigurations. This should be extended to:

- ~~package _all_ monorepo subdependencies (right now it only does `core-token-standard`, but any other one could also have undetected issues)~~ #516 

- run in both a `yarn` setup and `npm` setup (npm will likely be more widely used in the community)

- run a more complete diagnostic that repeats a basic example script (instantiate a WalletSDK instance, call a few simple methods on it... nothing crazy, since we test this extensively elsewhere)

alexmatson-da:


PHOL-DA (approved):

PHOL-DA:
i dont think this complete "fixes #437" since it does not run the example scripts as part of it.

PR #538, opened 2025-09-24
feat(core-ledger-client,wallet-sdk): upgrade to support splice 0.4.17

### LINKED ISSUE #539: [Wallet SDK] hanging endlessly when creating a party
Sometimes when calling the execute part of creating an external party the code will hang endlessly.

Result of debugging / replicating:

the problem lies with https://github.com/hyperledger-labs/splice-wallet-kernel/blob/46950df8f94fc288e58d07bbf74682aebd576008/core/ledger-client/src/ledger-client.ts#L116C9-L129C10

```
        while (!partyFound || tries >= maxTries) {
            const parties = await this.get('/v2/parties')
            partyFound =
                parties.partyDetails?.some(
                    (party) => party.party === partyId
                ) || false

            await new Promise((resolve) => setTimeout(resolve, 1000))
            tries++
        }

        if (tries >= maxTries) {
            throw new Error('timed out waiting for new party to appear')
        }
```
if the party is not found within the retries (5), then it will be stuck in an endless while loop

PHOL-DA:
Fixes #539


rukmini-basu-da (approved):

core/wallet-dapp-rpc-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@canton-network/core-wallet-dapp-rpc-client",
-    "version": "0.3.2",
+    "version": "0.1.0",
```
pawelstepien-da, line 3: Why decrease version? Isn't that gonna prevent publishing to npm?

pawelstepien-da (approved):

PR #547, opened 2025-09-24
feat: add internal core-rpc-errors package

### LINKED ISSUE #233: Login page requires a refresh to redirect
When logging into a new session, the callback page is supposed to redirect you to the home page. However, it usually fails because the `addSession` call returns a 500.

Image

### LINKED ISSUE #530: Consistent error handling
RPC error codes should become our way of consistently report/throw errors.
From these codes, we will derive the error codes which match the transport layer (e.g. for HTTP we derive 4XX from the RPC code).

One way of facilitating this is to re-export what is available here: https://github.com/MetaMask/rpc-errors
and make a `@canton-network/rpc-errors` workspace for it.

### LINKED ISSUE #533: Fix failed reload after login
Remote kernel does not refresh view after login.

alexmatson-da:
Fixes #530 
Fixes #533 
Fixes #233

### DIFF
--- a/clients/extension/package.json
+++ b/clients/extension/package.json
@@ -23,10 +23,10 @@
         "web-ext": "^8.8.0"
     },
     "dependencies": {
+        "@canton-network/core-rpc-errors": "workspace:^",
         "@canton-network/core-types": "workspace:^",
         "@canton-network/core-wallet-store": "workspace:^",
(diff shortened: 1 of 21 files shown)


rukmini-basu-da (approved):

mjuchli-da (approved):
thank you!

PR #551, opened 2025-09-25
docs: write token standard integration guide

PHOL-DA:


docs/wallet-integration-guide/examples/snippets/accept-or-reject-transfer.ts
```diff
@@ -0,0 +1,56 @@
...
+        )
+} else {
+    //accept the transaction
+    const [acceptTransferCommand, disclosedContracts] =
+        await sdk.tokenStandard!.exerciseTransferInstructionChoice(
+            myPendingTransactionCid,
+            'Accept'
+        )
+
+    const rejectCommandId =
```
pawelstepien-da, line 49:
```suggestion
    const acceptCommandId =
```

docs/wallet-integration-guide/examples/snippets/create-transfer.ts
```diff
@@ -0,0 +1,52 @@
...
+const sdk = new WalletSDKImpl().configure({
+    logger: console,
+    authFactory: localNetAuthDefault,
+    ledgerFactory: localNetLedgerDefault,
+    tokenStandardFactory: localNetTokenStandardDefault,
+})
+
+const sender = 'sender-party'
+const senderKey = 'private-key-for-my-party'
+const instrumentAdminPartyId = 'Admin of the instrument'
```
pawelstepien-da, line 18: I think this value could make zod unhappy when used as `PartyId`, because we have regex check `/^[a-zA-Z0-9:_-]*::[a-z0-9]*/`
Tho I'm not sure if it matters in case of snippets.

alexmatson-da (approved):

pawelstepien-da (approved):

PR #581, opened 2025-09-30
fix: return user url as part of kernel info

mjuchli-da:


core/wallet-dapp-rpc-client/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "@canton-network/core-wallet-dapp-rpc-client",
-    "version": "0.5.0",
+    "version": "0.1.0",
```
pawelstepien-da, line 3: Version got resetted. I was it because of codegen?

alexmatson-da: Yes. I actually think this might not be an issue, because Nx's formula for versioning (at release-time) is 

```
new_version = (latest version from git tag) + (semantic bump determined from commits)
```

Reading from package.json is only a fallback if there are no existing tags (i.e. first-time package publishing).

As for why "0.1.0" -- I believe the OpenRPC generator uses the version in the spec JSON (`info.version`), so perhaps we should also aim to keep this in sync with our releases

PHOL-DA (approved):

PR #590, opened 2025-10-01
feat(core-ledger-client, wallet-sdk): self-issue a FeaturedAppRight

rukmini-basu-da:
Related to https://github.com/hyperledger-labs/splice-wallet-kernel/issues/492

This sets up the featured exchange party. As per the[ doc recommendations](https://docs.digitalasset.com/integrate/devnet/exchange-integration/node-operations.html#reward-minting-and-traffic-funding), I've added a command to create it for an internal party (the validator user), but we can also self grant the rights for an external party.

### DIFF
--- a/core/ledger-client/src/token-standard-service.ts
+++ b/core/ledger-client/src/token-standard-service.ts
@@ -83,6 +83,19 @@ export class TokenStandardService {
         return transfer_preapproval
     }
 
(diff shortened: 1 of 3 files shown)


PHOL-DA (approved):
Very cool, do we need to do anything for getting rewards when performing transactions ? Also for external parties you need to self claim rewards to my knowledge

could we potentially also showcase with an example that proves we are getting rewards ?

PR #662, opened 2025-10-13
fix(clients-remote): add support for devnet auth0 validator

### LINKED ISSUE #224: Connect to devnet
Create a new Auth0 application and configure a new network (e.g. "DA Devnet") against the devnet validator

alexmatson-da:
Fixes #224


clients/remote/src/user-api/controller.ts
```diff
@@ -121,7 +121,7 @@ export const userController = (
             ).fetchToken({
                 clientId: network.auth.admin.clientId,
                 clientSecret: network.auth.admin.clientSecret,
-                scope: network.auth.scope,
+                scope: 'daml_ledger_api',
```
mjuchli-da: Why the hard code here?

alexmatson-da: because you need different scopes for the implicit auth used on the login page vs the client credentials auth used for the admin token

mjuchli-da: we should adjust the config then. Otherwise we'll drag it along for too long

alexmatson-da: hmm actually, seems like everything still works if I change the scope config to only `daml_ledger_api` for the user auth too 

Screenshot 2025-10-16 at 10 16 10 PM

mjuchli-da (approved):

PR #672, opened 2025-10-14
fix(core-wallet-ui-components,example): fix playwright script after ui changes

PHOL-DA:


alexmatson-da (approved):
much appreciated! after its merged, reminder to re-enable the CI blocking behavior 😁

PR #679, opened 2025-10-15
chore: upgrade splice

PHOL-DA:


pawelstepien-da (approved):
Looks good. 
Could you also run generation scripts and commit api specs / generated clients for 4.20, so it's in one PR?

PR #692, opened 2025-10-16
chore: make Start Canton wait for boostrap completion

PHOL-DA:


.github/workflows/build.yml
```diff
@@ -151,7 +151,28 @@ jobs:
...
+                        exit 0
+                      fi
+                      if ! kill -0 $pid 2>/dev/null; then
+                        echo "Canton process exited unexpectedly."
+                        cat "$LOGFILE"
+                        exit 1
+                      fi
+                      sleep 5
+                    done
+                    echo "Timeout: 'Bootstrap script successfully executed.' not found in 5 minutes."
```
pawelstepien-da: Isn't that an issue that we have here single quotes, while the whole thing is also wrapped in single quotes starting from `timeout 5m bash -c '`?

PHOL-DA: Nope works just fine 👍🫠 you can see it in the CI that start Canton completes after 40 s

pawelstepien-da (approved):

PR #700, opened 2025-10-16
feat(core-ledger-client, wallet-sdk): preapproval cancel and renew

### LINKED ISSUE #569: [Wallet SDK] allow cancellation of transfer Pre-approval
The default setup for transfer pre-approval setup (using the validator as the operator party) forces a renewal every 30 days when close to expiry.

We should allow cancellation of this using:
https://github.com/hyperledger-labs/splice/blob/main/daml/splice-amulet/daml/Splice/AmuletRules.daml#L1527

We should allow manual renewal using:
https://github.com/hyperledger-labs/splice/blob/b63f61048d336a7a08e5a5abb7b0139d73c6c755/daml/splice-amulet/daml/Splice/AmuletRules.daml#L1499

pawelstepien-da:


core/ledger-client/src/token-standard-service.ts
```diff
@@ -1332,4 +1341,88 @@ export class TokenStandardService {
...
+                synchronizerId,
+            },
+        ]
+
+        const inputHoldings = await this.core.getInputHoldingsCids(
+            provider,
+            inputUtxos
+        )
+
+        const newExpiresAt = new Date(
```
PHOL-DA: i would give this as a input parameter for the method instead

docs/wallet-integration-guide/examples/scripts/14-token-standard-preapproval-renew-cancel-localnet.ts
```diff
@@ -0,0 +1,140 @@
...
+// That function waits until getTransferPreApprovalByParty returns new CID
+const preapprovalAfterRenewal =
+    await sdk.tokenStandard!.waitForPreapprovalFromScanProxy(
+        receiver!.partyId,
+        'Amulet',
+        {
+            oldCid: preapproval.contractId,
+        }
+    )
+if (!preapprovalAfterRenewal) {
```
PHOL-DA: can we also validate that the new preapproval actually have a different expiry date ?

PHOL-DA (approved):

PR #719, opened 2025-10-17
feat(core-ledger-client): added retry logic and stress testing

### LINKED ISSUE #744: [Wallet SDK] add retry logic for get and post against ledger
retry logic should be in place for get and post against the ledger, especially for when under high stress

PHOL-DA:
Fixes #744

### DIFF
--- a/core/ledger-client/src/ledger-client.ts
+++ b/core/ledger-client/src/ledger-client.ts
@@ -502,7 +502,7 @@ export class LedgerClient {
     }
 
     /**
-     * @deprecated use postWithRetry instead, should be made private
+     * @deprecated use getWithRetry instead, should be made private
      */
     private async get<Path extends GetEndpoint>(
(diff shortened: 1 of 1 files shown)


core/ledger-client/src/ledger-api-utils.ts
```diff
@@ -549,6 +551,116 @@ export async function promiseWithTimeout<T>(
...
+    ],
+}
+
+export async function retryable<T>(
+    fn: () => Promise<T>,
+    retryableOptions: retryableOptions,
+    logger?: Logger
+): Promise<T> {
+    let attempts = 0
+    while (true) {
```
pawelstepien-da: `while (true)` makes me question every possible scenario that could lead to an infinite loop.
I think it would be safer to use here `attempts++ < retryableOptions.retries`, or perhaps use for loop.

PR #721, opened 2025-10-17
feat: logout ux

pixelplex:
Redirect from the /login page for the user who is logged in and trying to go to the /login page.
Similarly, with a user who has logged out and is trying to access pages that are only accessible with a token.
Added an automatic logout for a user whose token has expired.


alexmatson-da (approved):
thanks!

mjuchli-da (approved):

PR #726, opened 2025-10-17
fix: move docker to nx

jordan-da:
* make the `docker` directory a proper nx project
* move the targets to new `docker` project


jordan-da:
taking a different approach

PR #735, opened 2025-10-20
fix: form fields updating

pixelplex:
dynamically update form after authType changing


mjuchli-da:
@alexmatson-da please check. It's intended to fix a bug (not updating the auth form type) I've found in your recent network view changes.

PR #784, opened 2025-10-27
fix(core-ledger-client, wallet-sdk): refactor grant user rights

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/743

We already had a grantRights function that took in arrays of the readAs and actAs partyIds, but I refactored the function to include the `canReadAsAnyParty` and `canExecuteAsAnyParty` and renamed `grantUserRights` (we call this when we allocate a party) to `waitForPartyAndGrantUserRights` since we poll the participant to ensure the party has been created before granting the user rights.

Also, has the fix for https://github.com/hyperledger-labs/splice-wallet-kernel/issues/782 and https://github.com/hyperledger-labs/splice-wallet-kernel/issues/788

### DIFF
--- a/sdk/wallet-sdk/src/ledgerController.ts
+++ b/sdk/wallet-sdk/src/ledgerController.ts
@@ -704,20 +704,21 @@ export class LedgerController {
                 (r) => r.kind.CanReadAs?.value?.party
             )
 
-            let hasActAsRight = false
-            if (this.partyId) {
-                hasActAsRight =
-                    rights.rights?.some(
-                        (r) =>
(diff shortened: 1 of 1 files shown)


mjuchli-da:
@rukmini-basu-da could you make sure that this works in conjunction with `listWallets`? See also  https://github.com/hyperledger-labs/splice-wallet-kernel/issues/782 that was reported to me today. Not sure if it's a bug but my initial investigation pointed to it so I asked for a ticket.

PHOL-DA (approved):

PR #824, opened 2025-11-03
feat: implement creation wallet with fireblocks

pixelplex:
https://github.com/hyperledger-labs/splice-wallet-kernel/issues/610

### DIFF
--- a/wallet-gateway/remote/src/init.ts
+++ b/wallet-gateway/remote/src/init.ts
@@ -119,13 +119,20 @@ export async function initialize(opts: CliOptions, logger: Logger) {
 
     // Provide apiKey from User API in Fireblocks
     const apiPath = path.resolve(process.cwd(), 'fireblocks_api.key')
-    const apiKey = readFileSync(apiPath, 'utf8')
     const secretPath = path.resolve(process.cwd(), 'fireblocks_secret.key')
-    const keyInfo = {
-        apiKey,
-        apiSecret: readFileSync(secretPath, 'utf8'),
+    let apiKey = ''
+    let apiSecret = ''
+
+    if (existsSync(apiPath) && existsSync(secretPath)) {
+        apiKey = readFileSync(apiPath, 'utf8')
+        apiSecret = readFileSync(secretPath, 'utf8')
+    } else {
+        apiKey = 'missing'
+        apiSecret = 'missing'
+        logger.warn('Fireblocks keys files are missing')
     }
 
+    const keyInfo = { apiKey, apiSecret }
     const userApiKeys = new Map([['user', keyInfo]])
 
     const drivers = {


api-specs/openrpc-user-api.json
```diff
@@ -99,20 +99,70 @@
...
-                    "type": "object",
-                    "properties": {
-                        "wallet": {
-                            "type": "object",
-                            "description": "The newly created wallet.",
-                            "$ref": "#/components/schemas/Wallet"
-                        }
-                    },
-                    "required": ["wallet"]
+                    "$ref": "#/components/schemas/Null"
```
mjuchli-da: I think we can and should return a `Wallet`. For signing providers internal and participant, we should leave the code unchanged, and for fireblocks we can still create and return a wallet but just set the status accordingly.

api-specs/openrpc-user-api.json
```diff
@@ -99,20 +99,70 @@
...
+                                "title": "partyHint",
+                                "type": "string",
+                                "description": "The party hint and name of the wallet."
+                            },
+                            "signingProviderId": {
+                                "title": "signingProviderId",
+                                "type": "string",
+                                "description": "The signing provider ID the wallet corresponds to."
+                            },
+                            "txId": {
```
mjuchli-da: externalTxId

core/wallet-store-sql/src/migrations/002-init.ts
```diff
@@ -0,0 +1,63 @@
...
+        .addColumn('tx_id', 'text')
+        .addColumn('transactions', 'text')
+        .execute()
+
+    await sql`
+        INSERT INTO wallets_temp (party_id, "primary", hint, public_key, "namespace", chain_id, signing_provider_id)
+        SELECT party_id, "primary", hint, public_key, "namespace", chain_id, signing_provider_id FROM wallets
+    `.execute(db)
+
+    await db.schema.dropTable('wallets').execute()
```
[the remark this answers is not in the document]
mjuchli-da: Besides, did you think of potentially creating a separate table for the fireblocks related additional information?

core/wallet-dapp-remote-rpc-client/src/index.ts
```diff
@@ -134,19 +140,34 @@ export type Namespace = string
...
+ *
+ */
+export type TopologyTransactions = string
 /**
  *
  * Structure representing a wallet
  *
  */
 export interface Wallet {
+    id: WalletId
```
mjuchli-da: This feels unnecessary since `partyId` is unique...

mjuchli-da (changes requested):

api-specs/openrpc-user-api.json
```diff
@@ -99,20 +99,70 @@
...
-                        }
-                    },
-                    "required": ["wallet"]
+                    "$ref": "#/components/schemas/Null"
                 }
             },
             "description": "Allocates a new wallet and party with the given hint."
         },
+        {
+            "name": "allocateWallet",
```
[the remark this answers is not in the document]
linushkar: I think it's more convenient to separate the logic the way I did now, because they perform slightly different actions. But if you feel it's better to combine them into one, I can refactor it.

core/wallet-store-sql/src/migrations/002-init.ts
```diff
@@ -0,0 +1,63 @@
...
+        )
+        .addColumn('signing_provider_id', 'text', (col) => col.notNull())
+        .execute()
+
+    await sql`
+        INSERT INTO wallets_old (party_id, "primary", hint, public_key, "namespace", chain_id, signing_provider_id)
+        SELECT party_id, "primary", hint, public_key, "namespace", chain_id, signing_provider_id FROM wallets ON conflict do nothing
+    `.execute(db)
+
+    await db.schema.dropTable('wallets').ifExists().execute()
```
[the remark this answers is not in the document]
linushkar: The migration down happens in reverse order. We transition from the new table with id as the primary key back to the old table with party_id as the primary key.

core/wallet-dapp-remote-rpc-client/src/index.ts
mjuchli-da: We are able to form the `partyId` (`hint::namespace`, where `namespace=hash(publicKey)`) by using the public key you select in the controller here: (https://github.com/pixelplex/splice-wallet-kernel/blob/pixelplex/fireblocks-wallet-creation/wallet-gateway/remote/src/user-api/controller.ts#L224)
To create the fingerprint, you can use the `createFingerprintFromKey` method found here: https://github.com/hyperledger-labs/splice-wallet-kernel/blob/b0aab8b949739f0bc437795bef95dd59b7c2b0aa/core/ledger-client/src/topology-write-service.ts#L190

As a result, we can predefine the partyId, before it's transactions are signed by fireblocks and the party is allocated on the participant node.

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -215,26 +219,136 @@ export const userController = (
...
+                const key = keys?.keys?.find((k) => k.name === params.partyHint)
+                if (!key) throw new Error('Fireblocks key not found')
+
+                const { signature, status } = await driver.getTransaction({
+                    userId,
+                    txId: params.txId,
+                })
+
+                if (!['pending', 'signed'].includes(status)) {
+                    await store.removeWallet(params.id)
```
mjuchli-da: We do we remove it here?

core/wallet-store/src/Store.ts
```diff
@@ -60,20 +68,21 @@ export interface Store {
     getPrimaryWallet(): Promise<Wallet | undefined>
     setPrimaryWallet(partyId: PartyId): Promise<void>
     addWallet(wallet: Wallet): Promise<void>
-    // removeWallet(partyId: Wallet): Promise<void>
+    updateWallet(params: UpdateWallet): Promise<void>
```
mjuchli-da: Why don't we just allow updating all fields of a wallet (identified by partyId): 
```
updateWallet(wallet: Wallet): Promise<void>
```

linushkar: Maybe it could be better just rename this method from `updateWallet` to  `updateWalletStatus`?

wallet-gateway/remote/src/ledger/party-allocation-service.ts
```diff
@@ -117,6 +146,17 @@ export class PartyAllocationService {
 
         const signature = await signingCallback(transactions.multiHash)
 
+        if (!signature) {
```
mjuchli-da: I just discussed this with the team and we prefer to have `signingCallback` throw the error and just propagate here. 

When you invoke `party = await partyAllocator.allocateParty` in the controller.ts, you could try-catch and in the error case use the `createFingerprintFromKey(publicKey: string)` to form the namespace...

Would that work?

Issue #580, opened 2025-09-30
[Wallet SDK] buy traffic using external party.


PHOL-DA:
enable the purchase of traffic for external parties.

this can be done with the _`AmuletRules_BuyMemberTraffic`_  (https://github.com/hyperledger-labs/splice/blob/9d1025c0d5016c6040c49981c97d3f61a14edadf/daml/splice-amulet/daml/Splice/AmuletRules.daml#L200)

PHOL-DA:
related to #583

PR #826, opened 2025-11-04
feat: buy member traffic

rukmini-basu-da:
Closes [#580](https://github.com/hyperledger-labs/splice-wallet-kernel/issues/580)

Confirmed reason: "traffic purchase"  is listed under the holding transactions:

```
          {
            "label": {
              "burnAmount": "666.8",
              "mintAmount": "0",
              "type": "Burn",
              "tokenStandardChoice": null,
              "reason": "traffic purchase",
              "meta": {
                "values": {}
              }
            },
```

### DIFF
--- a/docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
+++ b/docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
@@ -11,7 +11,7 @@ import {
 import { pino } from 'pino'
 import { v4 } from 'uuid'
 
-const logger = pino({ name: '05-external-party-setup', level: 'info' })
+const logger = pino({ name: '16-buy-traffic', level: 'info' })
 
 // it is important to configure the SDK correctly else you might run into connectivity or authentication issues
 const sdk = new WalletSDKImpl().configure({


sdk/wallet-sdk/src/ledgerController.ts
```diff
@@ -235,6 +235,7 @@ export class LedgerController {
         timeoutMs: number = 15000
     ): Promise<Types['Completion']['value']> {
         const ledgerEnd = await this.ledgerEnd()
+        this.logger.info(commands)
```
PHOL-DA: i dont think this log command is needed ?

sdk/wallet-sdk/src/tokenStandardController.ts
```diff
@@ -327,6 +327,85 @@ export class TokenStandardController {
...
+     * Build an Exercise command to Buy Member Traffic.
+     * @param dso Party that the sender expects to represent the DSO party of the AmuletRules contract they are calling
+     * @param provider Provider party whose inputs will fund the traffic purchase
+     * @param trafficAmount Amount of traffic to purchase
+     * @param memberId The id of the sequencer member (participant or mediator) for which traffic has been purchased
+     * @param migrationId The migration id of the synchronizer for which this contract tracks purchased extra traffic
+     * @param inputUtxos Optional list of specific holding CIDs to use as inputs.
+     * @returns  AmuletRules_BuyMemberTraffic exercise command and disclosed contracts
+     */
+    async buyMemberTraffic(
```
PHOL-DA: so the principal is that Alice can buy Traffic to the validator operator party using her UTXOS (so an exchange can hold the money on an external party, but use it fuel the validator)

Can we make the naming used here a bit easier.

DSO & migrationId we can fetch from the scan proxy.

Provider should probably be renamed to `Buyer` (just so it is easier to use)

Can we use CCAmount instead of trafficAmount (only if possible), this is more in the way the Consumer is thinking.

docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
```diff
@@ -0,0 +1,220 @@
...
+await sdk.connectTopology(localNetStaticConfig.LOCALNET_SCAN_PROXY_API_URL)
+
+const sender = await sdk.userLedger?.signAndAllocateExternalParty(
+    keyPairSender.privateKey,
+    'alice'
+)
+logger.info(`Created party: ${sender!.partyId}`)
+await sdk.setPartyId(sender!.partyId)
+
+sender?.topologyTransactions!.map((topologyTx) => {
```
PHOL-DA: nit: we don't really need to decode the topology here, do we ?

docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
```diff
@@ -0,0 +1,220 @@
...
+    localNetStaticConfig.LOCALNET_REGISTRY_API_URL
+)
+
+await sdk.setPartyId(receiver?.partyId!)
+const validatorOperatorParty = await sdk.validator?.getValidatorUser()
+
+const instrumentAdminPartyId =
+    (await sdk.tokenStandard?.getInstrumentAdmin()) || ''
+
+await new Promise((res) => setTimeout(res, 5000))
```
PHOL-DA: nit: why do you need await for the promise here ?

docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
```diff
@@ -0,0 +1,220 @@
...
+    v4()
+)
+
+logger.info('transfer pre approval proposal is created')
+
+const participantId = await sdk.userLedger?.getParticipantId()
+
+logger.info(participantId, 'participant id is')
+
+await sdk.setPartyId(validatorOperatorParty!)
```
PHOL-DA: nit: why set party here to change it 2 lines below ?

docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
```diff
@@ -0,0 +1,220 @@
...
+logger.info(buyTrafficCommand)
+
+await sdk.userLedger?.prepareSignExecuteAndWaitFor(
+    buyTrafficCommand,
+    keyPairSender.privateKey,
+    v4(),
+    buyTrafficDisclosedContracts
+)
+
+logger.info(
```
PHOL-DA: what i would really love here to make this perfect is to check the traffic status of the validator operator party to ensure that we now have more traffic.

however that would probably require asking CN to expose `'/v0/domains/{domain_id}/members/{member_id}/traffic-status'` through the scan -proxy or if there is different way using the ledger api.

PHOL-DA (approved):

docs/wallet-integration-guide/examples/scripts/16-buy-traffic.ts
rukmini-basu-da: I've added this as a todo

PR #841, opened 2025-11-10
feat: signing provider store support

pixelplex:
https://github.com/hyperledger-labs/splice-wallet-kernel/issues/562

https://github.com/hyperledger-labs/splice-wallet-kernel/pull/736


mjuchli-da (changes requested):
Very good work, thank you!
There are just some leftovers to be cleaned up and then I think we're good to go.

mjuchli-da (approved):

PR #850, opened 2025-11-12
feat: differentiate between authenticated and connected

mjuchli-da:
image

image

dapp.status() will now return
```json
{
    "jsonrpc": "2.0",
    "id": "6179b028-0ebf-4f56-b488-2989337e625a",
    "result": {
        "kernel": {
            "id": "remote-da",
            "clientType": "remote",
            "url": "http://localhost:3030/api/v0/dapp",
            "userUrl": "http://localhost:3030"
        },
        "isConnected": true,
        "isNetworkConnected": false,
        "networkReason": "Ledger unreachable: fetch failed",
        "networkId": "canton:local-oauth"
    }
}
```

`isConnected` means the user is logged into the Wallet Gateway. 
`isNetworkConnected` means the ledger api is reachable.

### DIFF
--- a/api-specs/openrpc-dapp-remote-api.json
+++ b/api-specs/openrpc-dapp-remote-api.json
@@ -181,9 +181,6 @@
                     "title": "OnConnectedEvent",
                     "type": "object",
                     "properties": {
-                        "kernel": {
-                            "$ref": "#/components/schemas/KernelInfo"
-                        },
                         "status": {
                             "$ref": "#/components/schemas/StatusEvent"
                         },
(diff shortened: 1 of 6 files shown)


core/ledger-client/src/ledger-client.ts
```diff
@@ -676,10 +676,7 @@ export class LedgerClient {
         return this.acsHelper.getCacheStats()
     }
 
-    /**
-     * @deprecated use postWithRetry instead, should be made private
-     */
-    private async post<Path extends PostEndpoint>(
+    public async post<Path extends PostEndpoint>(
```
alexmatson-da: why undeprecate/unprivate these methods? Any reason not to use the retryable equivalents for the status check?

PR #855, opened 2025-11-12
feat: multiple splice versions

pawelstepien-da:


.github/workflows/build.yml
```diff
@@ -114,14 +114,14 @@ jobs:
             - uses: ./.github/actions/setup_yarn
 
             - name: Download Canton
-              run: yarn script:fetch:canton
+              run: yarn script:fetch:canton --network=mainnet
```
PHOL-DA: i would suggest to add a comment here to remove the network enforcement once we have upgraded the bootstrap script to support canton 3.4

docs/wallet-integration-guide/examples/snippets/read-active-contracts.ts
```diff
@@ -15,8 +28,9 @@ export default async function () {
     const myParty = global.EXISTING_PARTY_1
     const offset = (await sdk.userLedger!.ledgerEnd()!).offset
     //we use holdings as an example here
-    const myTemplateId =
-        '#canton-builtin-admin-workflow-ping:Canton.Internal.Ping:Ping'
+    const version = sdk.userLedger!.getCurrentClientVersion()
+    const isNewTemplate = greaterOrEqualVersion(version, '3.4')
+    const myTemplateId = isNewTemplate ? PING_NEW : PING_OLD
```
PHOL-DA: while this is a cool way to solve it would probably just change the templateId to use the string for holdings :sweat_smile:

.github/workflows/build.yml
```diff
@@ -243,9 +251,16 @@ jobs:
               if: failure()
               uses: actions/upload-artifact@v4
               with:
-                  name: docker-logs
+                  name: docker-logs-${{ matrix.network }}
                   path: logs/
 
+    test-wallet-sdk-e2e:
```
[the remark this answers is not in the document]
pawelstepien-da: Other than echo, test-wallet-sdk-e2e has "needs: sdk-e2e" where we have 2 runs for mainnet and devnet (with different names). So it aggregates 2 jobs passing ("sdk-e2e (mainnet)", "sdk-e2e (devnet)") to one "test-wallet-sdk-e2e"
Branch protection rules are for single "test-wallet-sdk-e2e". 
We could remove this aggregation step it if we updated branch protection rules to expect something like both:
test-wallet-sdk-e2e (mainnet)
test-wallet-sdk-e2e (devnet)

Issue #857, opened 2025-11-12
Can't run script test examples


OscBacon:
I started the localnet with `yarn start localnet`, when I curl `wallet.localhost:2000`, I get a reply, but when I run `yarn script:test:examples`, I get the following error:
```
❯ yarn script:test:examples

=== Executing script: 02-auth-localnet.ts ===
(node:24622) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
[18:54:15.299] INFO (02-auth-localnet/24622): SDK initialized
[18:54:15.303] INFO (02-auth-localnet/24622): Connected to ledger
node:internal/deps/undici/undici:13510
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13510:13
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async coreFetch (/home/baracoso/.yarn/berry/cache/openapi-fetch-npm-0.14.0-20e667e085-10c0.zip/node_modules/openapi-fetch/src/index.js:162:20)
    at async _ScanProxyClient.get (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/core/splice-client/src/scan-proxy-client.ts:133:22)
    at async _ScanProxyClient.fetchAmuletRulesOnce (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/core/splice-client/src/scan-proxy-client.ts:153:22)
    at async _ScanProxyClient.getAmuletRules (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/core/splice-client/src/scan-proxy-client.ts:177:26)
    at async _ScanProxyClient.getAmuletSynchronizerId (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/core/splice-client/src/scan-proxy-client.ts:306:29)
    at async WalletSDKImpl.connectTopology (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/sdk/wallet-sdk/src/index.ts:247:17)
    at async <anonymous> (/home/baracoso/projects/hyperledger-labs/splice-wallet-kernel/docs/wallet-integration-guide/examples/scripts/02-auth-localnet.ts:28:1) {
  [cause]: Error: getaddrinfo ENOTFOUND wallet.localhost
[...]

PR #858, opened 2025-11-12
fix: simplify localnet proxy domain to localhost

OscBacon:
Avoids any DNS issues in Node, and modifying /etc/hosts
See #857


OscBacon:
@PHOL-DA are you an authorized reviewer?

PR #877, opened 2025-11-18
fix(core-splice-provider): sockets can't be serialized

PHOL-DA:
This allows one to json serialize the provider (for instance as part of a log statement) without getting a nasty error.

### DIFF
--- a/core/splice-provider/src/SpliceProviderHttp.ts
+++ b/core/splice-provider/src/SpliceProviderHttp.ts
@@ -53,7 +53,7 @@ export class SpliceProviderHttp extends SpliceProviderBase {
 
         // Prevent serialization of the socket
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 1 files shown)


core/splice-provider/src/SpliceProviderHttp.ts
```diff
@@ -51,6 +51,10 @@ export class SpliceProviderHttp extends SpliceProviderBase {
 
         this.socket = this.openSocket(url)
 
+        // Prevent serialization of the socket
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        ;(this.socket as any).toJson = () => ({})
```
pawelstepien-da: I thought it's `toJSON` with all capitals on JSON.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify
Were you able to test that serialization when `this.socket` contains the socket object and it worked as expected?

PR #895, opened 2025-11-20
feat: connect to localnet (easily)

jaspervdj-da:
1. Made `synchronizerId` optional in most places.  It is retrieved from the validator app if not specified.
2. Added a `localnet` configuration.
3. Changed the example dApp to retrieve the Ledger version, and toggle the ping contract based on that.

### DIFF
--- a/api-specs/openrpc-user-api.json
+++ b/api-specs/openrpc-user-api.json
@@ -464,6 +464,17 @@
             }
         },
         {
+            "name": "removeSession",
+            "params": [],
+            "result": {
+                "name": "result",
+                "schema": {
+                    "$ref": "#/components/schemas/Null"
+                }
+            },
+            "description": "Removes the current network session."
+        },
+        {
             "name": "listSessions",
             "params": [],
             "result": {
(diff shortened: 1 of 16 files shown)


shell.nix
```diff
@@ -0,0 +1,22 @@
+{
```
jaspervdj-da: Will cull this from the branch

wallet-gateway/remote/src/ledger/party-allocation-service.ts
```diff
@@ -21,20 +21,30 @@ type SigningCbFn = (hash: string) => Promise<string>
...
+    private logger: Logger
     private ledgerClient: LedgerClient
-
-    constructor(
-        private synchronizerId: string,
-        accessTokenProvider: AccessTokenProvider,
-        httpLedgerUrl: string,
-        private logger: Logger,
-        accessToken: string = ''
-    ) {
```
jaspervdj-da: Making `synchronizerId` optional here would require me to reorder the positional parameters, so I turned this into named parameters.

PR #903, opened 2025-11-20
feat: merge utxos with delegate

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/847

### DIFF
--- a/sdk/wallet-sdk/src/tokenStandardController.ts
+++ b/sdk/wallet-sdk/src/tokenStandardController.ts
@@ -263,7 +263,7 @@ export class TokenStandardController {
      * @param includeLocked defaulted to true, this will include locked UTXOs.
      * @param limit optional limit for number of UTXOs to return.
      * @param offset optional offset to list utxos from, default is latest.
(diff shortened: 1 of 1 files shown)


sdk/wallet-sdk/src/tokenStandardController.ts
```diff
@@ -529,6 +531,260 @@ export class TokenStandardController {
...
+        [WrappedCommand<'ExerciseCommand'>, Types['DisclosedContract'][]]
+    > {
+        //determine if user has more than 10 Holding UTXOS
+
+        const ledgerEnd = await this.client.get('/v2/state/ledger-end')
+
+        const utxos = await this.listHoldingUtxos(
+            true,
+            100,
+            ledgerEnd.offset,
```
PHOL-DA: the offset is optional, so you could just remove this and line 584 and just pass down a null or undefined value.

PHOL-DA (approved):
some nit picking, but otherwise looks good

PR #919, opened 2025-11-25
feat(core-ledger-client): browser support

jaspervdj-da:
I want to be able to use the `@canton-network/core-ledger-client` in the browser in order to build the portfolio UI.

It has a bunch of useful utilities (tapping, ACS helper, type definitions...) that would be silly to replicate.

Two things prevented me from getting this to work:

1. `TopologyWriteService` uses local filesystem and `crypto` modules.  I split this off to a separate package `@canton-network/core-topology-write-service`, that's a bit of a mouthful though, maybe `@canton-network/core-topology-client` is better?  Suggestions welcome.

2. `submitExerciseCommand` was not used (in this repo) and also used `crypto` modules.  I deleted this function and its unexported dependencies.

This probably requires some version bumps since it is a breaking change, so opening this as a draft PR first.


core/topology-write-service/package.json
```diff
@@ -0,0 +1,62 @@
+{
+    "name": "@canton-network/core-topology-write-service",
+    "version": "0.20.0",
```
PHOL-DA:
```suggestion
    "version": "0.1.0",
```

PHOL-DA (approved):
LGTM

PHOL-DA (approved):

PR #942, opened 2025-11-28
fix(wallet-gateway-remote): store all tx params

jaspervdj-da:
In case we are dealing with participant signing, we submit the transaction using `/submit-and-wait`.  In contrast to the prepare/sign workflow of external signing, this needs all parameters (including, in particular, `.disclosedContracts`).

This means we need this info stored.  This PR modifies the semantics of `payload` to include all parameters rather than just `.commands`.

### DIFF
--- a/wallet-gateway/remote/src/user-api/controller.ts
+++ b/wallet-gateway/remote/src/user-api/controller.ts
@@ -23,7 +23,6 @@ import {
     Null,
     ListTransactionsResult,
 } from './rpc-gen/typings.js'
-import { PrepareExecuteParams } from '../dapp-api/rpc-gen/typings.js'
 import {
     Store,
     Transaction,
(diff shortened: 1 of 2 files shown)


wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -470,21 +470,17 @@ export const userController = (
...
-                        synchronizerId,
-                        packageIdSelectionPreference: [],
-                    }
                     try {
                         const res = await ledgerClient.postWithRetry(
                             '/v2/commands/submit-and-wait',
-                            request
+                            {
+                                synchronizerId,
+                                packageIdSelectionPreference: [],
```
mjuchli-da: We should assume that this and other fields can be present in `transaction.payload`.  
In fact, I think this bit here should 1:1 replicate (or rather reuse the same code) how a transaction is being prepared on the dApp API level, so that there is no ambiguity. 
See here: https://github.com/hyperledger-labs/splice-wallet-kernel/blob/96ebefb2f760086254f930579b7707bb66d109c2/wallet-gateway/remote/src/dapp-api/controller.ts#L270

mjuchli-da (commented):
My current understanding is that this re-implements what already exists in the dApp controller. Assuming that is true, we should reuse the same logic.

wallet-gateway/remote/src/user-api/controller.ts
jaspervdj-da: I pushed an update to this to factor out this code, is this along the lines of what you had in mind?

mjuchli-da (approved):

PR #948, opened 2025-12-01
feat: tx signedAt and createdAt

### LINKED ISSUE #939: Add timestamp to `transaction`
In the `store` we maintain a record of the transactions that are to be signed or have been signed in the past.
We currently lack an audit trail for these transactions and therefore should extend the transaction with the following two timestamps:
- createdAt
- signedAt (optional)

Also present these dates on the User UI pages /transactions and /approve/{id}.

pawelstepien-da:
For transactions in store add signedAt and createdAt and reflect this on UI.
Additionally fix bug with setSigningKeys and setSigningTransactions conflicts resolutions.
resolves #939


wallet-gateway/remote/src/web/frontend/approve/index.ts
```diff
@@ -244,6 +248,11 @@ export class ApproveUi extends LitElement {
                 <h3>Status</h3>
                 <p>${this.status}</p>
 
+                <h3>Created At</h3>
```
PHOL-DA: i would prefer to do something like:
```
{this.createdAt && (
  <>
    <h3>Created At</h3>
    <p>{this.createdAt}</p>
  </>
)}
{this.signedAt && (
  <>
    <h3>Signed At</h3>
    <p>{this.signedAt}</p>
  </>
)}
```

pawelstepien-da: Very well. I made those label-value pairs not render if there is no value now.

mjuchli-da (approved):

PR #982, opened 2025-12-04
refactor(core-ledger-client): split up token standard service

jaspervdj-da:
The `TokenStandardService` currently depends on the scan client and/or the scan proxy client.

However, in the token standard definition, there is nothing in the token standard that refers to scan APIs.  More practically, I would also like to be able to run the token standard portfolio without the user (or app provider) having to configure scan URLs.

This refactor puts us on the right path for that, it removes the following non-standard concerns into a separate `AmuletService`:

- Tapping
- Transfer preapprovals (this may become part of the token standard later)
- Featured apps

Note that this only affects the core libraries.  We could do a similar split in the Wallet SDK (but that would be a breaking change).

I picked the shortest name, we could also consider:

- `AmuletTokenService`
- `SpliceTokenService`
- `CantonTokenService`
- `CantonCoinService`

### DIFF
--- a/core/ledger-client/src/amulet-service.ts
+++ b/core/ledger-client/src/amulet-service.ts
@@ -254,7 +254,6 @@ export class AmuletService {
         return [exercise, disclosed]
     }
 
-    // TODO(#583) as it's not a part of token standard, should be moved somewhere else
     async createTap(
         receiver: string,
         amount: string,


PHOL-DA (approved):
Me love non-breaking changes :heart_hands:

PR #984, opened 2025-12-04
refactor(core-ledger-client): remove access token provider

jaspervdj-da:
This is a follow-up to #982 that removes the access token provider from the token standard service, where we don't need it.


jaspervdj-da:
This doesn't work, I don't fully understand why but I'll close it for now.

PR #1003, opened 2025-12-08
feat(wallet-gateway-remote): added connectivity status to login

PHOL-DA:
**Motivation:**
This PR introduces background runners and caching to the wallet gateway that allows us to (in memory) continuously maintain the status of ledger endpoints before authentication. This will become a lot handier further down the road when we split idp and network completely.

**considerations:**
as @mjuchli-da have mentioned in comments for persistence and scaling reason this should be in cache / memory only.
for now this is a proof of concept that delivers also a short term value (in the form of the status icon), that creates a basis for further similar works.

**With nothing running locally:**
[Screencast from 2025-12-11 08-48-10.webm](https://github.com/user-attachments/assets/7e812423-20ad-40b2-b492-45da92ea8755)

**With `start:canton` & `start:localnet`**
[Screencast from 2025-12-11 08-51-15.webm](https://github.com/user-attachments/assets/65364f0f-32f9-4ddd-aa26-b1a1c68cb4b8)

### DIFF
--- a/api-specs/openrpc-user-api.json
+++ b/api-specs/openrpc-user-api.json
@@ -561,11 +561,14 @@
                         "$ref": "#/components/schemas/Network"
                     },
                     {
+                        "title": "VerifiedNetworkProperties",
+                        "description": "holder for boolean verified property",
                         "type": "object",
                         "properties": {
                             "verified": {
-                                "type": "boolean",
-                                "description": "check if connectivity to the ledger node could be established"
+                                "title": "verified",
+                                "description": "true if we could succesfully call the ledger version endpoint",
+                                "type": "boolean"
                             }
                         },
                         "required": ["verified"]
--- a/core/wallet-user-rpc-client/src/index.ts
+++ b/core/wallet-user-rpc-client/src/index.ts
@@ -215,12 +215,17 @@ export type Signature = string
 export type SignedBy = string
 /**
  *
- * check if connectivity to the ledger node could be established
+ * true if we could succesfully call the ledger version endpoint
  *
  */
-export type BooleanYOuCSU8N = boolean
-export interface ObjectOfBooleanYOuCSU8NKjcqLmyA {
-    verified: BooleanYOuCSU8N
+export type Verified = boolean
+/**
+ *
+ * holder for boolean verified property
+ *
+ */
+export interface VerifiedNetworkProperties {
+    verified: Verified
     [k: string]: any
 }
 /**
@@ -228,7 +233,7 @@ export interface ObjectOfBooleanYOuCSU8NKjcqLmyA {
  * Structure representing the Networks with connectivity verification
  *
  */
-export type VerifiedNetwork = Network & ObjectOfBooleanYOuCSU8NKjcqLmyA
+export type VerifiedNetwork = Network & VerifiedNetworkProperties
 export type Networks = VerifiedNetwork[]
 export type Idps = Idp[]
 /**
--- a/core/wallet-user-rpc-client/src/openrpc.json
+++ b/core/wallet-user-rpc-client/src/openrpc.json
@@ -561,11 +561,14 @@
                         "$ref": "#/components/schemas/Network"
                     },
                     {
+                        "title": "VerifiedNetworkProperties",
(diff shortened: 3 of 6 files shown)


mjuchli-da:
@PHOL-DA I had a think about persistence/scalability. I agree, but this should not go into our domain models, but instead in a separate 'cache'.

wallet-gateway/remote/src/user-api/server.ts
```diff
@@ -23,14 +24,16 @@ export const user = (
     userUrl: string,
     notificationService: NotificationService,
     drivers: Partial<Record<SigningProvider, SigningDriverInterface>>,
-    store: Store & AuthAware<Store>
+    store: Store & AuthAware<Store>,
+    networkCacheStore: NetworkCacheStore
```
alexmatson-da: I think calling this a `networkCacheStore` might be slightly misleading since it doesnt implement the Store interface. How about just `NetworkCache`?

core/wallet-store/src/config/schema.ts
```diff
@@ -41,6 +41,11 @@ export const storeConfigSchema = z.object({
     networks: z.array(networkSchema),
 })
 
+export const verifiedNetworkSchema = networkSchema.extend({
```
mjuchli-da: Hm, I don't think this has anything to do with the `wallet-store`. Can we define it directly in the `wallet-gateway`?

wallet-gateway/remote/src/cache/network-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+const NetworkCache = new LRUCache<string, VerifiedNetwork>({
+    maxSize: DEFAULT_MAX_CACHE_SIZE,
+    entryExpirationTimeInMS: DEFAULT_ENTRY_EXPIRATION_TIME,
+})
+
+export class NetworkCacheStore {
+    constructor(
+        private readonly store: Store,
+        private readonly logger: Logger
+    ) {}
```
alexmatson-da: Do you need all this? I'm wondering why not decouple the cache from the underlying store?

The network cache can just store `{ networkId: string, verified: boolean }`, and only needs a getter and setter.

In the controller you retrieve the latest status from the store directly, and merge based on ID to include the verified status

wallet-gateway/remote/src/cache/network-cache.ts
```diff
@@ -0,0 +1,75 @@
...
+    Store,
+    Network,
+    VerifiedNetwork,
+} from '@canton-network/core-wallet-store'
+import { Logger } from 'pino'
+
+export const DEFAULT_MAX_CACHE_SIZE = 100
+export const DEFAULT_ENTRY_EXPIRATION_TIME = 10 * 60 * 1000
+
+const NetworkCache = new LRUCache<string, VerifiedNetwork>({
```
mjuchli-da: Why not simply cache the status for each network `NetworkStatusCache = LRU<NetworkId, NetworkStatus>`?

With 
```
NetworkStatus = { connected: boolean, reason?: string, lastChecked: Date }
```

This way, when listing existing networks, nothing needs to change, except for an additional cache lookup to get the status.

api-specs/openrpc-user-api.json
```diff
@@ -553,6 +553,28 @@
...
+                "allOf": [
+                    {
+                        "$ref": "#/components/schemas/Network"
+                    },
+                    {
+                        "title": "VerifiedNetworkProperties",
+                        "description": "holder for boolean verified property",
+                        "type": "object",
+                        "properties": {
+                            "verified": {
```
alexmatson-da: very nit-picky, but I'm not sure how I feel about the name `verified` for this property. I think `active`, `live`, or `reachable` maybe gives the idea better?

mjuchli-da (changes requested):
As mentioned in the comments, I believe it would be better to hold only the cached properties (status in this case) in the LRU. Then use this cache here needed: which is only when listing networks. This way, we have much better decoupling between the domain model (`Network`) and a (single) cached property (`Status`) that is not part of the domain model.

PR #1008, opened 2025-12-09
feat(core-wallet-ui-components): persist last used custom kernel in discovery

### LINKED ISSUE #1006: Add a "last used" gateway option to discovery popup
The discovery popup can store the last used custom URL in localstorage to remember for next time. This will be scoped only to the dapp origin, but should still provide a UX benefit to users reusing the same custom gateway+dapp combo

alexmatson-da:
https://github.com/user-attachments/assets/e2173b10-7e81-4a5f-b1a4-b21c13befc8b

Fixes #1006

### DIFF
--- a/core/wallet-ui-components/src/components/Discovery.ts
+++ b/core/wallet-ui-components/src/components/Discovery.ts
@@ -140,7 +140,11 @@ export class Discovery extends HTMLElement {
         )
 
         if (cachedLastUsed) {
-            this.lastUsed = JSON.parse(cachedLastUsed) as KernelType
+            try {
+                this.lastUsed = JSON.parse(cachedLastUsed) as KernelType
+            } catch {
+                localStorage.removeItem(this.DISCOVERY_LAST_USED_KEY)
+            }
         }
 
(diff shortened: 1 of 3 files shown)


core/wallet-ui-components/src/components/Discovery.ts
```diff
@@ -132,6 +135,14 @@ export class Discovery extends HTMLElement {
             )
         }
 
+        const cachedLastUsed = localStorage.getItem(
+            this.DISCOVERY_LAST_USED_KEY
+        )
+
+        if (cachedLastUsed) {
+            this.lastUsed = JSON.parse(cachedLastUsed) as KernelType
```
pawelstepien-da: Maybe it's a bit far-fetched, but what do you think about wrapping JSON.parse in try catch block, so in case of somehow malformed localStorage entry, we don't get runtime error in the middle of constructor? Especially that before this line we communicate to window.opener that things went well. 
Maybe also clear this localStorage entry in case of error in parsing.

mjuchli-da (approved):
nice!!!

PR #1016, opened 2025-12-11
feat(wallet-gateway-remote): determine signing provider on wallet sync

pawelstepien-da:


wallet-gateway/remote/src/ledger/wallet-sync-service.ts
```diff
@@ -31,6 +40,130 @@ export class WalletSyncService {
...
+              signingProviderId: Exclude<
+                  SigningProvider,
+                  SigningProvider.PARTICIPANT
+              >
+              publicKey: string
+          }
+    > {
+        try {
+            // If isLocal is false, skip participant check (definitely external party)
+            // If isLocal is true or undefined, check participant namespace first
```
alexmatson-da: its poorly named, but IIUC `isLocal` has no bearing on external vs internal party. From the ledger OpenAPI spec

> isLocal:
> true if party is hosted by the participant and the party shares the same identity provider as the user issuing the request.

both internal and external parties are hosted on the participant

PR #1034, opened 2025-12-17
feat(wallet-gateway-remote): expired token handling

pawelstepien-da:
-Handle redirects depending on token being valid state:
  -If logged in, redirect from /login to default, or intended page
  -If not logged in, redirect from protected pages to login

-Save intended page in case of automatic logout, which happens either by timeout based on token expiration time, or when getting 401 because token is either invalid, or expired. When logging in again, UI will redirect to intended page.

-Always try to remove session on logout.

-Fixed token expiration timeout triggering logout that was never cleared - before if we had token that is valid for 15 mins more, log out, log in with token valid for 60min, we would still get logged out in 15 mins.

### DIFF
--- a/wallet-gateway/remote/src/web/frontend/constants.ts
+++ b/wallet-gateway/remote/src/web/frontend/constants.ts
@@ -1,4 +1,4 @@
-// Copyright (c) 2025 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


wallet-gateway/remote/src/web/frontend/constants.ts
```diff
@@ -0,0 +1,24 @@
...
+    '/transactions',
+    '/approve',
+    '/',
+    '/404',
+] as const
+
+export type AllowedRoute = (typeof ALLOWED_ROUTES)[number]
+
+export function isAllowedRoute(path: string): path is AllowedRoute {
+    if (path.startsWith('/callback')) {
```
PHOL-DA: i assume the reason for /callback having a unique case is because it contains query string parameters? Should we simply not allow query string parameter handling for all the allowed routes ?

PHOL-DA (approved):

PR #1049, opened 2025-12-18
fix: stop double popup for prepareExecute

alexmatson-da:
we were opening the popup window for prepareExecute twice, once in the SpliceProviderHttp implementation, and once inside the SDK's provider wrapper.

Before this didnt cause any issues, because both popups had the same window `target` name. In that case, the 2nd call simply refocused the window opened by the 1st.

But now the gateway UI uses a different `target`, which resulted in two windows.

### DIFF
--- a/sdk/dapp-sdk/src/ui.ts
+++ b/sdk/dapp-sdk/src/ui.ts
@@ -9,8 +9,15 @@ import {
 import { popupHref } from '@canton-network/core-wallet-ui-components'
 
 class GatewayUi {
-    private identifier: string = 'splice_wallet_gateway_popup'
     private popup: WindowProxy | undefined = undefined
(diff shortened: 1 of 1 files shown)


sdk/dapp-sdk/src/ui.ts
```diff
@@ -19,45 +26,32 @@ class GatewayUi {
...
-                '',
-                this.identifier,
-                'popup width=100,height=100'
-            )
-
-            if (existing && !existing.closed) {
-                existing.close()
-            }
-        }
-
```
alexmatson-da: this approach would sometimes lead to a flash of an empty popup, which I thought looked bad. 

Instead, I decided to automatically close the popup if the dApp page is refreshed / closed via the `beforeunload` event

PR #1055, opened 2025-12-18
fix: update transaction after execute with internal party

mjuchli-da:


alexmatson-da (approved):
we should probably add an internal party execution scenario to the e2e test 🤦 

thanks!

PR #1088, opened 2026-01-06
feat(example-portfolio): add new header with connection and gateway triggers

fayi-da:
- [ ] Add screenshots/screenrecording

Initial view
Screenshot 2026-01-06 at 15 31 54

After connecting
Screenshot 2026-01-06 at 15 32 15

Can disconnect from the menu once connected.
Screenshot 2026-01-06 at 15 32 23


examples/portfolio/src/lib/theme-context.tsx
```diff
@@ -0,0 +1,18 @@
+import { createContext, useContext } from 'react'
```
jaspervdj-da: I've been using `src/contexts` for other Context/Provider pairs, should we standardize on using `lib/`?  I don't have a strong preference.

PR #1089, opened 2026-01-06
docs: add a section on high availability and disaster recovery

davidrichards-da:


docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
```diff
@@ -0,0 +1,73 @@
...
+----------------------------
+High Availability (HA) Setup
+----------------------------
+
+High Availability refers to the ability to continue operations during minor failures, such as a single node outage.
+
+Recommended Architecture
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+* **Redundant Validators**: Run 2 validators behind a single gateway.
```
PHOL-DA: do you mean a wallet gateway or a load balancer ?

docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
```diff
@@ -0,0 +1,73 @@
...
+* **Confirming Rights**: Host parties on both validators with confirming rights.
+* **Threshold Configuration**: Implement a confirming threshold of 1/2. This ensures that if one validator goes offline, the remaining node can still authorize transactions.
```
PHOL-DA: could you add a code snippet showcasing how to do this using the Wallet SDK ?

docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
```diff
@@ -0,0 +1,73 @@
...
+
+To further harden your HA setup, consider distributing your validator hosting across:
+
+* **Availability Zones**: Different AZs within the same cloud region.
+* **Cloud Regions**: Different cloud regions.
+* **Multi-Cloud**: Different cloud providers to mitigate provider-wide outages.
+* **Node Scaling**: Host more validators with a low confirming threshold.
+
+----------------------
+Disaster Recovery (DR)
```
PHOL-DA: would it not make more sense to simply link to:
https://docs.dev.sync.global/validator_operator/validator_disaster_recovery.html

you seem to be linking to it in all three subsections anyway.

PHOL-DA (commented):
Im hesitant about adding "HA and DR" to wallet integration section, this seems more like something that should be in the operate sections (and we have most of the cases covered in the exchange integration already, but under different naming)

docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
davidrichards-da: I can't. From what I can see we don't set a confirming threshold in our multi-hosting parties example either.

docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
```diff
@@ -0,0 +1,73 @@
+=======================================
+High Availability and Disaster Recovery
```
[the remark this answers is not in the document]
davidrichards-da: Could, but as discussed in the meeting, won't

docs/wallet-integration-guide/src/high-availability-and-disaster-recovery/index.rst
```diff
@@ -0,0 +1,73 @@
...
+Resilience Setup
+----------------
+
+Resilience refers to the ability to continue operations during minor failures, such as a single node outage.
+
+Recommended Architecture
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+* **Redundant Validators**: Run 2 validators behind a load balancer.
+* **Confirming Rights**: Host parties on both validators with confirming rights.
```
bame-da: This probably refers to the hosted parties, but there's also the matter of the provider party for CC preapprovals. Unfortunately the renewal automation only works if the preapproval provider is a node admin party. And if the node admin party is down, the preapproval doesn't work anymore so incoming transfers time out.

So the node admin party of one node has to be replicated to another node in confirming mode, which is a thus far undocumented procedure.

PR #1099, opened 2026-01-08
docs: added docs for wallet gateway configuration

PHOL-DA:


docs/wallet-gateway/src/configuration/index.rst
```diff
@@ -0,0 +1,163 @@
...
+Default configuration example against a localnet
+------------------------------------------------
+
+Here is a minimalistic configuration example that can be used against a
+splice localnet using in-memory stores and an unneeded mock auth setup.
+The mock auth is only there to showcase how an idp configuration could look like.
+
+.. literalinclude:: ../../examples/json/default-config.json
+    :language: json
+    :dedent:
```
alexmatson-da: I think worth mentioning somewhere that, via the CLI, you can generate a example config (which will always be up-to-date), as well as a JSON schema of all possible values:

1. Example config

```
npx @canton-network/wallet-gateway-remote@latest --config-example
```

(writes to stdout, can redirect into a file)

2. JSON Schema

```
npx @canton-network/wallet-gateway-remote@latest --config-schema
```

The JSON Schema can also include field-specific documentation that we will probably want to mirror in there

mjuchli-da (approved):

PR #1101, opened 2026-01-08
feat(wallet-gateway-remote): add public URL config option, deprecate host/tls options

alexmatson-da:


wallet-gateway/remote/src/config/Config.ts
```diff
@@ -16,12 +20,36 @@ export const kernelInfoSchema = z.object({
...
+    userPath: z.string().default('/api/v0/user').meta({
+        description: 'The path serving the user API. Defaults /api/v0/user',
+    }),
+    allowedOrigins: z
+        .union([z.literal('*'), z.array(z.string())])
+        .default('*')
+        .meta({
+            description:
+                'Allowed CORS origins, typically corresponding to which external dApps are allowed to connect. Use "*" to allow all origins, or set an array of origin strings.',
+        }),
```
alexmatson-da: @PHOL-DA , if we set up a [sphinx-jsonschema](https://sphinx-jsonschema.readthedocs.io/en/latest/installation.html) plugin, then we can keep the config file documentation close to the source and include these descriptions in the docs

PR #1112, opened 2026-01-12
feat: store disabled wallets on sync error

pawelstepien-da:
Add columns in wallets table `disabled` and `reason`.
When syncing wallets, if parties signing provider can't be figured out, instead of rejecting create a row, but make it disabled and with reason.
When syncing wallets, if there is already disabled wallet in db, reattempt to match it with signing provider.
In UI, indicate which wallets are disabled, show reason field, block "set at primary" button.


core/wallet-store-sql/src/store-sql.ts
```diff
@@ -80,14 +81,26 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
...
+                    userId: table.userId,
                     externalTxId: table.externalTxId ?? '',
                     topologyTransactions: table.topologyTransactions ?? '',
                     status: table.status ?? '',
-                })
-            )
+                    disabled: table.disabled ?? 0,
+                }
+                if (table.reason !== undefined) {
+                    walletTable.reason = table.reason
```
mjuchli-da: inline into above? -> i don't think there is a need for const walletTable

pawelstepien-da: Inlined

core/wallet-store-sql/src/schema.ts
```diff
@@ -167,11 +171,30 @@ export const toWalletStatus = (status?: string): WalletStatus => {
...
+            externalTxId: table.externalTxId,
+        }),
+        ...(table.topologyTransactions !== undefined && {
+            topologyTransactions: table.topologyTransactions,
+        }),
     }
+
+    wallet.disabled = table.disabled === 1
+
+    if (table.reason !== undefined) {
```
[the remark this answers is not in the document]
pawelstepien-da: Inlined

PR #1121, opened 2026-01-13
feat(wallet-gateway-remote): enable configurable max payload size

### LINKED ISSUE #1117: Expand upper limits for wallet gateway
currently when using the wallet gateway for signing transaction there seems to be an upper limit of ~ 128 kb of data, this results in a nasty 413 error message when approving the transaction for signing. Since transaction can easily exceed 128 kb we should expand this (it is probably a node.js configuration). 

- [ ] Catch 413 and give a nicer error message.
- [ ] Enable configuration for the upper limit config.

PHOL-DA:
fixes #1117

### DIFF
--- a/core/wallet-ui-components/src/handle-errors.ts
+++ b/core/wallet-ui-components/src/handle-errors.ts
@@ -32,7 +32,7 @@ export function handleErrorToast(e: unknown, fallback?: FallbackType) {
         code = parsed.error.code
         message = parsed.error.message
     }
-    console.log(code)
+
     switch (code) {
         case -32600:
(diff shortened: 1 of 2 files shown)


core/wallet-ui-components/src/handle-errors.ts
```diff
@@ -32,7 +32,7 @@ export function handleErrorToast(e: unknown, fallback?: FallbackType) {
         code = parsed.error.code
         message = parsed.error.message
     }
-
+    console.log(code)
```
pawelstepien-da: leftover

PR #1125, opened 2026-01-13
test(example-portfolio): add two-step transfer E2E test

jaspervdj-da:
This adds a simple E2E test for two-step token standard transfer in the portfolio example.

Adding this E2E exposed a few race conditions in the UI that I resolved.  I left comments on the PR to make it easier to review.

There is a lot of room for improvement and DRY-ing up the test case, but I will take care of that when I add a follow-up test, since it's already getting sizable.


examples/portfolio/src/contexts/ConnectionProvider.tsx
```diff
@@ -74,8 +76,16 @@ export const ConnectionProvider: React.FC<{ children: React.ReactNode }> = ({
...
-        const messageListener = (event: sdk.dappAPI.TxChangedEvent) => {
+        const messageListener = async (event: sdk.dappAPI.TxChangedEvent) => {
             console.log('incoming event', event)
+            if (event.status === 'executed') {
+                await queryClient.invalidateQueries({
+                    queryKey: ['listPendingTransfers'],
+                })
+                await queryClient.invalidateQueries({
+                    queryKey: ['getTransactionHistory'],
+                })
```
jaspervdj-da: @fayi-da Problematically, when we get a notification that transactions have been executed, we can't (easily) tell the party.  That means I'm invalidating these by prefix, but I would prefer to reuse code from `query-keys`.  Should those have prefixed versions as well?

wallet-gateway/remote/src/web/frontend/wallets/index.ts
```diff
@@ -192,6 +189,8 @@ export class UserUiWallets extends LitElement {
                 </button>
             </div>
 
+            ${this.wallets === undefined ? 'Loading wallets…' : ''}
```
jaspervdj-da: @mjuchli-da I made a tiny change to the frontend here to show a message while the wallets haven't been set yet.  This is necessary for playwright to be able to wait on _something_ (since we always render the list).

examples/portfolio/src/contexts/ConnectionProvider.tsx
fayi-da: not sure what you mean by prefix here. Looking at fetch hook, the query key is `['getTransactionHistory', primaryParty]` so your invalidation above will invalidate tx history for all known parties

jaspervdj-da: @fayi-da Right, I want to invalidate it for all parties (since we can't tell from the emitted event which party it applies to).  But this means I can't use the helper from `query-keys`, and I need to hardcode the prefix here.  I was wondering if you knew about a more elegant solution that would still let me use helpers from `query-keys`.

wallet-gateway/remote/src/middleware/rateLimit.ts
```diff
@@ -5,7 +5,7 @@ import rateLimit from 'express-rate-limit'
 
 export const rpcRateLimit = rateLimit({
     windowMs: 1 * 60 * 1000, // 1 minute
-    max: 100, // limit each IP to 100 requests per windowMs
+    max: 1000, // limit each IP to 100 requests per windowMs
```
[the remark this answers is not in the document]
alexmatson-da: I'm planning to expose this through a config file option, I've been encountering 429s frequently with just standard usage in a cluster deployment

examples/portfolio/src/hooks/useCreateTransfer.ts
```diff
@@ -0,0 +1,36 @@
...
+            createTransfer({
+                registryUrls,
+                ...args,
+            }),
+        onSuccess: async (_, args) => {
+            await queryClient.invalidateQueries({
+                queryKey: queryKeys.listPendingTransfers(args.sender),
+            })
+            await queryClient.invalidateQueries({
+                queryKey: queryKeys.listPendingTransfers(args.receiver),
```
fayi-da: Do we actually need to do this? Wouldn't you most likely be transferring to a wallet that's not yours? Also, if you're sending to your own wallet, you'd have to navigate to that wallet's page which will fetch fresh data anyway

wallet-gateway/remote/src/middleware/rateLimit.ts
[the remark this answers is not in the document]
jaspervdj-da: Thanks, I merged in `main` and now I'm modifying the config before the E2E test (also not perfect but better than what I had before): https://github.com/hyperledger-labs/splice-wallet-kernel/blob/e041e9010b9e5484d0eee12ed4dbd50d85a66cc2/.github/workflows/build.yml#L191

examples/portfolio/src/hooks/useCreateTransfer.ts
jaspervdj-da: We currently need it due to the way the E2E test works (directly switching accounts while on the pending transfers page).  We can probably remove it in the future (but also it doesn't really hurt, most of the time the query wouldn't invalidate any data and would be fast, I assume).

.github/workflows/build.yml
```diff
@@ -210,6 +169,54 @@ jobs:
...
+    test-portfolio-e2e:
+        runs-on: ubuntu-latest
+        needs: build
+        steps:
+            - name: Checkout
+              uses: actions/checkout@v4
+            - uses: ./.github/actions/setup_yarn
+
+            - name: Download Canton
+              run: yarn script:fetch:localnet
```
mjuchli-da: Are these cached? `sdk-e2e` uses it too

examples/portfolio/src/contexts/ConnectionProvider.tsx
mjuchli-da: In most cases a wallet would actually be able to tell which party the tx is concerning. Would it help if the `txChanged` event would carry this information?

wallet-gateway/remote/src/web/frontend/wallets/index.ts
mjuchli-da: Nice!

mjuchli-da (approved):
Briefly went over it and concerning the WG it looks good, thanks!
As for pipelines and playwright, @alexmatson-da should review too.

alexmatson-da (approved):
looks good! thx!

PR #1166, opened 2026-01-19
docs: adding a section on generating keys from mnemonic phrases

davidrichards-da:


docs/wallet-integration-guide/src/party-management/index.rst
```diff
@@ -47,6 +47,16 @@ By default an **Ed25519** encryption is used. There exists many libraries that c
    :language: typescript
    :dedent:
 
+Generating Keys from a Mnemonic Phrase (BIP-0039)
```
PHOL-DA: i would not add this section without a code snippet that showcasing how to do it.

Wallet SDK does not currently support Mnemonic phrases out of the box, which mean custom code implementation surrounding it.

PR #1174, opened 2026-01-21
feat: multi network wallets

### LINKED ISSUE #1127: Multi-network support for a wallet
Existing parties that are synced to the Wallet Gateway will have the currently connected `networkId` attached. 
If the user switches to another network which points to the same participant (or possibly different participant with the same party hosted [multi hosting of parties]), then the sync service attempts to create a wallet for the same party, which is correct, but fails with a unique constraint on the `partyId`. 

We should therefore consider spanning the unique constraint across `(partyId, networkId)`. 

Before doing so, we need to uncover possible unwanted side effects this change might have.

pawelstepien-da:
Resolves #1127 

-Allow multiple wallets for the same party in store if they are on different networks
-Primary wallet now can be 1 per network & user combination
-Fixes to wallet sync ui component after moving it to wallets page

### DIFF
--- a/core/signing-store-sql/src/migrations/003-alter-date-fields.ts
+++ b/core/signing-store-sql/src/migrations/003-alter-date-fields.ts
@@ -60,17 +60,9 @@ async function dropConstraintIfExists(
 
 async function dropIndexIfExists(
     db: Kysely<DB>,
-    indexName: string,
-    tableName: string,
-    isPg: boolean
+    indexName: string
 ): Promise<void> {
-    if (isPg) {
-        // PostgreSQL: Drop index by name only (no ON clause)
-        await sql.raw(`DROP INDEX IF EXISTS "${indexName}"`).execute(db)
-    } else {
-        // SQLite: Drop index with ON clause
-        await db.schema.dropIndex(indexName).ifExists().on(tableName).execute()
-    }
+    await sql.raw(`DROP INDEX IF EXISTS "${indexName}"`).execute(db)
 }
 
 export async function up(db: Kysely<DB>): Promise<void> {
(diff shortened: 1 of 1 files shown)


core/wallet-store-inmemory/src/StoreInternal.ts
```diff
@@ -24,6 +24,7 @@ import {
     LedgerClient,
     defaultRetryableOptions,
 } from '@canton-network/core-ledger-client'
+import { CurrentNetworkWalletFilter } from '@canton-network/core-wallet-store/dist'
```
mjuchli-da: -dist

core/wallet-store-sql/src/migrations/006-change-wallet-primary-key-to-composite.ts
```diff
@@ -0,0 +1,221 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Kysely, sql } from 'kysely'
+import { DB } from '../schema.js'
+
+async function isPostgres(db: Kysely<DB>): Promise<boolean> {
```
mjuchli-da: this is helpful. move to separate `utils.ts`

core/wallet-store-sql/src/migrations/007-add-unique-primary-per-network.ts
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Kysely, sql } from 'kysely'
+import { DB } from '../schema.js'
+
+async function isPostgres(db: Kysely<DB>): Promise<boolean> {
```
mjuchli-da: reuse (see other comment)

PR #1178, opened 2026-01-22
feat(core-signing-blockdaemon): added user identifier

PHOL-DA:


core/signing-blockdaemon/src/index.test.ts
```diff
@@ -0,0 +1,68 @@
...
+
+    beforeEach(() => {
+        jest.clearAllMocks()
+
+        mockClient = {
+            signTransaction: jest.fn(),
+        } as unknown as jest.Mocked<SigningAPIClient>
+
+        driver = new BlockdaemonSigningDriver(config)
+        driver.setClient(mockClient)
```
pawelstepien-da: I think you could just overwrite it in test without exposing client setter in the class.
`(driver as unknown).client = mockClient`

PR #1205, opened 2026-01-29
fix(wallet-gateway-remote,dapp-sdk): proper event handling

### LINKED ISSUE #1143: dapp-sdk: switching primary parties does not reflect in dapp

### LINKED ISSUE #1200: When disconnecting, error 0 is thrown into the console
Image

I suppose it could be related to the `events.ts`, where we have `assertProvider()` check on every status event. So on a disconnect event, we do `window.canton = null`, then we emit the event.

PHOL-DA:
List of changes and their impact:
- Removed the `connected` event, all cases have been exchanged for `statusChanged`.
- Added `SPLICE_WALLET_LOGOUT` to postMessage events.
- Added new views to Ping Example for postEvents and windowMessages (for better debugging).
- Proper cleanup for the dApp sdk on `SPLICE_WALLET_LOGOUT` (now have same behavior as disconnect).
- changed `statusChanged` reason from `unauthorized` when `removeSession` or `disconnect` is called.
- dapp.connect() now fire a `StatusChanged` event  (on top of returning it), this allows better and easier hooks.

Fixes #1200 
Fixes #1143

### DIFF
--- a/examples/ping/src/hooks/useConnect.ts
+++ b/examples/ping/src/hooks/useConnect.ts
@@ -36,7 +36,7 @@ export function useConnect(): {
         })
     }
 
-     useEffect(() => {
+    useEffect(() => {
         sdk.status()
             .then(setStatus)
             .catch(() => {
--- a/examples/ping/tests/example.spec.ts
+++ b/examples/ping/tests/example.spec.ts
@@ -64,5 +64,5 @@ test('dApp: execute externally signed tx', async ({ page: dappPage }) => {
     // Wait for command to have fully executed
     //TODO: we use 2 because we have one in the transaction list and one in the event list
     //TODO: fix this so we check each list properly once
-    await expect(dappPage.getByText("executed")).toHaveCount(2)
+    await expect(dappPage.getByText('executed')).toHaveCount(2)
 })


wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -846,7 +846,7 @@ export const userController = (
                 kernel: kernelInfo,
                 isConnected: false,
                 isNetworkConnected: false,
-                networkReason: 'Unauthenticated',
+                networkReason: 'removed session',
```
mjuchli-da: I'd leave it at `Unauthenticated`. We're planning to use enums here.

wallet-gateway/remote/src/dapp-api/controller.ts
```diff
@@ -74,19 +75,22 @@ export const dappController = (
...
-            } as StatusEventAsync
+            }
+            notifier.emit('statusChanged', StatusEvent)
+            return StatusEvent as StatusEventAsync
         },
         disconnect: async () => {
             if (!context) {
                 return null
             } else {
+                console.log('disconnect request')
```
mjuchli-da: Remove console log

PR #1212, opened 2026-01-30
fix: wallet uniqueness by party,network,+user

mjuchli-da:


core/wallet-store-sql/src/migrations/006-change-wallet-primary-key-to-composite.ts
```diff
@@ -190,13 +202,14 @@ export async function down(db: Kysely<DB>): Promise<void> {
...
-        // Use DISTINCT ON to pick first row per party_id
+        // Pick the wallet with the smallest user_id, then network_id as tie-breaker
         await sql`
             DELETE FROM public.wallets w1
             WHERE EXISTS (
               SELECT 1
               FROM public.wallets w2
               WHERE w2.party_id = w1.party_id
+                AND w2.user_id = w1.user_id
                 AND w2.network_id < w1.network_id
```
mjuchli-da: @pawelstepien-da why `<` here?

PR #1224, opened 2026-02-03
fix(wallet-gateway-remote): dApp ui in pop-up window

### LINKED ISSUE #1223: When a dapp call sdk.open() sometimes the dapp itself will populate the UI
This usually happens after the first logon, most likely this is related to the SSE.

PHOL-DA:
Fixes #1223


fayi-da (approved):
thanks

PR #1233, opened 2026-02-04
fix(example-ping): ui rework

### LINKED ISSUE #1232: Status Changed event is fired every time the sdk.open() is called
whenever the sdk.open() is called a statusChanged event is fired, even if the status is the same as before.

PHOL-DA:
small update to the ping example alongside fixing a bug reported.

Add a fixture for the playwright tests to better capture what goes wrong.

also Fixes #1232.

image

### DIFF
--- a/core/wallet-test-utils/src/fixtures.ts
+++ b/core/wallet-test-utils/src/fixtures.ts
@@ -1,0 +1,266 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import {
+    test as base,
+    Page,
+    ConsoleMessage,
+    Request,
+    Response,
+} from '@playwright/test'
+import * as fs from 'fs'
+import * as path from 'path'
+
+interface LogEntry {
+    timestamp: string
+    type: string
+    message: string
+    source?: string
+}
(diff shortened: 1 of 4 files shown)


core/wallet-test-utils/src/fixtures.ts
```diff
@@ -0,0 +1,266 @@
...
+            }
+            networkLogs.push(entry)
+            console.warn(
+                `[${source}] HTTP ${response.status()}: ${response.url()}`
+            )
+        }
+    })
+}
+
+export const test = base.extend<{ page: Page }>({
```
alexmatson-da: whats wrong with using the Playwright GUI? You get all this -- screenshots, console logs, network logs, and more out-of-the-box.

You can even inspect CI runs by downloading the report and opening it through the Playwright UI

alexmatson-da: > add them to the GUI

which GUI do you mean?

Have you ever ran `yarn playwright test --ui`? I've been using this from the beginning to debug a failing playwright test and its really exceptional

# Timeline of screenshots (for all open windows, dapp & popup)

image

# Built-in console inspector

Screenshot 2026-02-04 at 7 14 01 AM



# Built-in network inspector 

including all headers, payloads, and responses

Screenshot 2026-02-04 at 7 14 54 AM

PR #1241, opened 2026-02-04
feat(core-ledger-client,wallet-sdk): add pagination for listing utxos

mateuszpiatkowski-da:


core/ledger-client/src/ledger-client.ts
```diff
@@ -88,6 +88,8 @@ export type PrepareSubmissionResponse =
 // Any options the client accepts besides body/params
 type ExtraPostOpts = Omit<FetchOptions<paths>, 'body' | 'params'>
 
+const ACTIVE_CONTRACTS_MAX_LIMIT = 200
```
PHOL-DA: so this limit is technically configurable on the ledger, the default is just 200. Instead of doing the limit <= ACTIVE_CONTRACTS_MAX_LIMIT, you can just extend the call with an optional boolean (default to false) `continueUntilCompletion`

core/types/src/index.ts
```diff
@@ -32,7 +32,7 @@ export type SuccessResponse = z.infer<typeof SuccessResponse>
 
 export const ErrorResponse = z.object({
     error: z.object({
-        code: z.number(),
+        code: z.union([z.number(), z.string()]),
```
PHOL-DA: i don't believe this is a needed change ? The code should always be a number

core/ledger-client/src/ledger-client.ts
```diff
@@ -604,6 +597,85 @@ export class LedgerClient {
...
+
+        const bodyRequest: PostRequest<'/v2/updates'> = {
+            beginExclusive: 0,
+            endInclusive: ledgerEnd.offset,
+            verbose: false,
+        }
+        if (activeContractsArgs.filter)
+            bodyRequest.filter = activeContractsArgs.filter
+        let currentOffset = 0
+        while (currentOffset < ledgerEnd.offset) {
```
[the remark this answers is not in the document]
PHOL-DA: could probably be a good idea to have one unrelated transactions in your test 20 (create a new party 'Charlie' and have them perform a tap after you have done all your transfers)

PR #1247, opened 2026-02-05
feat(dapp-sdk): improved Provider interfaces

alexmatson-da:
This PR refactors the Provider hierarchy as outlined in the provider architecture doc:

- `Provider<T extends RpcTypes>` interface replaces `SpliceProvider`, and is generic to the specific API methods supported by the provider implementation
- `SpliceHttpProvider` (transport-based naming) has been replaced with `DappRemoteProvider` (API-based naming)
- `SpliceWindowProvider` has been replaced with `DappProvider` (the latter is still transport-agnostic) 
- The wrapper provider exposed by the dapp-sdk has been renamed to `DappSDKProvider` for clarity. It exposes a consistent `Provider<DappRpcTypes>` API

Also includes changes to the client RPC codegen to make easier use for typing the provider(s)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,48 +1,161 @@
-# Wallet Gateway
+# Splice Wallet Kernel
 
-## Project
+A TypeScript framework for building wallet integrations on the [Canton Network](https://www.canton.network/). It provides the **Wallet Gateway** (server and browser extension), the **dApp SDK**, the **Wallet SDK**, and a set of shared core modules.
 
-### Short Description
+## Architecture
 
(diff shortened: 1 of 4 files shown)


core/splice-provider/package.json
```diff
@@ -30,6 +30,7 @@
     "dependencies": {
         "@canton-network/core-rpc-transport": "workspace:^",
         "@canton-network/core-types": "workspace:^",
+        "@canton-network/core-wallet-dapp-remote-rpc-client": "workspace:^",
```
mjuchli-da: why is this needed?

PR #1254, opened 2026-02-06
refactor: core ledger client

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1240

This moves the core-amulet-service, core-token-standard-service, and core-tx-parser modules out of the core-ledger-client.

### DIFF
--- a/core/ledger-client/src/ledger-client.ts
+++ b/core/ledger-client/src/ledger-client.ts
@@ -10,8 +10,7 @@ import {
     defaultRetryableOptions,
     retryable,
     retryableOptions,
-} from './ledger-api-utils'
-
+} from './ledger-api-utils.js'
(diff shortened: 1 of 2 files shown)


core/ledger-client/src/ledger-client.ts
```diff
@@ -10,7 +10,7 @@ import {
     defaultRetryableOptions,
     retryable,
     retryableOptions,
-} from './ledger-api-utils.js'
+} from './ledger-api-utils'
```
pawelstepien-da: We should keep `.js` here

pawelstepien-da (approved):
Looks good! 
I left 2 minor comments.
Also please add new packages to `ecosystem.config.js`

PR #1259, opened 2026-02-09
fix: disable additionalProperties in OpenRPC specs (dapp, dapp-remote, user, & signing)

alexmatson-da:
Results in stricter types for RPC schemas

### DIFF
--- a/api-specs/openrpc-signing-api.json
+++ b/api-specs/openrpc-signing-api.json
@@ -14,6 +14,7 @@
                     "schema": {
                         "type": "object",
                         "title": "SignTransactionParams",
+                        "additionalProperties": true,
                         "properties": {
                             "tx": {
                                 "title": "tx",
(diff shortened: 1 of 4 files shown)


api-specs/openrpc-dapp-api.json
```diff
@@ -187,6 +188,7 @@
                 "title": "Provider",
                 "type": "object",
                 "description": "Represents a Provider.",
+                "additionalProperties": false,
```
mjuchli-da: I think we must allow additional properties for the `dapp-api` so that wallet providers can go beyond what is defined in the dApp API / CIP.

mjuchli-da (commented):
I have some reservations when it comes to the dApp API.
I see it makes sense for the user- and signing APIs.

api-specs/openrpc-dapp-api.json
alexmatson-da: I think a safer approach would be to spec a subproperty for that:

```
{
   id: string,
   version: string,
   ...
   meta: {
      // spec `meta` with `additionalProperties: true`
      customField1: string,
      customField2: string
    }
}
```

That way the root-level of the schema remains "closed"

PR #1317, opened 2026-02-17
feat: ledger command submission sdk v1

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1316

As indicated in the pr, once the LedgerProvider is implemented we will switch to the /executeAndWait interactive submission endpoint. We would've had to make core-ledger-client modifications that would only be temporary since the /executeAndWait path isn't exposed because it's only a 3.4 endpoint.

### DIFF
--- a/sdk/wallet-sdk/src/v1/ledger/types.ts
+++ b/sdk/wallet-sdk/src/v1/ledger/types.ts
@@ -1,0 +1,37 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { PartyId } from '@canton-network/core-types'
+import { Types } from '@canton-network/core-ledger-client'
+import { SignedTransaction } from '../transactions/signed'
+
+export type PrepareOptions = {
+    userId: string
+    partyId: PartyId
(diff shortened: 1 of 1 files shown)


docs/wallet-integration-guide/examples/scripts/v1/01-init.ts
```diff
@@ -31,3 +34,63 @@ const alice = await sdk.party.external
...
+await (
+    await sdk.ledger.prepare({
+        userId,
+        partyId: alice.partyId,
+        commands: pingCommand,
+        disclosedContracts: [],
+    })
+)
+    .sign(aliceKeys.privateKey)
+    .execute({ userId, partyId: alice.partyId })
```
alexmatson-da: I noticed `userId` is passed in explicitly a lot of places. My mental model is that each Wallet SDK instance is 1-to-1 per ledger user, determined by the token used to authenticate. 

Would there be any issues if we removed `userId` params from all public methods, and simply stored it in the wallet context privately?

alexmatson-da (approved):
thanks!

docs/wallet-integration-guide/examples/scripts/v1/01-init.ts
rukmini-basu-da: yeah, that is a good idea, i'll add this in a follow up pr

PR #1326, opened 2026-02-18
feat: add tap command

rukmini-basu-da:
1. Adds asset list to sdk context (This will likely need to be reworked later once we get an official `assets.json` in the api-specs. @alexmatson-da  made a good suggestion to use a Zod schema to define the AssetList structure, which can validate the json file). Created an issue for this here: https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1334 
2. Adds tap command in the amulet namespace (closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1319)
3. Adds list holdings in the token namespace (closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1328)
4. Changes default ledger.execute to wait for command completion (this will be replaced with the executeAndWait endpoint once we switch over to the LedgerProvider)

This is all required for initiating a transfer (related to https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1318)

### DIFF
--- a/sdk/wallet-sdk/src/v1/token/client.ts
+++ b/sdk/wallet-sdk/src/v1/token/client.ts
@@ -11,7 +11,7 @@ import { Holding } from '@canton-network/core-tx-parser'
  * @param includeLocked defaulted to true, this will include locked UTXOs.
  * @param limit optional limit for number of UTXOs to return.
  * @param offset optional offset to list utxos from, default is latest.
- * @param party optional party to list utxos
+ * @param partyId party to list utxos
  * @param continueUntilCompletion optional search the whole ledger for active contracts. Use only when the amount of contracts exceeds the limit defined in http-list-max-elements-limit
  */
 export type ListHoldingsParams = {


docs/wallet-integration-guide/examples/scripts/v1/01-init.ts
```diff
@@ -91,3 +91,33 @@ const signed = SignedTransaction.fromSignature(
...
 
 logger.info('Ping command submitted with offline signing')
+
+const [amuletTapCommand, amuletTapDisclosedContracts] = await sdk.amulet.tap(
+    alice.partyId,
+    '10000'
+)
+
+await (
+    await sdk.ledger.prepare({
```
PHOL-DA: one thing that could be very cool here (not for this PR but maybe something to consider in the future is the either chain or invoke the tap so you get something like:

`sdk.ledger.parepare(sdk.amulet.tap(alice.partyId, '10000')).sign(aliceKeys.privateKey).execute({partyId: alice.partyId})`

docs/wallet-integration-guide/examples/scripts/v1/01-init.ts
```diff
@@ -91,3 +91,33 @@ const signed = SignedTransaction.fromSignature(
...
+        partyId: alice.partyId,
+        commands: amuletTapCommand,
+        disclosedContracts: amuletTapDisclosedContracts,
+    })
+)
+    .sign(aliceKeys.privateKey)
```
PHOL-DA: probably we need to change the signature for .sign so it is more inlign with other places like:

```suggestion
    .sign({privateKey: aliceKeys.privateKey})
```

PR #1333, opened 2026-02-18
feat: two step transfer

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1318

Initial two-step transfer with online signing. Will need to test the reject/withdraw workflows and add the functionality for fetching the registry context liker we do in v0 version for offline signing.

### DIFF
--- a/core/ledger-client/src/ledger-client.ts
+++ b/core/ledger-client/src/ledger-client.ts
@@ -686,6 +686,15 @@ export class LedgerClient {
                                     }
                                 ).CreatedEvent
                         )
+                        // TODO: remove the filter once /v2/updates is fixed
+                        .filter((event) =>
+                            Object.keys(
+                                activeContractsArgs.filter?.filtersByParty ?? {}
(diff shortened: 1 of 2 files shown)


sdk/wallet-sdk/src/v1/token/client.ts
```diff
@@ -22,9 +28,98 @@ export type ListHoldingsParams = {
...
+            )
+
+            if (!asset || asset === undefined) {
+                throw new Error(
+                    `Asset with id ${params.instrumentId} not found in asset list for registry URL: ${params.registryUrl.href}`
+                )
+            }
+
+            const [transferCommand, disclosedContracts] =
+                await this.sdkContext.tokenStandardService.transfer.createTransfer(
```
alexmatson-da: the long args list here makes me wonder if this function should just take a whole `TransferParams` object and an asset?

alexmatson-da (approved):
thx!

PR #1351, opened 2026-02-24
feat(wallet-sdk): add option to execute party with ext sign v1

mateuszpiatkowski-da:

### DIFF
--- a/sdk/wallet-sdk/src/v1/party/external/prepared.ts
+++ b/sdk/wallet-sdk/src/v1/party/external/prepared.ts
@@ -36,7 +36,7 @@ export class PreparedPartyCreation {
                 ),
             })
         )
-        this.ctx.logger.info('Signed party successfully.')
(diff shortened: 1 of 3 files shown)


sdk/wallet-sdk/src/v1/party/external/prepared.ts
```diff
@@ -0,0 +1,74 @@
...
+        const signedParty = new SignedPartyCreation(
+            this.ctx,
+            signedPartyPromise,
+            this.createPartyOptions
+        )
+
+        return await signedParty.execute(options)
+    }
+
+    public async getParty() {
```
rukmini-basu-da: getParty() seems a bit misleading here because the party doesn't exist yet. What we're really getting are unsigned topology transactions, hash (that needs to be signed), and a fingerprint. It might be nice to rename this to response() or topology() instead.

PR #1358, opened 2026-02-25
feat: use ledger provider in wallet-sdk

rukmini-basu-da:
Closes https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1359

This adds the LedgerProvider to the wallet-sdk v1. This replaces all of the direct core-ledger-client calls via the sdk. 

We have a separate [issue](https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1360) for replacing the LedgerClient with the LedgerProvider in the core-token-standard-service and core-tx-parser. Once this is complete, we can remove the LedgerClient from the wallet-sdk context.

### DIFF
--- a/sdk/wallet-sdk/src/v1/party/external/types.ts
+++ b/sdk/wallet-sdk/src/v1/party/external/types.ts
@@ -22,39 +22,3 @@ export type ExecuteOptions = {
     party: GenerateTransactionResponse
     signature: string
 }
-
-/*
-
-    private async generateTopology(
-        params: GenerateTopologyParams
-    ): Promise<
-        Ops.PostV2PartiesExternalGenerateTopology['ledgerApi']['result']
-    > {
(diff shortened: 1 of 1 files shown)


sdk/wallet-sdk/src/v1/party/client.ts
```diff
@@ -21,18 +21,28 @@ export default class Party {
...
      */
     public async list(): Promise<PartyId[]> {
-        const rights = await this.ctx.ledgerClient.getWithRetry(
-            '/v2/users/{user-id}/rights',
-            defaultRetryableOptions,
-            {
-                path: { 'user-id': this.ctx.userId },
-            }
-        )
+        //TODO: what's the best way to handle retries
```
PHOL-DA: it is important we don't remove retry logic where we have it in place, either we need to wrap it around the ledgerProvider or have it inside the ledger provider itself.

@alexmatson-da what is your recommendation here ?

PR #1363, opened 2026-02-26
fix(wallet-gateway-remote): added cost estimation logging

PHOL-DA:


examples/ping/src/components/LedgerSubmission.tsx
```diff
@@ -36,6 +40,65 @@ export function LedgerSubmission(props: {
...
+        setLoading(true)
+
+        const responseByUpdateId = await getByUpdateId(updateId)
+        const contractId =
+            responseByUpdateId.transaction.events[0].CreatedEvent.contractId
+        sdk.prepareExecute(
+            exercisePongCommand(props.ledgerApiVersion, contractId)
+        )
+            .then(() => {
+                setLoading(false)
```
pawelstepien-da: You can use .finally to set loading false in case of both success and failure.

pawelstepien-da (approved):

PR #1381, opened 2026-02-27
feat: wallet reallocation in sync

pawelstepien-da:
This PR handles cases when there is a wallet in store, but there is no party anymore.
During wallet sync, if such case is detected, the wallet will be set to `status = 'initialized'`, therefore allowing user to re-allocate such party with "Allocate" button.

Edge case: if participant namespace changed (like when resetting localnet), then internal party can't be exactly re-allocated with same party id. I solved it by disabling the old wallet after re-allocation with reason "participant namespace changed" and result of re-allocation is creation of new wallet with same hint, but with current participant namespace. We could also disabled such wallet from the start without showing Allocate button. **Your input on it is welcome.**

I added toast messages when creating wallet, or clicking "allocate" on existing initialized wallet that say what whether:
-Regular case when wallet was created successfully
-A case when it was created but you need to sign with external signing provider to finalize
-A case when you you click allocate, but tx is not signed yet
-A case when you click allocate and it is signed and party gets allocated
-A case when you click allocate, but tx is rejected / failed and wallet gets removed

Fixes:
I fixed store.updateWallet that was overwriting some properties to empty value, namely `external_tx_id` and `topology_transactions`. This was happening when external signing provider wallet (fireblocks or blockdaemon) got allocated. Lack of stored `external_tx_id` made it impossible to check if topology transaction was signed externally if we wanted to re-allocate the external party on participant.

I fixed cases when disabled wallet was set as primary.

When clicking "Allocate" and transaction in fb/bd was failed or rejected, wallet was removed (which is good) but there was no information shown to the user and response contained wallet object without any information about what happened which was confusing. Now response contains info that wallet was removed and because of which tx status and a toast shows info about wallet being removed.

Notable refactor:
I extracted per-signing-provider logic of creating wallets from big switch in user api to separate service `wallet-creation-service`. I split the 2 separate flows that each signing provider had when initializing and allocating wallet to 3 methods each for better understanding, testing and potential usability without going through user api.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -17,5 +17,5 @@
         "source.fixAll.eslint": "always"
     },
     "metals.inlayHints.implicitArguments.enable": true,
-    "metals.inlayHints.implicitConversions.enable": true
+    "metals.inlayHints.implicitConversions.enable": true,
 }
--- a/core/acs-reader/package.json
+++ b/core/acs-reader/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@canton-network/core-acs-reader",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "type": "module",
     "description": "Reader for active contract set (ACS) data from the Canton ledger, providing functions to retrieve and process active contract information for specified filters.",
     "license": "Apache-2.0",
--- a/core/amulet-service/package.json
+++ b/core/amulet-service/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@canton-network/core-amulet-service",
-    "version": "0.3.0",
+    "version": "0.3.1",
     "type": "module",
     "description": "A service layer for interacting with non-token-standard Amulet operations on Canton.",
     "license": "Apache-2.0",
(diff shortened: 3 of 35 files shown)


core/wallet-store-inmemory/src/StoreInternal.ts
```diff
@@ -278,14 +278,28 @@ export class StoreInternal implements Store, AuthAware<StoreInternal> {
...
+        disabled,
+        reason,
+        primary,
     }: UpdateWallet): Promise<void> {
         const storage = this.getStorage()
-        // Use provided networkId or get current network from session
         const targetNetworkId = networkId ?? (await this.getCurrentNetwork()).id
 
+        const updates: Partial<Wallet> = {}
+        if (status !== undefined) updates.status = status
```
PHOL-DA: can't we use zod instead of having all these !== comparisons ?

wallet-gateway/remote/src/ledger/wallet-sync-service.ts
```diff
@@ -242,125 +229,192 @@ export class WalletSyncService {
...
+                }
+
+                this.logger.info({ ...wallet }, 'Wallet sync result')
+
+                return wallet
+            })
+        )
+
+        // Remove disabled wallets that are being re-synced before adding them back.
+        await Promise.all(
```
mjuchli-da: Get's very complicated here. Either remove if not needed or extract in separate method and document more extensively.

pawelstepien-da: We didn't need it any more, I removed it.

PR #1398, opened 2026-03-03
fix(wallet-gateway-remote): consistent relative pathing

PHOL-DA:
fixes the absolute pathing set multiple places in the files.

### DIFF
--- a/wallet-gateway/remote/src/web/frontend/constants.ts
+++ b/wallet-gateway/remote/src/web/frontend/constants.ts
@@ -31,7 +31,17 @@ export function normalizePathname(pathname: string): string {
         return '/'
     }
 
-    return pathname.replace(/\/+$/, '') || '/'
+    const trimmedPathname = pathname.replace(/\/+$/, '') || '/'
+
+    if (trimmedPathname === '/index.html') {
+        return '/'
+    }
+
+    if (trimmedPathname.endsWith('/index.html')) {
+        return trimmedPathname.slice(0, -'/index.html'.length) || '/'
+    }
+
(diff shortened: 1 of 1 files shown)


wallet-gateway/remote/src/web/frontend/constants.ts
```diff
@@ -22,4 +22,84 @@ export function isAllowedRoute(path: string): path is AllowedRoute {
     return ALLOWED_ROUTES.includes(path as (typeof ALLOWED_ROUTES)[number])
 }
 
+const NON_ROOT_ROUTES = ALLOWED_ROUTES.filter(
+    (route): route is Exclude<AllowedRoute, '/'> => route !== '/'
+).sort((left, right) => right.length - left.length)
+
+export function normalizePathname(pathname: string): string {
```
mjuchli-da: This method needs some documentation. There is too much regex and slicing going on.
I also think we don't need to export this.

wallet-gateway/remote/src/web/frontend/constants.ts
```diff
@@ -22,4 +22,84 @@ export function isAllowedRoute(path: string): path is AllowedRoute {
...
+    for (const route of NON_ROOT_ROUTES) {
+        if (normalizedPath === route || normalizedPath.endsWith(`${route}`)) {
+            return route
+        }
+    }
+
+    return null
+}
+
+export function getGatewayBasePath(pathname: string): string {
```
mjuchli-da: Export not needed.

pawelstepien-da (approved):

PR #1422, opened 2026-03-05
feat(wallet-sdk): amulet preapproval - fetch & create command v1

mateuszpiatkowski-da:
Note: this PR completes only a portion of preapproval items (see the issue' sub-issues).

### DIFF
--- a/core/acs-reader/package.json
+++ b/core/acs-reader/package.json
@@ -1,6 +1,6 @@
 {
     "name": "@canton-network/core-acs-reader",
-    "version": "1.1.0",
+    "version": "1.1.1",
     "type": "module",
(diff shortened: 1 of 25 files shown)


sdk/wallet-sdk/src/v1/party/client.ts
```diff
@@ -62,4 +62,10 @@ export default class Party {
 
         return Array.from(new Set(parties))
     }
+
+    public async getValidator() {
```
rukmini-basu-da: maybe we should make this part of the sdk context? I don't think we use any of the other validator endpoints, so we can fetch this once during the start up and then re-use it

PR #1425, opened 2026-03-05
feat(wallet-sdk): remove ledger client from sdk v1

rukmini-basu-da:
There were still a few leftover references to the LedgerClient in the wallet sdk during the external party creation. This replaces the LedgerClient with the LedgerProvider


rukmini-basu-da:
> I wonder if `core-ledger-client` can be uninstalled as a dependency of wallet-sdk now?

Yeah, I'll think we'll uninstall it as a dependency once we remove all the v0 components, which we'll do after we migrate more of the functionality because it helps being able to run the old tests

PR #1427, opened 2026-03-05
feat: improve discovery UI

fayi-da:
https://github.com/user-attachments/assets/13f5067a-9d3e-4a1e-b73f-b1cae8d7564d

image

### DIFF
--- a/core/wallet-ui-components/src/components/wallet-picker.ts
+++ b/core/wallet-ui-components/src/components/wallet-picker.ts
@@ -174,8 +174,8 @@ const SUBSTITUTABLE_CSS = cssToString([
         }
 
         .btn-add {
-            background: var(--wg-theme-primary-color, #80deea);
-            color: var(--wg-theme-text-color, #1a1a1a);
+            background: var(--wg-theme-primary-color, #000000);
+            color: var(--wg-theme-primary-text-color, #ffffff);
             border: none;
             border-radius: 20px;
             padding: 10px 24px;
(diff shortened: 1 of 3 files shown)


sdk-support/provider-adapter-loop/src/assets.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Base64-encoded icon assets for the 5N Loop Wallet adapter.
+ *
+ * These icons are embedded as data URLs so they work immediately after npm install
+ * without requiring consumers to host image files or configure asset paths.
+ */
+
+/**
+ * 5N Loop Wallet logo (PNG)
+ */
+export const FIVEN_LOOP_LOGO_PNG =
```
fayi-da: ok so this is probably going to be the most contentious thing in this PR. We can't reference actual images because whoever uses the sdk will have to deal with image bundling issues or configuring asset paths i node_modules. It also felt weird that an image is bundled into an SDK. 

Anyway, after some research, I chose to `base64` the 2 logos we use, canton logo and the 5n logo. This means there no bundling issue and images always work regardless of where they are used from. Thoughts?

fayi-da (commented):
.

sdk-support/provider-adapter-loop/src/assets.ts
mateuszpiatkowski-da: IMO this should work, although this way we won't be able to cache the images if we'd like to. I think the best way would be to upload these imgs somewhere on the cdn and cache the link from which we upload them ;) But base64 is also good as the size of them shouldn't be an issue to us.


Also, I'd consider uploading standard .png images anyway in case a client wishes to process them with their own bundler :)

Just dropping free thoughts :D

PR #1486, opened 2026-03-18
chore: improve playwright stability

alexmatson-da:
the playwright tests for the example ping app feel very flaky when I run them on my machine... this PR tries to make things more reliable


PHOL-DA:
i have fixed this over in https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1586, the primary reason it flakes locally is because we have multi-runner setup (so multiple calls using the same user)

PHOL-DA:
@alexmatson-da is this something we can close or anything worth salvaging ?

alexmatson-da:
> @alexmatson-da is this something we can close or anything worth salvaging ?

will close for now

PR #1490, opened 2026-03-19
feat(wallet-sdk): merge utxos with delegation proxy v1

mateuszpiatkowski-da:


sdk/wallet-sdk/src/v1/namespace/ledger/internal.ts
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { PostEndpoint, PostRequest } from '@canton-network/core-ledger-client'
+import { WalletSdkContext } from '../../sdk.js'
+import { v4 } from 'uuid'
+import { Ops } from '@canton-network/core-provider-ledger'
+
+type InternalPartyOperationPaths = Extract<
```
rukmini-basu-da: nitpitck: any way we can use: Is there any way we can use Ops.PostV2CommandsSubmitAndWait['ledgerApi'] and extract the body via ['params']['body'] similar to here: https://github.com/hyperledger-labs/splice-wallet-kernel/blob/295da1e687488f781ef29429af95d2d0ec2f3aef/sdk/wallet-sdk/src/v1/namespace/ledger/client.ts#L39

PR #1492, opened 2026-03-19
feat: user party rights

mjuchli-da:


core/wallet-store-sql/src/schema.ts
```diff
@@ -45,19 +52,32 @@ interface WalletTable {
...
-    status?: string
+    externalTxId?: string | undefined
+    topologyTransactions?: string | undefined
+    status?: string | undefined
     disabled?: number
-    reason?: string
+    reason?: string | undefined
+}
+
+interface UserPartyRightTable {
```
mjuchli-da: @alexmatson-da @pawelstepien-da what is your take on this?
An alternative (which I had first) was to extend the `WalletTable` with fields:
- rights_can_act_as
- rights_can_read_as
- rights_can_execute_as
Both works.

PR #1590, opened 2026-04-09
fix(wallet-gateway-remote): rate limit based on key

PHOL-DA:


wallet-gateway/remote/src/middleware/rateLimit.ts
```diff
@@ -2,11 +2,59 @@
...
+    })
+}
+
+export function preAuthIpRateLimiter(requestRateLimit: number) {
+    return rateLimit({
+        windowMs: 1 * 60 * 1000,
+        max: requestRateLimit,
+        keyGenerator: ipRateLimitKeyGenerator,
+        // Only protect unauthenticated traffic before JWT verification.
+        skip: hasBearerToken,
```
mjuchli-da, line 45: Returns true for any `Authorization: Bearer` (or `?token=…`) even if the token is invalid. That means an attacker can send many requests with junk bearer tokens and skip the only limiter that runs before authService.verifyToken(...), forcing expensive JWT verification work with no rate limit. 

Could/should we just remove this skip step, so that all traffic is IP-rate-limited before JWT verification, (including invalid tokens)?

PHOL-DA: so the current flow is:
1. preAuthIpRateLimiter 
2. JwtVerification
3. postAuthIpRateLimiter

i don't think there is a way to prevent junk bearer tokens without validating them. The RateLimiter can not really be put down. I "assume" that the jwt verification can quickly throw away an invalid token (and is not subject to DDOS).

removing the skip means that we rate limit everything twice, which does not solve the problem (since behind a reverse proxy all users will share the same IP address and somebody malicious will take it down regardless of junk tokens)

mjuchli-da: but isn't it in general better to not skip in (1) and therefore prevent ddos on (2)?

mjuchli-da: or is this anyways covered in non-reverse-proxy setups?

PHOL-DA: well you have to ddos (2) within the rate limit (167 request per minute) which i don't believe is even feasible (or atleast not for a long time). The advantage here is that people with a valid token will not be hit by the 167 limit

PHOL-DA: also a reverse proxy should be configured to rate limit based on IP before ours get hit

mjuchli-da (approved):

PR #1592, opened 2026-04-09
fix(core-wallet-store): prevent overwrites of transactions

PHOL-DA:


mjuchli-da (approved):
Thanks.

I created a follow-up issue: https://github.com/hyperledger-labs/splice-wallet-kernel/issues/1596
Not security related but a potential source for error handling imrovements.

PR #1593, opened 2026-04-09
docs: wallet-sdk v1 party namespace migration

### LINKED ISSUE #1540: Party namespace migration

mateuszpiatkowski-da:


docs/wallet-integration-guide/_static/custom.css
```diff
@@ -0,0 +1,15 @@
+@media (width >= 900px) {
```
PHOL-DA, line 1: what is this custom.css file for (we are not displaying directly from this repo and rather it is copied over to a different repo where styling is applied)

docs/wallet-integration-guide/_extensions/before_after.py
```diff
@@ -0,0 +1,63 @@
+from docutils import nodes
+from docutils.parsers.rst import Directive
+from docutils.statemachine import StringList
+
+class BeforeAfterDirective(Directive):
```
PHOL-DA, line 5: this looks nice, i would however ask Davin if there i something similar already supported for the docs repo.

PHOL-DA (approved):

docs/wallet-integration-guide/_static/custom.css
mateuszpiatkowski-da: @PHOL-DA I was thinking of expanding the container where content is displayed for convenience and better readability. It was a bit frustrating for me to see the code blocks being overflown with horizontal sliders, hence the custom.css. I can remove it if you don't see the value though :)

rukmini-basu-da (approved):
nice! thanks

docs/wallet-integration-guide/_extensions/before_after.py
mateuszpiatkowski-da: I've relayed the message. I'll modify the code in case there is some tool already available :)

PR #1594, opened 2026-04-09
fix: restore party allocation toasts

pawelstepien-da:
when creating or allocating wallets show toast messages based on status


fayi-da (approved):
thanks

PR #1595, opened 2026-04-10
feat: wallet connect integration

ganchoradkov:
## Summary

Adds WalletConnect transport support to the dApp SDK, enabling dApps to connect to Canton wallets over the WalletConnect v2 relay. Includes a standalone wallet example app.

### Architecture

```mermaid
graph LR
    subgraph dApp
        SDK[dApp SDK + SignClient]
    end

    subgraph WCRelay
        WC[Encrypted relay]
    end

    subgraph WalletGateway
        WKPage[WalletConnect Page + WalletKit]
        Handler[Request Handler]
        API[dApp API / User API]
    end

    subgraph Canton
        Ledger[Ledger API]
    end

    SDK -- canton methods --> WC
    WC -- session_request --> WKPage
    WKPage --> Handler
    Handler -- JSON-RPC --> API
    API -- Bearer JWT --> Ledger
```

### Request flow

```mermaid
sequenceDiagram
    participant dApp
    participant Relay as WC Relay
    participant GW as Gateway WC Page
    participant API as Gateway API
    participant L as Ledger API

    dApp->>Relay: canton_prepareSignExecute
    Relay->>GW: session_request

    alt Read-only
        GW->>API: forward immediately
    else Mutating
        GW-->>GW: queue for user approval
        Note over GW: User clicks Approve
    end

    GW->>API: prepareExecute
    API->>L: prepare + sign + execute
    L-->>API: completion
    API-->>GW: result
    GW->>Relay: respond
    Relay->>dApp: executed result
```

### WC protocol spec

| SDK method | Wire method | Approval |
|---|---|---|
| `prepareExecute` | `canton_prepareSignExecute` | Manual |
| `prepareExecuteAndWait` | `canton_prepareSignExecute` | Manual |
| `ledgerApi` | `canton_ledgerApi` | Auto |
| `listAccounts` | `canton_listAccounts` | Auto |
| `getPrimaryAccount` | `canton_getPrimaryAccount` | Auto |
| `getActiveNetwork` | `canton_getActiveNetwork` | Auto |
| `status` | `canton_status` | Auto |
| `signMessage` | `canton_signMessage` | Manual |

- `connect` / `disconnect` are handled locally and never sent over the relay
- Both `prepareExecute` and `prepareExecuteAndWait` collapse into `canton_prepareSignExecute` — the wallet does the full prepare-sign-execute cycle
- Events: `accountsChanged`, `statusChanged` via session_event; `session_delete` is a built-in WC event

### sdk/dapp-sdk

- **WalletConnectAdapter** — single class implementing both `ProviderAdapter` and `Provider<DappRpcTypes>`. Calls `signClient.request()` directly with `canton_` prefix. No intermediate transport or RPC client layers. Includes a WalletConnect icon for the wallet picker.
- **DappSDK** — WalletConnect adapter is registered by dApps via `additionalAdapters` (not built into SDK core). Added `init()` method for passive session restoration.
- **Exported** `WalletConnectAdapter` and `WalletConnectAdapterConfig` from the package index.

### examples/wallet

Standalone React app on port 8082 that acts as a WalletConnect wallet:
- WalletKit initialization and session management
- Auto-dispatch for read-only methods (status, ledgerApi, listAccounts, etc.)
- Manual approval UI for `canton_prepareSignExecute` and `canton_signMessage`
- Session proposals with network selection
- Full prepare + sign + execute flow via the gateway dApp and User APIs
- **auth.ts** — replaced custom `SignJWT` implementation with `AuthTokenProvider` from `@canton-network/core-wallet-auth` shared library

### examples/ping

- Instantiates `WalletConnectAdapter` directly, passes via `additionalAdapters`
- Session restore via `sdk.init()` on mount

### examples/portfolio

- **Fixed disconnect handling**: split the single mount effect into two effects so `onStatusChanged` is registered after connection
- Instantiates `WalletConnectAdapter` directly, passes via `additionalAdapters`
- Session restore via `sdk.init()` on mount

### core/wallet-ui-components

- Wallet picker handles `wc-uri` postMessage to display the WC pairing URI for copy/paste
- Added WalletConnect navigation link in the app header

## Test plan

- [ ] Ping + WalletConnect: start Canton, Gateway, Wallet Example, Ping. Connect via WC, create Ping contract, verify transaction appears
- [ ] Portfolio + WalletConnect: connect via WC, verify holdings load, disconnect from wallet side, verify portfolio resets cleanly
- [ ] Session restore: connect via WC, refresh dApp page, verify session is restored without re-pairing
- [ ] Wallet-side disconnect: connect, disconnect from wallet/gateway WC page, verify dApp resets to disconnected state
- [ ] dApp-side disconnect: connect, disconnect from dApp settings, verify wallet shows session removed
- [ ] Existing flows: verify HTTP gateway connection still works for both ping and portfolio


examples/walletconnect/src/walletkit/gateway.ts
```diff
@@ -0,0 +1,163 @@
...
+        'listNetworks',
+        {}
+    )
+    return result.networks
+}
+
+export async function bootstrapSession(
+    networkId: string
+): Promise<{ id: string }> {
+    return callUserApi<{ id: string }>('addSession', { networkId })
```
mjuchli-da, line 75: This only works with localnet because unauthenticated calls are enabled. For any other setups where the user connects (via the Wallet Gateway) to an authenticated network (`yarn start:canton` is a local example), this will break: the `addSession` endpoint attempts to call the ledgerApi.

examples/walletconnect/src/App.tsx
```diff
@@ -0,0 +1,176 @@
+import { useState, useCallback } from 'react'
```
mjuchli-da, line 1: It would probably be easier and more efficient to just extend the `wallet-gateway-remote` directly.

This way, all we'd have to add is the ability to establish and list WC sessions. In fact, we would ideally provide this via a dedicated package under `core/wallet-connect`, which the Remote Wallet Gateway depends on, and therefore serves as a reference for other wallet providers to integrate the core package the same way (or replicate if they wish to do so).

Currently, this dApp depends on the WG as a backend and the dApp and replicates the same that the Wallet Gateway frontend does – not per se a bad thing!, but in this case I don't see the benefit over the above suggestion.

core/wallet-ui-components/src/components/wallet-picker.ts
```diff
@@ -710,21 +711,49 @@ export class WalletPicker extends HTMLElement {
         )
 
         const view = this.el('div', '', { class: 'status-view' })
-        view.appendChild(this.el('div', '', { class: 'spinner' }))
-        view.appendChild(
-            this.el(
-                'h3',
-                'Connecting to ' + (this.selectedEntry?.name || '') + '...'
+
+        if (this.wcUri) {
```
mjuchli-da, line 807: @fayi-da please check the WC integration in the picker.

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -52,6 +53,8 @@ export interface DappSDKConnectOptions<
 > {
     defaultAdapters?: TDefaultAdapter[]
     additionalAdapters?: ProviderAdapter[] | undefined
+    /** WalletConnect Cloud project ID. When provided, a WalletConnect adapter is automatically registered. */
+    walletConnectProjectId?: string
```
mjuchli-da: If I understand correctly, every dApp will need to have their projectId entered manually. In this case, I suggest we move the projectId to the configuration of the `WalletConnectAdapter` and provide a developer friendly (static) method to create the adapter instance. Then register the adapter under `additionAdapters`.

mjuchli-da (commented):
Thank you very much for the contribution!

I just made a first pass and reported some of my high-level findings to discuss here. 
I will make a second and more detailed pass no later than next Monday.

examples/walletconnect/src/App.tsx
ganchoradkov: implemented new page in the wallet-gateway `/walletconnect`

mjuchli-da: we agreed that we skip the integration into the WG for now. Instead the team plans to build a dedicated core module which can be used by wallet providers (including the ones building on top of the wallet gateway).

core/wallet-ui-components/src/components/wallet-picker.ts
```diff
@@ -799,21 +803,66 @@ export class WalletPicker extends HTMLElement {
...
-        )
-        view.appendChild(
-            this.el(
-                'p',
-                this.selectedEntry?.type === 'remote'
-                    ? 'Approve the connection in the wallet popup'
-                    : 'Approve the connection in your extension'
+
+            const code = this.el('code', this.wcUri)
+            code.style.cssText =
```
mjuchli-da, line 828: Avoid inline styles. Whole block below needs refactoring.

mjuchli-da (commented):
Thanks! I think we are almost there. I've added some changes to the dApp SDK directly and the only thing left I see as mandatory is refactoring the wallet picker. I saw a few bits on the Wallet UI as well but given that this is a prototype, it is IMHO fine for this PR.

mjuchli-da (approved):

core/wallet-ui-components/src/components/wallet-picker.ts
mjuchli-da: To be fixed in a follow-up @fayi-da

core/wallet-ui-components/src/components/wallet-picker.ts
mjuchli-da: to be fixed in a follow-up.

PHOL-DA (approved):

PR #1602, opened 2026-04-10
feat: migrate blockdaemon signing sdk from test network to caip2

LimKianAn:
- Migrate signing SDK from `testNetwork` to `caip2`
- `setConfiguration` still accepts `TestNetwork` for backward compatibility
- Remove `BlockDaemonSignTransactionParams` workaround

### DIFF
--- a/core/signing-blockdaemon/src/signing-api-sdk.test.ts
+++ b/core/signing-blockdaemon/src/signing-api-sdk.test.ts
@@ -1,0 +1,68 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { expect, describe, test, beforeEach } from 'vitest'
+import BlockdaemonSigningDriver from './index.js'
+import { SigningAPIClient } from './signing-api-sdk.js'
+
+describe('BlockdaemonSigningDriver constructor', () => {
+    test('passes caip2 from config to the client', () => {
+        const driver = new BlockdaemonSigningDriver({
+            baseUrl: 'http://localhost:9999',
+            apiKey: 'key',
+            caip2: 'canton:testnet',
+        })
+        const client = (driver as unknown as { client: SigningAPIClient })
+            .client
+        expect(client.getConfiguration().CAIP2).toBe('canton:testnet')
+    })
+
+    test('defaults to canton:devnet when caip2 not provided', () => {
+        const driver = new BlockdaemonSigningDriver({
+            baseUrl: 'http://localhost:9999',
+            apiKey: 'key',
+        })
+        const client = (driver as unknown as { client: SigningAPIClient })
+            .client
+        expect(client.getConfiguration().CAIP2).toBe('canton:devnet')
+    })
+})
+
+describe('SigningAPIClient.setConfiguration CAIP2', () => {
+    let client: SigningAPIClient
+
+    beforeEach(() => {
+        client = new SigningAPIClient('http://localhost:9999')
+    })
+
+    test('defaults to canton:devnet', () => {
+        expect(client.getConfiguration().CAIP2).toBe('canton:devnet')
+    })
+
+    test('sets Caip2 directly', () => {
+        client.setConfiguration({ Caip2: 'canton:testnet' })
+        expect(client.getConfiguration().CAIP2).toBe('canton:testnet')
+    })
+
+    test('maps deprecated TestNetwork to CAIP2', () => {
+        client.setConfiguration({ TestNetwork: false })
+        expect(client.getConfiguration().CAIP2).toBe('canton:mainnet')
+
+        client.setConfiguration({ TestNetwork: true })
+        expect(client.getConfiguration().CAIP2).toBe('canton:devnet')
+    })
+
+    test('accepts consistent Caip2 + TestNetwork', () => {
+        client.setConfiguration({ Caip2: 'canton:testnet', TestNetwork: true })
+        expect(client.getConfiguration().CAIP2).toBe('canton:testnet')
+    })
+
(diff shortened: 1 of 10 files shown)


core/signing-blockdaemon/src/signing-api-sdk.ts
```diff
@@ -132,26 +128,37 @@ export class SigningAPIClient {
...
+     *
      * @param params - Configuration parameters to set. All fields are optional.
      */
     public setConfiguration(params: {
         BaseURL?: string
         ApiKey?: string
         MasterKey?: string
+        /** CAIP2 chain identifier (e.g. `canton:devnet`, `canton:mainnet`). */
+        Caip2?: string
+        /** @deprecated Use `Caip2` instead. */
```
PHOL-DA: since this is just internal code then there is no need to add the deprecation, which would also make the code a lot simpler

core/signing-blockdaemon/src/signing-api-sdk.test.ts
```diff
@@ -0,0 +1,68 @@
...
+            baseUrl: 'http://localhost:9999',
+            apiKey: 'key',
+        })
+        const client = (driver as unknown as { client: SigningAPIClient })
+            .client
+        expect(client.getConfiguration().CAIP2).toBe('canton:devnet')
+    })
+})
+
+describe('SigningAPIClient.setConfiguration CAIP2', () => {
```
PHOL-DA: this test dosen't really make sense, it test the same logic as above however there is no reason to do it on the SigningAPICllient.

core/signing-blockdaemon/src/signing-api-sdk.ts
```diff
@@ -21,24 +17,24 @@ export class SigningAPIClient {
...
         params: I
     ): Promise<O> {
         const url = `${this.baseUrl}${endpoint}`
-        // Merge context params (masterKey and testNetwork) into the request body
+        // Merge context params (masterKey and caip2) into the request body
         const bodyToSend = {
             ...params,
             masterKey: this.masterKey,
-            testNetwork: this.testNetwork,
+            caip2: this.caip2,
```
PHOL-DA, line 38: @LimKianAn have you ensured on your side that the api can both handle caip2 and testNetwork or is this another "breaking change" ?

wallet-gateway/remote/src/env.ts
```diff
@@ -8,6 +8,7 @@ export class Env {
         Env.get('BLOCKDAEMON_API_URL', { fallback })
     static BLOCKDAEMON_API_KEY = (fallback: string) =>
         Env.get('BLOCKDAEMON_API_KEY', { fallback })
+    static BLOCKDAEMON_CAIP2 = () => Env.get('BLOCKDAEMON_CAIP2')
```
PHOL-DA: we probably don't need to use an ENV variable here (and can just a regular config)

core/signing-blockdaemon/src/signing-api-sdk.ts
PHOL-DA: i can see from our CI tests that you have not make this backwards compatible

core/signing-blockdaemon/src/signing-api-sdk.ts
```diff
@@ -21,24 +17,24 @@ export class SigningAPIClient {
...
     private apiKey: string | undefined
     private masterKey: string
-    private testNetwork: boolean
+    private caip2: string
 
     constructor(baseUrl: string) {
         this.baseUrl = baseUrl.endsWith('/') ? baseUrl.slice(0, -1) : baseUrl
         this.masterKey = 'Default'
-        this.testNetwork = true
+        this.caip2 = 'canton:devnet'
```
PHOL-DA, line 27: does it accept any caip2 format or only a selected few (like canton:devnet, canton:testnet, canton:mainnet) ?

core/signing-blockdaemon/src/signing-api-sdk.ts
mjuchli-da, line 38: We should allow passing this in via the params:
`caip2: params.caip2 ?? this.caip2`

Maybe ensure via types even?

mjuchli-da (changes requested):
Thanks for the contribution!

One request from my side: the CAIP-2 setting is currently "driver-global", not runtime/per-network. Users can switch networks at runtime, and those networks have different CAIP-2 namespaces, then the current design forces either:
- one driver instance per network, or
- mutating `this.caip2` at runtime

Instead, I suggest we allow a `caip2` override in each SDK call.

core/signing-blockdaemon/src/signing-api-sdk.ts
LimKianAn: Adopted

core/signing-blockdaemon/src/signing-api-sdk.test.ts
LimKianAn: Adopted

wallet-gateway/remote/src/env.ts
LimKianAn: Adopted

core/signing-blockdaemon/src/signing-api-sdk.ts
LimKianAn: Adopted

core/signing-blockdaemon/src/signing-api-sdk.ts
LimKianAn: Adopted

alexmatson-da: is there any reason to make this part of the static config then? (I'm assuming) the WG will always pass in the CAIP corresponding to the active network

PHOL-DA:
closed in favor of #1611

PR #1607, opened 2026-04-13
feat: added custom dars example

PHOL-DA:
anything in the **damljs/** is codegen, so i leave it up to the reviewer if they want to go through those 600+ files.

recommended to reivew:
[script 14, showcasing custom dar implementation](https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1607/changes/8e8ab0744c4dc8dc1480fee3314d4931f71070e6..91420eec195e2559f9ec98dc89dff1e79b1958f5#diff-def5dbfffc7762cbc3bdc29d4f1fac4118e56bbfed085227abe8903d31a18f14)
[Typed contracts](https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1607/changes#diff-ef4db739064f3b685e8428851f3eb11bdb44ed9ddddefeee34dfeb1febb49eb3)
[implementation guide](https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1607/changes/8e8ab0744c4dc8dc1480fee3314d4931f71070e6..91420eec195e2559f9ec98dc89dff1e79b1958f5#diff-55adcc84cc775c3366911447728672c93ef23fff9400cba9eac580dbc58562a7)


core/acs-reader/src/index.ts
```diff
@@ -2,3 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0
 
 export { AcsReader, AcsOptions } from './acs-reader.js'
+export {
```
mjuchli-da, line 5: relevant

core/acs-reader/src/typed-contract.ts
```diff
@@ -0,0 +1,62 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export type AcsContractLike = {
```
mjuchli-da, line 4: relevant

package.json
```diff
@@ -30,6 +30,7 @@
         "script:generate:protobufs": "tsx ./scripts/src/generate-protobufs.ts",
         "script:generate:tokenstandard": "tsx ./scripts/src/generate-token-standard.ts",
         "script:generate:featured-dars": "tsx ./scripts/src/generate-featured-dars.ts",
+        "script:generate:dvp-testing-app": "tsx ./scripts/src/generate-dvp-testing-app.ts",
```
mjuchli-da, line 33: Command specific so an example app

package.json
```diff
@@ -53,7 +54,11 @@
         "api-specs",
         "core/**",
         "docs/wallet-integration-guide/examples",
-        "damljs/**",
+        "damljs/token-standard-models/**",
```
mjuchli-da, line 57: Also an app under /examples can depend on some 3rd party package – either internally generated or from an external source

package.json
```diff
@@ -122,7 +127,31 @@
         "glob": "^10.5.0",
         "vite": "^7.2.4",
         "uri-js": "npm:uri-js-replace@^1.0.1",
-        "seroval": "^1.4.2"
+        "seroval": "^1.4.2",
+        "@daml.js/daml-prim-DA-Exception-ArithmeticError-1.0.0": "workspace:^",
```
mjuchli-da, line 131: This is a classical setup (standard practice since Daml 2.0), where a frontend app (dvp example in this case) depends on 1) generic daml types and 2) dar specific codegen. As such, we shouldn't have any of these dependencies in our root package.json

scripts/src/generate-dvp-testing-app.ts
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
mjuchli-da, line 1: Put in /examples/dvp folder directly

package.json
mjuchli-da, line 33: Move to example app

package.json
mjuchli-da, line 57: Codegen should not be part of the repo

docs/wallet-integration-guide/src/custom-dars/index.rst
```diff
@@ -0,0 +1,59 @@
...
+                    transferLegs,
+                    approvers: [alice.partyId, bob.partyId],
+                },
+            },
+        },
+4. use ``toTemplateContracts`` and ``toTemplateContract`` to easily convert ACS responses to type-strong objects like:
+
+    .. code-block:: typescript
+    
+        const proposalsForBob = toTemplateContracts<
```
mjuchli-da, line 49: I would try to hide this in `sdk.ledger.acs.read(args, TradingApp.COMPANION)`

core/acs-reader/src/typed-contract.ts
```diff
@@ -0,0 +1,62 @@
...
+    expectedTemplateId: string
+): TypedAcsContract<TPayload, TContract> | undefined {
+    if (contract.templateId !== expectedTemplateId) {
+        return undefined
+    }
+
+    return contract as TypedAcsContract<TPayload, TContract>
+}
+
+export function toTemplateContracts<
```
mjuchli-da, line 27: In the java codegen there are COMPANION objects for each template, isn't this the case in typescript?
If so, they come with a parse function.

mjuchli-da, line 27: In the java codegen there are COMPANION objects for each template, isn't this the case in typescript?
If so, they come with a parse function.

mjuchli-da (changes requested):
The focus of this PR lays imho on `typed-contract.ts`, which is a (de-)serialization facility from/to codegen objects.

This has been clunky for a while in the current codegen and there is an ongoing (scala) project that is supposed to introduce an improved codegen that covers the features here (will remember the name of the project tomorrow hopefully). 

That said, we have done similar things in Java in the past where one could use the COMPANION object (belonging to a codegenned template): https://github.com/mjuchli-da/pqsj/blob/main/src/main/java/com/daml/pqsj/repository/ContractRepository.java#L44. Maybe there is similar stuff for Typescript given?

As for the rest: this resembles a classical App that makes use of some arbitrary daml codegen (and daml generic types) – the same way it has been done since Daml 2.0. As such, any reference to dvp should imho reside as an independent example app under /examples/dvp (or similar). This includes all the boilerplating necessary to build this app; except of course the codegen itself which must be built as part of the example app build.

(Extremely hard to review, highlights that codegen does not belong in a repo).

core/acs-reader/src/typed-contract.ts
PHOL-DA: no companion object here (this is the TS file for dvp trading app):
https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1607/changes#diff-9bff8d85136cfa7acd2689527cd10e0daef5ec18d9431c1aaf973ae8f190ff71

package.json
PHOL-DA: so the dvp trading is not an example app, we use it several places in code (script 04 and otc-trade.ts specifically in our repo).

so any place where we test transfer allocations, then we need the dvp trading app

package.json
PHOL-DA: codegen is already part of the repo, we do it for the token standard. All i did was make it work with more than one codegen from daml in the repo

PHOL-DA: this was actually to avoid to many of the codegen packages being part of our repo (now we only have the four main packages instead of every single codegen package)

package.json
PHOL-DA: this is not dependencies but resolutions, im just telling the repo (now that it is finding multiple versions) which version to use

PR #1612, opened 2026-04-13
feat: rename namespaces and remove core-ledger-client dep

rukmini-basu-da:


sdk/wallet-sdk/src/wallet/namespace/amulet/client.ts
```diff
@@ -26,14 +26,14 @@ export type AmuletNamespaceConfig = {
     validatorParty: PartyId
 }
 
-export class Amulet {
+export class AmuletNamespace {
     public readonly traffic: Traffic
     public readonly preapproval: Preapproval
```
alexmatson-da: Should we also consider these as (nested) namespaces?

alexmatson-da (approved):
nice!

alexmatson-da (approved):

PR #1616, opened 2026-04-14
docs: add offline signing script

rukmini-basu-da:
Adds a script to show the full offline signing flow, comparable to the[ v0 version ](https://docs.digitalasset.com/integrate/devnet/preparing-and-signing-transactions/index.html#how-to-use-the-sdk-to-offline-sign-a-transaction)

### DIFF
--- a/docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
+++ b/docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
@@ -1,0 +1,279 @@
+import {
+    localNetStaticConfig,
+    SDK,
+    signTransactionHash,
+} from '@canton-network/wallet-sdk'
+import { pino } from 'pino'
+import { v4 } from 'uuid'
+import {
+    TOKEN_NAMESPACE_CONFIG,
+    TOKEN_PROVIDER_CONFIG_DEFAULT,
+    AMULET_NAMESPACE_CONFIG,
+} from './utils/index.js'
+
+const onlineLogger = pino({ name: '14-online-localnet', level: 'info' })
+const offlineLogger = pino({ name: '14-oggline-localnet', level: 'info' })
+
+const onlineSDK = await SDK.create({
+    auth: TOKEN_PROVIDER_CONFIG_DEFAULT,
+    ledgerClientUrl: localNetStaticConfig.LOCALNET_APP_USER_LEDGER_URL,
+})
+
+onlineLogger.info(`Online sdk initialized.`)
+
+const offlineSdk = await SDK.create({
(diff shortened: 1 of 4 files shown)


docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
```diff
@@ -0,0 +1,278 @@
...
+
+const onlineLogger = pino({ name: '14-online-localnet', level: 'info' })
+const offlineLogger = pino({ name: '14-oggline-localnet', level: 'info' })
+
+const onlineSDK = await SDK.create({
+    auth: TOKEN_PROVIDER_CONFIG_DEFAULT,
+    ledgerClientUrl: localNetStaticConfig.LOCALNET_APP_USER_LEDGER_URL,
+})
+
+onlineLogger.info(`Online sdk initialized.`)
```
PHOL-DA, line 21: i would like to make the lines between online and offline more visible, especially when you are scrolling through the script file.

docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
```diff
@@ -0,0 +1,278 @@
...
+    }
+)
+
+const senderPartyTopology = await senderPartyPrepared.topology()
+
+onlineLogger.info(
+    `Prepared sender onboarding with multi hash: ${senderPartyTopology.multiHash}`
+)
+
+const senderSignedTopologyTx = signTransactionHash(
```
PHOL-DA, line 66: "offline" does not trust "online", we therefore need to recompute the multihash and check that it matches

docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
```diff
@@ -0,0 +1,278 @@
...
+    '10000'
+)
+
+const preparedTapCommand = onlineSDK.ledger.prepare({
+    partyId: senderParty.partyId,
+    commands: amuletTapCommand,
+    disclosedContracts: amuletTapDisclosedContracts,
+})
+
+const { response: preparedTapCommandResponse } =
```
PHOL-DA: can't we just await on line 101 instead ?

PHOL-DA (commented):
looks good, we are missing the multi-hash recomputation and then make it clear in the file when data is switching from online to offline mode

docs/wallet-integration-guide/examples/scripts/14-offline-signing.ts
rukmini-basu-da: sounds good, added a function for the topology tx multi hash computation (previously we just had one for the prepared tx for ledger command submission)

PHOL-DA (approved):

PR #1619, opened 2026-04-14
chore: change secrests to vars

PHOL-DA:


pawelstepien-da (commented):
What is the reason for using vars instead of secrets for actual API keys?

PHOL-DA:
> What is the reason for using vars instead of secrets for actual API keys?

so it can be used from forks as well

pawelstepien-da (approved):

PR #1628, opened 2026-04-15
feat(core-wallet-store): changed transaction to have unique id

### LINKED ISSUE #1626: Custom identifier for transaction
We currently rely on the `commandId` as the primary key for the `Transaction` entity.
This may cause problems as in Canton, a command is considered a duplicate if all three of these match:
- `commandId`
- `actAs` (submitting party or parties)
- within the deduplication period
Therefore the uniqueness cannot be guaranteed, especially not for a `Transaction` entity which represents the lifecycle of a command from creation to execution against the ledger -> e.g. may be repeated by the user.

Instead, we should make use of a `UUID` type to represent the `id` field (PK) of the `Transaction` table.
As for the migration, we can alter(postgres)/recreate(sqlite) the `Transaction` table in such a way that we create the new `id` column and generate a `UUID` during the migration. The existing `commandId` remains.

PHOL-DA:
Fixes #1626 

- [X] added transaction id
- [X] wrote migration script
- [X] exposed the transaction id all the way to the ui
- [X] enable committing multiple transaction with same command id (but different transaction id)

### DIFF
--- a/core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
+++ b/core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
@@ -162,6 +162,7 @@ export async function down(db: Kysely<DB>): Promise<void> {
             external_tx_id
         )
         SELECT
+            command_id || ':' || id,
             status,
             prepared_transaction,
             prepared_transaction_hash,
--- a/core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
+++ b/core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
@@ -1,0 +1,180 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { randomUUID } from 'node:crypto'
+import { Kysely, sql } from 'kysely'
+import { DB } from '../schema.js'
+import { isPostgres } from '../utils.js'
+
+export async function up(db: Kysely<DB>): Promise<void> {
+    console.log('Migrating transactions primary key from command_id to id')
+
+    const pg = await isPostgres(db)
+    if (pg) {
+        await sql`CREATE EXTENSION IF NOT EXISTS pgcrypto`.execute(db)
+
+        await sql`ALTER TABLE transactions ADD COLUMN id UUID`.execute(db)
+        await sql`UPDATE transactions SET id = gen_random_uuid() WHERE id IS NULL`.execute(
+            db
+        )
+        await sql`ALTER TABLE transactions ALTER COLUMN id SET NOT NULL`.execute(
+            db
+        )
+        await sql`ALTER TABLE transactions ALTER COLUMN id SET DEFAULT gen_random_uuid()`.execute(
+            db
+        )
+        await sql`ALTER TABLE transactions DROP CONSTRAINT IF EXISTS transactions_pkey`.execute(
+            db
+        )
+        await sql`ALTER TABLE transactions ADD CONSTRAINT transactions_pkey PRIMARY KEY (id)`.execute(
+            db
+        )
+        await sql`
+            CREATE INDEX IF NOT EXISTS transactions_command_user_network_idx
+            ON transactions (command_id, user_id, network_id)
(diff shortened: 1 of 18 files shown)


wallet-gateway/remote/src/web/frontend/activities/index.ts
```diff
@@ -132,7 +130,7 @@ export class UserUiActivities extends BaseElement {
 
     private _onReview(e: TransactionCardReviewEvent) {
         const approveHref = toRelHref('/approve')
-        window.location.href = `${approveHref}?commandId=${e.commandId}`
+        window.location.href = `${approveHref}?transactionId=${e.transactionId}`
```
pawelstepien-da, line 133: Without `commandId` query param it doesn't load transaction data

wallet-gateway/remote/src/web/frontend/approve/index.ts
```diff
@@ -67,8 +70,10 @@ export class ApproveUi extends BaseElement {
 
         const result = await userClient.request({
             method: 'getTransaction',
-            params: { commandId: this.commandId },
+            params: { commandId: this.urlCommandId },
```
pawelstepien-da: I think we should add param / endpoint in user api to query by transactionId and use it here.
Otherwise if we have multiple txes with same commandId, when clicking `review` on not-latest one we would see data and act upon only the latest one. 
i.e. we have 2 txes with same commandId, older is pending, latest is executed. You want to enter review for the older one and approve or delete it, but not it's impossible, because it will always load latest and replace `this.transactionId`.

pawelstepien-da:
I think that in user-api we should use `transactionId` for identifying transaction instead of getting latest with `commandId` for methods like `sign` `execute` and `getTransaction`. Otherwise we only can read and write on latest transaction with `commandId` leaving previous as unusable zombies is `pending` or `signed` state.

core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
```diff
@@ -0,0 +1,180 @@
...
+    `.execute(db)
+
+    await db.schema.dropTable('transactions').execute()
+    await db.schema
+        .alterTable('transactions_new')
+        .renameTo('transactions')
+        .execute()
+}
+
+export async function down(db: Kysely<DB>): Promise<void> {
```
pawelstepien-da, line 109: Should we remove rows that have duplicate commandIds here? I would expect down-migration to fail if there are such rows.

PHOL-DA: i would hate to remove transactions just because of this, we could loose history... we could just put the transactionID into the commandID for downscaling...

PHOL-DA: @mjuchli-da what is your preference here ?

mjuchli-da: in the unlikely event of this down-migration we could just concat the `id` (uuid): `commandId: <commandId>-<id>`, provided `<commandId> != <id>`

core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
```diff
@@ -0,0 +1,183 @@
...
+    console.log('Reverting transactions primary key from id to command_id')
+
+    const pg = await isPostgres(db)
+    if (pg) {
+        await sql`
+            DROP INDEX IF EXISTS transactions_command_user_network_idx
+        `.execute(db)
+        await sql`
+            UPDATE transactions
+            SET command_id = command_id || ':' || id
```
pawelstepien-da, line 119: This happens only for postgres

core/wallet-store-sql/src/migrations/011-add-transaction-id-primary-key.ts
```diff
@@ -0,0 +1,183 @@
...
+            payload,
+            origin,
+            user_id,
+            network_id,
+            created_at,
+            signed_at,
+            external_tx_id
+        )
+        SELECT
+            status,
```
pawelstepien-da, line 166:
```suggestion
            command_id,
            status,
```

pawelstepien-da (approved):

PR #1630, opened 2026-04-15
feat: adjust typescript typings for wallet-sdk v1 fields

### LINKED ISSUE #1610: Adjust typescript typings for SDK fields
As a dev using wallet-sdk, I want to have the power to configure the sdk the way I feel comfortable with and to have returned an instance which holds all properties I'm interested in based on the config I'd previously passed. This way we can avoid the discrepancy of having both async and sync properties of sdk return object.

+ proper namespace type naming

https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1552 - bigger picture (enabling sdk extensibility)

mateuszpiatkowski-da:
### 1. Namespace Initialization at SDK Creation

Previously, namespaces like `token`, `amulet`, `asset`, and `events` were initialized separately using async method calls. Now they can be configured directly during SDK creation or added later via the new `.extend()` method.

#### Before

```typescript
const sdk = await SDK.create({ auth, ledgerClientUrl })
const token = await sdk.token(tokenConfig)
await token.utxos.list({ partyId })
```

#### After - Option 1: Configure at creation

```typescript
const sdk = await SDK.create({
    auth,
    ledgerClientUrl,
    token: tokenConfig,
})
await sdk.token.utxos.list({ partyId })
```

#### After - Option 2: Extend later

```typescript
const basicSDK = await SDK.create({ auth, ledgerClientUrl })
const extendedSDK = await basicSDK.extend({ token: tokenConfig })
await extendedSDK.token.utxos.list({ partyId })
```

#### Custom provider 

This PR also changes the structure of SDK.create arguments. Now, instead of `WalletSdkOptions | AbstractLedgerProvider` we enable the user to provide additional configuration alongside with a custom provider via [BasicSDKOptions](https://github.com/hyperledger-labs/splice-wallet-kernel/pull/1630/changes#diff-225584768e084ebfac1dce9b9ca7f856627c47766dc7dea1517072c9b2a15f89R32). 

##### Before

```typescirpt
const sdk = await SDK.create({
    auth: TOKEN_PROVIDER_CONFIG_DEFAULT,
    ledgerClientUrl: localNetStaticConfig.LOCALNET_APP_USER_LEDGER_URL,
})
const amulet = sdk.amulet(config)
// OR
const sdk1 = await SDK.create(ledgerProvider)
const amulet1 = sdk.amulet(config)
```

##### After

```typescirpt
const sdk = await SDK.create({
    amulet: config,
    ledgerProvider: provider
})
```

### 2. Introduction of Immutable `.extend()` Method

A new extension mechanism that allows adding namespaces to an existing SDK instance. **Importantly, it returns a new SDK instance rather than mutating the existing one.**

### 3. Type-Safe Namespace Tracking

The SDK now uses TypeScript's type system to track which namespaces are available at compile-time, preventing usage of unconfigured namespaces through the generic type parameter `SDKInterface<ExtendedItems>`.

### 4. Simplified API Surface

Direct namespace access on the SDK instance eliminates the intermediate namespace object:

- ~~`const token = await sdk.token(config); await token.utxos.list()`~~
- `await sdk.token.utxos.list()` ✓

### Major New Files

- `sdk/wallet-sdk/src/wallet/init/initializedSDK.ts` - New SDK initialization logic
- `sdk/wallet-sdk/src/wallet/init/types/*.ts` - Type definitions for SDK options and interfaces

### Major Refactored Files

- `sdk/wallet-sdk/src/wallet/sdk.ts` - Simplified SDK creation
- Migration guide documentation extensively updated

---

#### 1. **Predictability & Safety**

- Original SDK instance remains unchanged, preventing unexpected side effects
- No risk of shared state mutations affecting other parts of the application
- Easier to reason about program flow: `const newSDK = sdk.extend(config)` clearly creates a new instance
- Access to extended properties without the need of using non-null assertions or optional chaining

#### 2. **Functional Programming Patterns**

- Aligns with immutable data structures common in modern TypeScript/functional programming
- Enables pure, testable code without side effects
- Composable: `sdk.extend(config1).then(s => s.extend(config2))`

#### 3. **Type Safety**

- TypeScript can accurately track which namespaces are available on which instance
- Return type explicitly shows new capabilities: `SDKInterface<'token'>` vs `SDKInterface<'token' | 'amulet'>`
- Prevents accidental usage of namespaces that weren't configured
- **With mutable approach type generation wouldn't be even possible - once established, a variable cannot have its type changed**

#### 1. **Memory Overhead**

- Multiple extensions create a chain of objects which may be more difficult to handle

#### 2. **Developer Convenience**

- Requires reassignment or new variables:
    - `sdk = await sdk.extend(config)` or
    - `const newSdk = await sdk.extend(config)`
- More verbose than in-place mutation: `await sdk.extend(config); sdk.token.utxos.list()`

#### 4. **Variable Management**

- Must track multiple SDK variables if extending multiple times
- In long functions, can lead to `sdk1`, `sdk2`, `sdk3` naming or variable shadowing
- Temptation to shadow variables (`sdk = await sdk.extend()`) which can be confusing

## Recommendation

**The immutable approach is the better choice** for this SDK because:

1. **Type safety is paramount** in a financial/ledger SDK where mistakes are costly
2. **Single initialization pattern**: The SDK is typically configured once at application startup, not repeatedly extended
3. **Safety over convenience**: The benefits of predictability and safety outweigh the convenience of mutation
4. **Modern best practices**: Most modern TypeScript patterns favor immutability

### Why the Cons Are Minimal

- **Memory**: Extension typically happens once or twice per application lifecycle
- **Type safety benefits**: Clear type safety prevents entire classes of runtime errors
- **Documentation**: Clear examples can mitigate learning curve

## Migration Path

All example files demonstrate the new pattern:

- `docs/wallet-integration-guide/examples/scripts/*.ts` - 15+ example files updated
- `docs/wallet-integration-guide/examples/snippets/*.ts` - 20+ snippets updated
- Core test utilities updated in `setupTests.ts`

Developers migrating from v0 → v1 should follow the updated migration guide in `docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/`.


sdk/wallet-sdk/src/wallet/namespace/party/internal/index.ts
```diff
@@ -2,14 +2,14 @@
...
 import { Ops } from '@canton-network/core-provider-ledger'
-import { CommonCtx } from '../../sdk.js'
+import { SDKContext } from '../../../sdk.js'
 import { v4 } from 'uuid'
 import { PartyId } from '@canton-network/core-types'
-import { SDKLogger } from '../../logger/logger.js'
+import { SDKLogger } from '../../../logger/logger.js'
 
-export class InternalPartyNamespace {
+export class InternalPartyService {
```
alexmatson-da: I think we are in a similar spot as before (and sorry for bringing back the nitpickiness ;D) -- how exactly do you define the difference between a "service" and a "namespace"?

docs/wallet-integration-guide/examples/snippets/setupTests.ts
```diff
@@ -90,11 +93,11 @@ async function beforeEachSetup() {
...
     }
 
-    const token = await sdk.token(global.TOKEN_NAMESPACE_CONFIG)
+    const token = await sdk.token
 
-    const amulet = await sdk.amulet(global.AMULET_NAMESPACE_CONFIG)
+    const amulet = await sdk.amulet
 
-    const asset = await sdk.asset(global.ASSET_CONFIG)
+    const asset = await sdk.asset
```
alexmatson-da: could consider dropping these and just changing the examples to use `sdk.token.*...` directly

docs/wallet-integration-guide/examples/scripts/02-two-step-transfer/index.ts
```diff
@@ -69,6 +67,8 @@ if (senderAmuletUtxos.length === 0) {
 
 const transferTestScriptParameters: TransferTestScriptParameters = {
     sdk,
+    token: sdk.token,
+    amulet: sdk.amulet,
```
alexmatson-da: seems a bit redundant, any way this can just take an `sdk` and check if the token / amulet namespaces exist? Maybe something like `if ('token' in sdk) { ... } ` to narrow it

sdk/wallet-sdk/src/wallet/init/types/sdk.ts
```diff
@@ -0,0 +1,113 @@
...
+    'events',
+] as const satisfies ReadonlyArray<keyof ExtendedSDKOptions>
+
+// Compile-time validation that all keys from ExtendedSDKOptions are present in the tuple
+type _ValidateAllKeysPresent =
+    keyof ExtendedSDKOptions extends (typeof EXTENDED_SDK_OPTION_KEYS)[number]
+        ? true
+        : 'ERROR: Missing keys in EXTENDED_SDK_OPTION_KEYS'
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const _validateKeys: _ValidateAllKeysPresent = true
```
alexmatson-da: hmm... another possibility here is to make `EXTENDED_SDK_OPTION_KEYS` the source of truth instead, and then enforce the properties against the `ExtendedSDKOptions` type:

```ts
export const EXTENDED_SDK_OPTION_KEYS = [
    'amulet',
    'token',
    'asset',
    'events',
] as const

type EnforceKeys<
  K extends PropertyKey,
  T extends { [P in K | keyof T]: P extends K ? unknown : never }
> = T;

export type ExtendedSDKOptions = EnforceKeys<
    (typeof EXTENDED_SDK_OPTION_KEYS)[number],
    Readonly<{
        amulet: AmuletConfig
        token: TokenConfig
        asset: AssetConfig
        events: EventsConfig
    }>
>
```

its (arguably) cleaner and removes the need for the unused validator checker stuff. The downside though is its slightly less intuitive, since if the array is wrong, the error is pointed at the SDKOptions type definition instead. (Maybe a code comment for future developers would be enough to clarify that)

mateuszpiatkowski-da: Good idea! I'll apply the suggestion ;)

sdk/wallet-sdk/src/wallet/namespace/party/internal/index.ts
mateuszpiatkowski-da: I haven't rebased the branch in that moment of time, now it should be fine ;)

core/wallet-test-utils/src/otc-trade.ts
```diff
@@ -222,21 +223,21 @@ export class OTCTrade {
         const sdk = await SDK.create({
```
PHOL-DA: could we reuse the already instanciated SDK from line 55 (and just use the .extend method) ?

docs/wallet-integration-guide/examples/scripts/utils/index.ts
```diff
@@ -1,11 +1,13 @@
...
-    AmuletConfig,
-    AssetConfig,
     localNetStaticConfig,
-    TokenConfig,
 } from '@canton-network/wallet-sdk'
+import {
+    TokenConfig,
+    AmuletConfig,
+    AssetConfig,
+} from '../../../../../sdk/wallet-sdk/dist/wallet/init/types/config.js'
```
PHOL-DA: could we import these from the `@canton-network/wallet-sdk` instead ?

docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/ledger.rst
```diff
@@ -3,6 +3,21 @@ Ledger
 
 The ledger namespace is used for preparing, signing, and executing transactions and other Ledger API operations.
 
+Availability
+------------
+
+The ledger namespace is always available as part of the basic SDK interface. It's initialized automatically when you create an SDK instance and doesn't require additional configuration via ``extend()``.
+
+.. code-block:: javascript
```
PHOL-DA: can we do this as a snippet instead ? (that way we ensure that it is always compatible with latest changes)

docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/party.rst
```diff
@@ -8,6 +8,21 @@ Parties
 
 The party namespace provides methods to manage wallet parties on the Canton Network. In v1, the party namespace replaces the stateful party management from v0.
 
+Availability
+------------
+
+The party namespace is always available as part of the basic SDK interface. It's initialized automatically when you create an SDK instance and doesn't require additional configuration via ``extend()``.
+
+.. code-block:: javascript
```
PHOL-DA: can we do as a snippet instead aswell ?

docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/token.rst
```diff
@@ -8,6 +8,50 @@ Token
...
+Availability and extensibility
+------------------------------
+
+The token namespace is an extended namespace that requires configuration. You can initialize it either during SDK creation or later using the ``extend()`` method.
+
+**Option 1: Initialize during SDK creation**
+
+.. code-block:: javascript
+
+   const sdk = await SDK.create({
```
PHOL-DA: one more snippet 😄

docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/token.rst
```diff
@@ -8,6 +8,50 @@ Token
...
+
+   // token namespace is now available
+   await sdk.token.utxos.list({ partyId })
+
+**Option 2: Add token namespace later using extend()**
+
+.. code-block:: javascript
+
+   // Create basic SDK first
+   const basicSDK = await SDK.create({
```
PHOL-DA: another one 😃

docs/wallet-integration-guide/src/wallet-sdk-v1-migration-guide/user.rst
```diff
@@ -6,6 +6,21 @@ Users
...
 The user namespace provides methods for user management on the Canton Network.
 
+Availability
+------------
+
+The user namespace is always available as part of the basic SDK interface. It's initialized automatically when you create an SDK instance and doesn't require additional configuration via ``extend()``.
+
+.. code-block:: javascript
+
+   const sdk = await SDK.create({
```
PHOL-DA: another one

sdk/wallet-sdk/src/wallet/namespace/keys/index.ts
```diff
@@ -1,4 +1,31 @@
 // Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export * from './client.js'
+import {
```
PHOL-DA, line 4: these changes looks suspiciously like something that was just copied over from a different class. Just highlighting (feel free to ignore this comment if it was intentional)

sdk/wallet-sdk/src/wallet/namespace/ledger/dar/index.ts
```diff
@@ -1,4 +1,57 @@
 // Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-export * from './client.js'
+import { SDKContext } from '../../../sdk.js'
```
PHOL-DA, line 4: this also looks moved over (feel free to ignore if intetional)

PHOL-DA (approved):
Absolutely love everyhing in this PR! Left a few nit comments (you can address them in a secondary PR instead if you would prefer to merge this one)

thank you very much for doing this 😍

sdk/wallet-sdk/src/wallet/namespace/keys/index.ts
mateuszpiatkowski-da: Could you elaborate on that? I'm not exactly sure I fully understand the comment :D

PHOL-DA (approved):

PHOL-DA (approved):

PR #1638, opened 2026-04-17
fix(wallet-gateway-remote): fix wallet sync report

### LINKED ISSUE #1634: Fix wallet sync report
After wallet sync runs, server returns amount of added, updated and disabled wallets and UI displays a toast with that info. 
It should return changes in current iteration of sync, but if there were disabled wallets before, `updated` and `disabled` counts are increased by amount of already disabled wallets for each response.

pawelstepien-da:
Fix wrong counts of updated and disabled in message that shows after wallet sync - pre-existing disabled wallets were wrongly increasing updated and disabled counts in each call.
Add tests for sync report.
Reduce amount of calls to `/v2/parties/participant-id` to one per wallet sync.


wallet-gateway/remote/src/ledger/wallet-sync-service.ts
```diff
@@ -531,24 +545,36 @@ export class WalletSyncService {
...
-                    updated: [...walletsWithoutParty, ...rightsUpdatedWallets],
-                    disabled: disabled,
+                    added,
+                    updated,
+                    disabled,
                 },
                 'Wallet sync completed.'
             )
 
+            // added - non disabled wallets added in this sync call
```
PHOL-DA: would it make more sense to have these as descriptors on the returned object (`SyncWalletsResult`) instead of line comments ?

PHOL-DA (approved):

wallet-gateway/remote/src/ledger/wallet-sync-service.ts
pawelstepien-da: great idea

mjuchli-da (approved):
thanks!

PHOL-DA (approved):

PR #1639, opened 2026-04-17
feat(core-daml-codegen-helpers): created new helper classes

PHOL-DA:
split out the helper / companion usage from #1607

### DIFF
--- a/core/daml-codegen-helpers/package.json
+++ b/core/daml-codegen-helpers/package.json
@@ -1,0 +1,47 @@
+{
+    "name": "@canton-network/core-daml-codegen-helpers",
+    "version": "1.0.0",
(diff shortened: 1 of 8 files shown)


core/daml-codegen-helpers/package.json
```diff
@@ -0,0 +1,47 @@
+{
+    "name": "@canton-network/core-daml-codegen-helpers",
```
mjuchli-da, line 2: Maybe `-utils` instead of `-helpers`? (just as a consideration, helpers is equally fine)

mjuchli-da (approved):
Nice!

Maybe let's get @alexmatson-da 's opinion on it as well.

PR #1640, opened 2026-04-17
test(core-wallet-store-sql): test sqlite cascade deletes

pawelstepien-da:
Added unit tests validating sqlite cascade delete behavior
Set `foreign_keys = ON` explicitly for sqlite connections.

Note:
sqlite3 by default has `foreign_keys = OFF`, but we establish connection to db with better-sqlite3, which uses sqlite3 build with a different value for compilation flag `SQLITE_DEFAULT_FOREIGN_KEYS=1`. So it wasn't a problem before, but it's not obvious or advertised as a contract that will never change so I think it's worth setting explicitly and leaving a comment.

Only source I found pointing at this difference between sqlite3 and better-sqlite3: https://github.com/WiseLibs/better-sqlite3/blob/master/docs/compilation.md


core/wallet-store-sql/src/store-sql.ts
```diff
@@ -736,11 +736,18 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
...
 export const connection = (config: StoreConfig) => {
+    let database
     switch (config.connection.type) {
         case 'sqlite':
+            database = new Database(config.connection.database)
+            // normally sqlite3 has foreign_keys = OFF for each connection,
+            // but better-sqlite3 uses custom build with compile flag SQLITE_DEFAULT_FOREIGN_KEYS=1,
+            // making it ON by default.
+            // Set explicitly ON anyway as redundancy
+            database.pragma('foreign_keys = ON')
```
mjuchli-da, line 747: Should we also comment out this line so to implicitly test that bettersql sets it?

pawelstepien-da: So if better-sqlite changes default for foreign_keys to OFF

1. If I comment the line that sets it explicitly, then we start getting errors because of unexpected stuff in db. But my unit test should catch it.
2. If the line stays, then I would expect everything works business as usual and we don't know that changed.

Do we care that much that better-sqlite changed the default if we set it up the way we want anyway?
I would leave it, but I don't mind approch 1 either.

mjuchli-da: You're right, I missed that his is not the test file. As long as the unit test catches it its fine, and ofc better to set it explicitly here.

mjuchli-da (approved):

PR #1647, opened 2026-04-20
fix: preapproval test script + fetch status command

rukmini-basu-da:
We were previously ignoring the preapproval script in CI (this was leftover from the wallet sdk v0.

The test was falsely succeeding during the renewal step because `fetchedPreapprovalStatus?.expiresAt === fetchedStatusAfterRenew?.expiresAt` was always not equal to each other (even if is the same time) and failing at the `fetchStatus` step when checking for the cancellation. This updates fetchStatus to check for different conditions (create, renewal or cancel) and adds the tapInternal, which is required for the CI otherwise there will not be enough funds to create the transfer preapproval. 

Additionally, this updates the batch test execution to batches of 5 and modified the function slightly so that the tests can run concurrently. Previously, we would need batch 1 to all finish before the batch 2 scripts could start, now once one script finishes then the next one can start.

### DIFF
--- a/sdk/wallet-sdk/src/wallet/namespace/amulet/namespace.ts
+++ b/sdk/wallet-sdk/src/wallet/namespace/amulet/namespace.ts
@@ -66,6 +66,22 @@ export class AmuletNamespace {
         return [{ ExerciseCommand: tapCommand }, disclosedContracts]
     }
 
+    async tapOperator(amount: string, synchronizerId?: string) {
+        const [tapCommand, disclosedContracts] = await this.tap(
+            this.sdkContext.validatorParty,
+            amount
+        )
+
+        await this.ledger.internal.submit({
+            commands: [tapCommand],
+            disclosedContracts,
+            synchronizerId:
+                synchronizerId ??
+                this.sdkContext.commonCtx.defaultSynchronizerId,
+            actAs: [this.sdkContext.validatorParty],
+        })
+    }
+
     featuredApp: FeaturedAppNamespace = {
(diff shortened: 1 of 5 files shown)


sdk/wallet-sdk/src/wallet/namespace/amulet/namespace.ts
```diff
@@ -66,6 +66,34 @@ export class AmuletNamespace {
...
+     * Creates and submits a tap command for a specified amount for an internal party
+     * This is useful for tests and can only be used locally or against devnet
+     * @param amount The amount to be tapped.
+     * @param options Optional settings.
+     * @param options.synchronizerId defaults to the first connected synchronizer
+     * @param options.partyId optional internal party to receive tap, defaults to validator operator party
+     * @returns the updateId and completionOffset for the submitted tap command
+     */
+
+    async tapInternal(
```
alexmatson-da, line 79: Does this need to be supported in the SDK, or can it be implemented as a utility function in the docs/tests? 

Unless we think there is a strong production use case I would try to avoid expanding the public SDK surface with test-specific methods

rukmini-basu-da: Yeah, we can move it to a utility function in the tests. The main reason I have it here is because we don't expose any of the internal validator endpoints, but we have this as part of our context for other operations and didn't want to create a separate validator client in the tests/add an additional import.

alexmatson-da (approved):

PR #1653, opened 2026-04-21
fix(dapp-sdk): empty adapter list

mjuchli-da:
If an empty list of adapters (default- or additional) is provided, we currently fallback to the respective default values. Instead, we should only fall back in case of `undefined`.


PHOL-DA (approved):

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -310,12 +341,26 @@ export class DappSDK {
     }
 
     async connect(options?: DappSDKConnectOptions): Promise<ConnectResult> {
-        await this.ensureInit(options)
```
mjuchli-da, line 313: I will take this PR now as an opportunity to clean up the initialization process.

PHOL-DA (approved):

PHOL-DA (approved):

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -301,21 +305,70 @@ export class DappSDK {
...
+                )
+            )
+        }
+
+        await this.initPromise
     }
 
     async connect(options?: DappSDKConnectOptions): Promise<ConnectResult> {
-        await this.ensureInit(options)
+        // Prefer init({ ... }) once at startup. Passing options here remains supported
```
alexmatson-da, line 361: should we mark this as deprecated?

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -301,21 +305,70 @@ export class DappSDK {
...
+        // Register adapters and store them in the SDK instance.
+        if (options) {
+            this.configuredAdapters = normalizeConnectOptions(options)
+        }
+
+        // Create discovery and attempt restore.
+        if (!this.initPromise) {
+            this.initPromise = this.ensureDiscovery(
+                this.configuredAdapters
+            ).then(() => undefined)
```
alexmatson-da, line 340: the chained `then` is a bit random... assuming this is because the type of `initPromise` is `Promise<void>`, I wonder if you'd be able to drop this by changing the type to `Promise<unknown>`

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -301,21 +305,70 @@ export class DappSDK {
...
+            await this.initPromise
+            return
+        }
```
alexmatson-da, line 343: I think the control flow of this whole function can be restructured in a cleaner way..

```ts
init(options) {

  // initialize an empty promise if none exists, to allow chaining.
  // maybe even initialize it like this in the constructor from the very start?
  if (!this.initPromise) {
    this.initPromise = Promise.resolve(undefined)
  }
  
  // in the case of a _first time init_, this still only calls ensureDiscovery once due to the empty promise. For reconfigured inits, the discovery is called again with new options 
  if (options) {
    this.configuredAdapters = ...;
    this.initPromise = this.initPromise.then(() => this.ensureDiscovery(...))
  }
  
  await this.initPromise
}
```

sdk/dapp-sdk/src/sdk.ts
mjuchli-da: ah indeed! thx, learned something here

sdk/dapp-sdk/src/sdk.ts
mjuchli-da: `this.configuredAdapters = ... ` needs to be assigned prior to 
```
this.initPromise = this.ensureDiscovery(
                this.configuredAdapters
            )
```

PR #1682, opened 2026-04-28
feat: add dfns signing driver

noah-dfns:
Dfns Signing Driver

### DIFF
--- a/core/signing-dfns/src/dfns.ts
+++ b/core/signing-dfns/src/dfns.ts
@@ -1,0 +1,347 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { DfnsApiClient } from '@dfns/sdk'
+import { AsymmetricKeySigner } from '@dfns/sdk-keysigner'
+import { SigningStatus } from '@canton-network/core-signing-lib'
+import { pino } from 'pino'
+
+const logger = pino({ name: 'dfns-handler', level: 'debug' })
+
+export type DfnsCantonNetwork = 'Canton' | 'CantonTestnet'
+
+const CANTON_NETWORKS: ReadonlySet<string> = new Set<DfnsCantonNetwork>([
+    'Canton',
+    'CantonTestnet',
+])
+
+const DFNS_LIST_PAGE_SIZE = 50
+
+export interface DfnsCredentials {
+    credId: string
+    privateKey: string
+    authToken: string
+}
+
+export interface DfnsTransaction {
+    txId: string
+    status: SigningStatus
+    signature?: string
+    publicKey?: string
+    txHash?: string
+}
+
+export interface DfnsKey {
+    id: string
+    name: string
+    publicKey: string
+    network: DfnsCantonNetwork
+}
+
+export interface DfnsWalletInfo {
+    id: string
+    name: string
+    address: string
+    network: DfnsCantonNetwork
+    status: string
+}
+
+function isCantonNetwork(network: string): network is DfnsCantonNetwork {
+    return CANTON_NETWORKS.has(network)
+}
+
+export class DfnsHandler {
+    private client: DfnsApiClient
+    private orgId: string
+    private baseUrl: string
+
+    constructor(orgId: string, baseUrl: string, credentials: DfnsCredentials) {
+        this.orgId = orgId
+        this.baseUrl = baseUrl
+        this.client = this.createClient(credentials)
+    }
+
+    private createClient(credentials: DfnsCredentials): DfnsApiClient {
+        const signer = new AsymmetricKeySigner({
+            credId: credentials.credId,
+            privateKey: credentials.privateKey,
+        })
+
+        return new DfnsApiClient({
+            orgId: this.orgId,
+            authToken: credentials.authToken,
+            baseUrl: this.baseUrl,
+            signer,
+        })
+    }
+
+    /**
+     * Create a new Canton wallet.
+     */
+    public async createWallet(
+        name: string,
+        network: DfnsCantonNetwork = 'Canton'
+    ): Promise<DfnsKey> {
+        try {
+            const wallet = await this.client.wallets.createWallet({
+                body: { network, name },
+            })
+
+            const validators = await this.client.networks.listCantonValidators({
+                network: network === 'Canton' ? 'canton' : 'canton-testnet',
+            })
+
+            if (validators.items.length === 0) {
+                throw new Error(
+                    `No Canton validators available for network ${network}`
+                )
+            }
+
+            if (validators.items.length > 1) {
+                logger.warn(
+                    `Multiple validators found for network ${network}, picking the first one.`
+                )
+            }
+
+            const validatorId = validators.items[0].id
+            const activation = await this.client.wallets.activateWallet({
+                walletId: wallet.id,
+                body: { validatorId },
+            })
+            logger.info(
+                `Wallet ${wallet.id} activation initiated: ${activation.status}`
+            )
+
+            return {
+                id: wallet.id,
+                name: wallet.name || wallet.id,
+                publicKey: wallet.signingKey.publicKey,
+                network,
+            }
+        } catch (error) {
+            logger.error(error, 'Error creating Dfns wallet')
+            throw error
+        }
+    }
+
+    /**
+     * Iterate every Canton wallet on the org, paging through Dfns 50-at-a-time.
+     */
+    public async *iterateWallets(): AsyncGenerator<DfnsWalletInfo> {
+        let paginationToken: string | undefined
+        do {
+            const response = await this.client.wallets.listWallets({
+                query: {
+                    limit: DFNS_LIST_PAGE_SIZE,
+                    ...(paginationToken ? { paginationToken } : {}),
+                },
+            })
+
+            for (const wallet of response.items) {
+                if (
+                    !isCantonNetwork(wallet.network) ||
+                    wallet.status !== 'Active' ||
+                    !wallet.address
+                ) {
+                    continue
+                }
+
+                yield {
+                    id: wallet.id,
+                    name: wallet.name || wallet.id,
+                    address: wallet.address,
+                    network: wallet.network,
+                    status: wallet.status,
+                }
+            }
+
+            paginationToken = response.nextPageToken
+        } while (paginationToken)
+    }
+
+    public async listWallets(): Promise<DfnsKey[]> {
+        const keys: DfnsKey[] = []
+        try {
+            for await (const wallet of this.iterateWallets()) {
+                keys.push({
+                    id: wallet.id,
+                    name: wallet.name,
+                    publicKey: wallet.address,
+                    network: wallet.network,
+                })
+            }
+        } catch (error) {
+            logger.error(error, 'Error listing Dfns wallets')
+            throw error
+        }
+        return keys
+    }
+
+    public async getWallet(
+        walletId: string
+    ): Promise<DfnsWalletInfo | undefined> {
+        try {
+            const wallet = await this.client.wallets.getWallet({ walletId })
+
+            if (!isCantonNetwork(wallet.network) || !wallet.address) {
+                return undefined
+            }
+
+            return {
+                id: wallet.id,
+                name: wallet.name || wallet.id,
+                address: wallet.address,
+                network: wallet.network,
+                status: wallet.status,
+            }
+        } catch (error) {
+            logger.error(error, `Error fetching wallet ${walletId}`)
+            return undefined
+        }
+    }
+
+    /**
+     * Sign a Canton transaction using Dfns. The caller must supply a Dfns
+     * walletId (resolved at the gateway from the key identifier) and a
+     * deterministic externalTxId used as the Dfns idempotency key.
+     */
+    public async signTransaction(
+        walletId: string,
+        preparedTransaction: string,
+        externalTxId: string
+    ): Promise<DfnsTransaction> {
+        try {
+            const transactionHex = Buffer.from(
+                preparedTransaction,
+                'base64'
+            ).toString('hex')
+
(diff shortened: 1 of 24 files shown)


core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+                logger.warn(
+                    `Multiple validators found for network ${network}, picking the first one.`
+                )
+            }
+
+            // For simplicity, pick the first validator
+            const validatorId = validators.items[0].id
+            const activation = await client.wallets.activateWallet({
+                walletId: wallet.id,
+                body: { validatorId },
```
PHOL-DA: is a validator required for activating the wallet? The wallet gateway act as a middle man for Bring your own validator so the entire flow (as needed) would require creating the wallet on a potentially unknown to dfns

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+            })
+            logger.info(
+                `Wallet ${wallet.id} activation initiated: ${activation.status}`
+            )
+
+            const walletName = wallet.name || wallet.id
+
+            // Cache the wallet
+            const info: DfnsWalletInfo = {
```
PHOL-DA: why do you see the need to add this to a cache ? The state of wallets is handled outside of the signing driver.

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public async listWallets(userId: string | undefined): Promise<DfnsKey[]> {
+        const client = this.getClient(userId)
+        const keys: DfnsKey[] = []
+
+        try {
+            const response = await client.wallets.listWallets({})
+
+            for (const wallet of response.items) {
+                // Filter for Canton network wallets only
```
PHOL-DA: this filtering can't be done on the api side ?

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+    /**
+     * List all Canton wallets available to the user.
+     * Updates the internal wallet caches.
+     */
+    public async listWallets(userId: string | undefined): Promise<DfnsKey[]> {
+        const client = this.getClient(userId)
+        const keys: DfnsKey[] = []
+
+        try {
+            const response = await client.wallets.listWallets({})
```
PHOL-DA: according to the [docs](https://docs.dfns.co/api-reference/wallets/list-wallets#list-wallets) there is a limit of 50, so this should probably be adjusted to handle pagination as well.

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+     * 1. Commands mode - pass commands and disclosed contracts, Dfns prepares the tx
+     * 2. Binary mode - pass pre-prepared transaction (hex), Dfns signs it
+     *
+     * For the signing driver, we use binary mode with the prepared transaction.
+     */
+    public async signTransaction(
+        userId: string | undefined,
+        preparedTransaction: string,
+        keyIdentifier: KeyIdentifier,
+        externalTxId?: string
```
PHOL-DA: i would not make the externalTxId optional

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+     */
+    public async *listTransactions(
+        userId: string | undefined,
+        walletId: string
+    ): AsyncGenerator<DfnsTransaction> {
+        const client = this.getClient(userId)
+        const wallet = this.walletsById.get(walletId)
+
+        try {
+            const response = await client.wallets.listTransactions({
```
PHOL-DA: this one also seems to have pagination that needs to be handled.

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+            )
+            throw error
+        }
+    }
+
+    /**
+     * Enable transfer pre-approval for a Canton wallet.
+     * This allows automatic approval of Canton coin transfers.
+     */
+    public async enableTransferPreapproval(
```
PHOL-DA: The wallet is not hosted on dfns, so this should not be possible ?

core/signing-dfns/src/index.test.ts
```diff
@@ -0,0 +1,181 @@
...
+            txIds: ['test'],
+        })
+
+        // Not implemented yet, returns empty object
+        expect(result).toEqual({})
+    })
+})
+
+// Integration tests that require real Dfns credentials
+describe.skip('DfnsSigningDriver Integration', () => {
```
PHOL-DA: this we would rather do as playwright e2e test, but this can done later on in a second pr, remove this for now

core/signing-dfns/src/index.ts
```diff
@@ -0,0 +1,385 @@
...
+    credId: z.string().min(1),
+    privateKey: z.string().min(1),
+    authToken: z.string().min(1),
+})
+
+const DfnsConfigSchema = z.object({
+    orgId: z.string().min(1),
+    baseUrl: z.string().url(),
+    defaultCredentials: DfnsCredentialsSchema.optional(),
+    userCredentials: z.map(z.string(), DfnsCredentialsSchema),
```
PHOL-DA: for the first iteration it is one user pr. wallet gateway so i don't believe this mapping would work (we can remove it and just use default credentials).

core/signing-dfns/src/index.ts
```diff
@@ -0,0 +1,385 @@
...
+    userWallets: z.map(z.string(), z.string()).optional(),
```
PHOL-DA: the wallets and state is managed outside of the signing driver, so i would simply remove this.

core/signing-dfns/src/index.ts
```diff
@@ -0,0 +1,385 @@
...
+    defaultCredentials?: DfnsCredentials
+    userCredentials: Map<string, DfnsCredentials>
+    // Optional mapping of userId -> walletId for transaction lookups
+    userWallets?: Map<string, string>
+}
+
+const DfnsCredentialsSchema = z.object({
+    credId: z.string().min(1),
+    privateKey: z.string().min(1),
+    authToken: z.string().min(1),
```
PHOL-DA, line 45: what is the purpose of the authToken here and why is it injected through config ? I assume this is some kind of bearer token and would have to be recomputed or regenerated using the privateKey and credId on a regular basis in case of expiry ?

core/signing-dfns/src/index.ts
```diff
@@ -0,0 +1,385 @@
...
+                        params.keyIdentifier,
+                        params.internalTxId
+                    )
+
+                    const result: Transaction = {
+                        txId: tx.txId,
+                        status: tx.status,
+                    }
+
+                    // Canton updateId is returned as txHash
```
PHOL-DA: how can you return the updateId when you are not the one submitting the transaction ?

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,479 @@
...
+    /**
+     * Create a new Canton wallet.
+     * @param userId - The user ID for credential lookup
+     * @param name - Human-readable name for the wallet
+     * @param network - 'Canton' or 'CantonTestnet' (defaults to 'Canton')
+     */
+    public async createWallet(
+        userId: string | undefined,
+        name: string,
+        network: 'Canton' | 'CantonTestnet' = 'Canton'
```
PHOL-DA: i would suggest to use something akin to [CantonCaip2](https://github.com/hyperledger-labs/splice-wallet-kernel/blob/922271a9072033bdec866a315bfbd02dab437782/core/signing-blockdaemon/src/signing-api-sdk.ts#L13)

core/signing-dfns/src/index.ts
```diff
@@ -0,0 +1,385 @@
...
+             * Get the current configuration with secrets hidden.
+             */
+            getConfiguration: async (): Promise<GetConfigurationResult> => {
+                const hideCredentialSecrets = (
+                    credentials: DfnsCredentials | undefined
+                ): DfnsCredentials | undefined => {
+                    if (!credentials) return undefined
+                    return {
+                        credId: credentials.credId,
+                        privateKey: '***HIDDEN***',
```
PHOL-DA: this will break, since we rely on these variables for admin update operations where we `getConfiguration` and then call `setConfiguration` based on the received value from the first call

core/signing-lib/src/SigningDriverStore.ts
```diff
@@ -70,7 +70,7 @@ export interface SigningKey {
     id: string
     name: string
     publicKey: string
-    privateKey?: string // Optional for external providers like Fireblocks
+    privateKey?: string // Optional for external providers like Dfns and Fireblocks
```
PHOL-DA:
```suggestion
    privateKey?: string // Optional for external signing providers```
to keep it future agnostic

File: wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/dfns-wallet-allocator.ts [hunk 16]
```diff
@@ -0,0 +1,105 @@
...
+            `Error from signing driver: ${result.error_description}`
+        )
+    }
+    return result
+}
+
+/**
+ * Dfns provisions Canton wallets through its validator integration: when we ask
+ * Dfns to create a key it also activates the wallet on Canton, so the Gateway
+ * does not run its own topology-transaction flow. We just record the wallet.
```
PHOL-DA: Wallet gateway needs to run its own topology flow and wallets should not be created on dfns validator, this breaks the fundamentals of the bring-your-own-validator and does not align with institutional setups as needed.

PHOL-DA (changes requested):
the dfns signing driver should not handle the submission of topology or other transactions, that lies with the wallet gateway against the registered validators that it is connected to. dfns should resolve to handle secure key storage and transactions signage flow (with user approvals etc.)

core/signing-dfns/src/dfns.ts
noah-dfns: The wallet activation requires a signature as it's an external party. When we activate a wallet the validator id (from dfns) is required.

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,317 @@
...
+        }
+    }
+
+    public async *iterateKeys(): AsyncGenerator<DfnsKey> {
+        let paginationToken: string | undefined
+        do {
+            const response = await this.client.keys.listKeys({
+                query: {
+                    limit: DFNS_LIST_PAGE_SIZE,
+                    ...(paginationToken ? { paginationToken } : {}),
```
PHOL-DA, line 121: would it make sense to add the "owner" parameter as well? https://docs.dfns.co/api-reference/keys/list-keys

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,317 @@
...
+        hashBase64: string,
+        externalId: string
+    ): Promise<DfnsSignature> {
+        try {
+            const result = await this.client.keys.generateSignature({
+                keyId,
+                body: {
+                    kind: 'Hash',
+                    hash: base64ToHex(hashBase64),
+                    blockchainKind: 'Canton',
```
PHOL-DA, line 215: you don't need the 'network' value here ? https://docs.dfns.co/api-reference/keys/generate-signature

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/dfns-wallet-allocator.ts
```diff
@@ -0,0 +1,218 @@
...
+        private signingDriver: SigningDriverInterface
+    ) {}
+
+    async createWallet(
+        userId: UserId,
+        email: string | undefined,
+        partyHint: PartyHint,
+        primary: Primary = false
+    ): Promise<Wallet> {
+        const driver = this.signingDriver.controller(email)
```
PHOL-DA: fyi email might not be unique, but userId will

PHOL-DA: in some cases `email === userid`, depends on the IDP

core/signing-dfns/src/dfns.ts
```diff
@@ -0,0 +1,317 @@
...
+                    externalId,
```
PHOL-DA: if i reads the [docs](https://docs.dfns.co/api-reference/keys/generate-signature) correctly, then this is purely for idempotent behavior, right ? If that is the case then i would not pass it around as an argument and just generate a random UUID in here (or reuse the hash to be signed). It is no problem that the same transaction gets signed multiple times since canton has command deduplication.

PHOL-DA (commented):
@noah-dfns very nice adjustments! once final comments are cleared / questions addresses then i will approve this PR 👍

core/signing-dfns/src/dfns.ts
noah-dfns: Resolved

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/dfns-wallet-allocator.ts
noah-dfns: Resolved

core/signing-dfns/src/dfns.ts
noah-dfns: Owner is for delegated wallets which wouldn't quite make sense here.

core/signing-dfns/src/dfns.ts
noah-dfns: You only need one, kind here just ensures we use an eddsa key

core/signing-dfns/src/index.ts
noah-dfns: In the Service Account and PAT model it is a read bearer token that is good for up to 2 years. It does need to be refreshed after that mark.

PHOL-DA (approved):
🥳

PR #1688, opened 2026-04-29
chore: fix a few small ci log files

PHOL-DA:
- [X] adds wallet gateway logs to artifacts.
- [X] remove the "example-ping-logs" (which is just build logs and not needed)
- [X] properly prefixed the "playwright-report" with "example-ping"


fayi-da (approved):
thanks

PR #1689, opened 2026-04-29
chore: improve otc script for testing allocations

fayi-da:
- add 3rd party(charlie) so we can demo multiple legs with alice as sender
- made expected allocation a computed property which should make extending the script easier
- small refactoring to heal some OCD

### DIFF
--- a/core/wallet-test-utils/src/otc-trade.ts
+++ b/core/wallet-test-utils/src/otc-trade.ts
@@ -25,22 +25,30 @@ export class OTCTrade {
     private venue: PartyId
     private alice: PartyId
     private bob: PartyId
+    private charlie: PartyId
     private logger: Logger
     private sdk: SDKInterface<'asset'> | null = null
+    private expectedAllocationCount = 0
 
     constructor(args: {
         logger: Logger
         venue: PartyId
         alice: PartyId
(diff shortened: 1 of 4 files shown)


core/wallet-test-utils/src/otc-trade.ts
```diff
@@ -205,10 +199,60 @@ export class OTCTrade {
...
+
+        this.logger.info(`${approverName} accepted OTCTradeProposal`)
+
+        return otcpCid
     }
 
-    async settle(args: { otcTradeCid: string }): Promise<void> {
+    async settle(args: {
+        otcTradeCid: string
+        expectedAllocationCount?: number
```
pawelstepien-da: Does it make sense to have this optional override of `expectedAllocationCount` when the legs are hardcoded?

pawelstepien-da (approved):

core/wallet-test-utils/src/otc-trade.ts
fayi-da: Yeah good catch. I don't need this anymore and can rely on `this.expectedAllocationCount`. Thanks!

pawelstepien-da (approved):

PR #1690, opened 2026-04-29
chore: capture proper wg logs

PHOL-DA:
latest run works: https://github.com/hyperledger-labs/splice-wallet-kernel/actions/runs/25118875810?pr=1690

attached file has proper content


fayi-da (approved):
👍

.github/workflows/build.yml
```diff
@@ -261,7 +261,10 @@ jobs:
               run: |
                   BLOCKDAEMON_API_URL="${{ vars.BLOCKDAEMON_API_URL }}" \
                   BLOCKDAEMON_API_KEY="${{ vars.BLOCKDAEMON_API_KEY }}" \
-                  yarn nx run @canton-network/wallet-gateway-remote:start > "wallet-gateway-remote.log" 2>&1 &
+                  yarn nx run @canton-network/wallet-gateway-remote:start &
+
+            - name: Capture remote PM2 logs
+              run: yarn pm2 logs remote --raw > "wallet-gateway-remote.log" 2>&1 &
```
alexmatson-da: I dont think this works because `yarn nx run ...` starts a raw NodeJS process outside of pm2. We probably want to just redirect the stdout to a file to capture the logs, or use the pm2 start:all script in CI (last time I tried this, I ran into CI problems but dont recall specifics...)

PHOL-DA: yeah im testing a bit around, the problem is that the yarn nx run just pushes the build output (and we wan't the run output)

ecosystem.config.js
```diff
@@ -9,7 +9,7 @@ const sharedEnvDevelopment = {
 export const apps = [
     {
         name: 'remote',
-        script: 'yarn workspace @canton-network/wallet-gateway-remote dev',
+        script: 'yarn workspace @canton-network/wallet-gateway-remote start',
```
alexmatson-da: maybe better to write a separate ecossytem file for CI? `ecosystem.config.ci.js` or something. 

That way local development can still enjoy the benefit of live reloading the WG service when changing the source files

fayi-da (approved):

fayi-da (approved):

PR #1691, opened 2026-04-29
chore: core-amulet-service test

rukmini-basu-da:
image


core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+                        round: {
+                            number: '19',
+                        },
+                        issuancePerValidatorRewardCoupon: '0.2',
+                    },
+                    created_event_blob:
+                        'CgMyLjESmwQKRQAnijL23GttusO+DHGLunrd3gPF8cgMzC5eHxCRqbKJLsoSEiAhj1RmawmW7HyvE+0aG+8p/0gdW2pCGYQj6I2wIPh2VxINc3BsaWNlLWFtdWxldBplCkA2YzU4MDJmODY3MDlhMGFkNDc4NGFmODFmMGJhYjQwZjMwNzBiMmY1ODEyOGQ4ODQzZGExZTE3ODRjMTQ3ODAyEgZTcGxpY2USBVJvdW5kGhJJc3N1aW5nTWluaW5nUm91bmQi2wFq2AEKTQpLOklEU086OjEyMjBlYmEzNGQxM2FiMjIzZjFhOTMzZmJkZTRlNzYwZGZmOWEzYTk2NTAzMTE1MWIwOTE4Y2E5NzM5NDI0NDA2ZGVkCgoKCGoGCgQKAhgmChAKDjIMMC4yMDAwMDAwMDAwChIKEDIOMTAwLjAwMDAwMDAwMDAKEAoOMgwwLjYwMDAwMDAwMDAKEQoPMg01Ny44Mzg2NjA1Nzg0CgsKCSnxo9emiVAGAAoLCgkp8S9e7olQBgAKFgoUUhIKEDIONTcwLjAwMDAwMDAwMDAqSURTTzo6MTIyMGViYTM0ZDEzYWIyMjNmMWE5MzNmYmRlNGU3NjBkZmY5YTNhOTY1MDMxMTUxYjA5MThjYTk3Mzk0MjQ0MDZkZWQ58V0Ug4lQBgBCKgomCiQIARIgxX5IC4Ip9KlGprdg1EnCFCeB0pXuHqzM2YZ4iL/sdC0QHg==',
+                    created_at: '2026-04-28T18:36:40.155633Z',
+                },
+                domain_id:
```
pawelstepien-da: I would move those big mocks to a separate file, so it's easier to navigate tests.

core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+        const result =
+            await service.getFeaturedAppsByParty('featured-party-123')
+
+        expect(mockScanProxyClient.get).toHaveBeenCalledWith(
+            '/v0/scan-proxy/featured-apps/{provider_party_id}',
+            { path: { provider_party_id: 'featured-party-123' } }
+        )
+        expect(result).toEqual(mockResponse.featured_app_right)
+    })
+    it('should correctly call the scan proxy and return the transfer pre-approval', async () => {
```
pawelstepien-da: I think that's a duplicate of the first test

core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+    it('should correctly fetch the member traffic status', async () => {
+        const mockResponse = {
+            traffic_status: {
+                actual: { total_consumed: 0, total_limit: 1600000 },
+                target: { total_purchased: 1600000 },
+            },
+        }
+
+        const domainId = 'fakeDomainId'
+        const memberId = 'PAR::fakeparticipant'
```
pawelstepien-da, line 111: Maybe let's do version without `PAR::` and then in the expect at the end of that test we check that `member_id: 'PAR::fakeparticipant'` to validate that ScanProxy is called with member_id passed through `toQualifiedMemberId`

core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+                domain_id:
+                    'global-domain::1220eba34d13ab223f1a933fbde4e760dff9a3a965031151b0918ca9739424406ded',
+            },
+        ],
+    }
+
+    let service: AmuletService
+
+    beforeEach(() => {
+        vi.clearAllMocks()
```
pawelstepien-da: `vi.clearAllMocks` resets mocks' calls (how many time and with what they were called) but does not reset return values. As return values are set inline in test cases, I would use `vi.resetAllMocks`, which returns mocks to default value, which is `vi.fn()` - same as in the moment of assignment, removing potential side effects between tests.

core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+        ])
+
+        const [command, dc] = await service.renewTransferPreapproval(
+            'cId',
+            '6c5802f86709a0ad4784af81f0bab40f3070b2f58128d8843da1e1784c147802:Splice.AmuletRules:TransferPreapproval',
+            'provider-party',
+            amuletRules.amulet_rules.domain_id
+        )
+
+        expect(command.choice).toEqual('TransferPreapproval_Renew')
```
pawelstepien-da, line 282: Could you validate command output in more detail? Especially `choiceArgument` shape.
Also would be good to validate cases with and without `newExpiresAt`passed to `renewTransferPreapproval`

core/amulet-service/src/amulet-service.test.ts
```diff
@@ -0,0 +1,623 @@
...
+        const [command, tapDisclosedContracts] = await service.createTap(
+            'receiverParty',
+            '2000',
+            amuletRules.amulet_rules.contract.payload.dso,
+            'Amulet',
+            'registry'
+        )
+
+        expect(tapDisclosedContracts).toHaveLength(1)
+        expect(command.choice).toBe('AmuletRules_DevNet_Tap')
```
pawelstepien-da, line 244: Also here I would like to see more thorough validation of command shape. Feel free to log what actual return value is, copy it and use `expect(xxx).toEqual({ ... })` for exact object matching.
Also would be good to see what is passed to `fetchTransferFactoryChoiceContext` as `choiceArgs`. I would also copy actual object, remove date fields as they would differ with each run and use `expect.objectContaining` which is similar to `toEqual`, but allows for more fields being in actual value than in expected object.

pawelstepien-da (commented):
LGTM! 
Please bump values in `vitest.config.ts` in test.coverage.thresholds to 80 for all categories, so CI will catch coverage regression and we are good to merge.

pawelstepien-da (approved):

PR #1709, opened 2026-05-05
chore: run curl to disable policies in ci

PHOL-DA:
blockdaemon enables all policies with each deployment, with this we should disable them before each run

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -267,7 +267,6 @@ jobs:
 
             - name: Start remote WK
               run: |
-                  set -euo pipefail
                   BLOCKDAEMON_API_URL="${{ vars.BLOCKDAEMON_API_URL }}"
                   BLOCKDAEMON_API_KEY="${{ vars.BLOCKDAEMON_API_KEY }}"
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -46,6 +46,29 @@ jobs:
...
+        runs-on: ubuntu-latest
+        steps:
+            - name: Clear Blockdaemon policies
+              run: |
+
+                  BLOCKDAEMON_API_URL="${{ vars.BLOCKDAEMON_API_URL }}"
+                  BLOCKDAEMON_POLICY_BASE_URL="${BLOCKDAEMON_API_URL%/api/cwp/canton}"
+
+                  # this clear policies, with each deployment blockdaemon renables them
+                  curl -s -X PUT "$BLOCKDAEMON_POLICY_BASE_URL/api/v2/policies/12" \
```
mjuchli-da, line 59: /12 looks like an awfully specific id – which might change in the future?...

mjuchli-da (approved):

.github/workflows/build.yml
PHOL-DA: i have no idea, those was just the two commands BD gave me to fix policy related problem 🤷

PR #1714, opened 2026-05-06
[PoC] Ressaign correction

jarekr-da:
THis is PoC mostly genrated using agent  - just to show working example
This is only demonstration: do not merge
1. removed one manual reassignment - maybe second also can be 
(one manual reassignment needed)
2. removed party.ts 
3. improved logs


sdk/wallet-sdk/src/wallet/namespace/party/external/signed.ts
```diff
@@ -39,6 +39,7 @@ export class SignedPartyCreationService {
         options?: Partial<{
             expectHeavyLoad?: boolean
             grantUserRights?: boolean
+            skipExistenceCheck?: boolean
```
jarekr-da: this is not a good solution - only quickly generated,
actually change code below to check if  party exists on given synchronizer (if provided)

sdk/wallet-sdk/src/wallet/namespace/ledger/internal/index.ts
```diff
@@ -22,9 +22,101 @@ type InternalOperationParams<Operation extends AllowedOperation> = Required<
         Omit<OperationBodyRequest<Operation>, UnusedParams | RequiredParams>
     >
 
+export interface ReassignParams {
```
jarekr-da, line 25: Actually we do not use this function in this version of example....but only cause we did quite artificial hack with Bob Transfer  - so I keep it  -> Maybe it is actually better to keep it

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_trade_ops.ts
```diff
@@ -734,23 +524,123 @@ export async function settleOtcTrade(
...
-    aliceTokenCid: string
     tokenRulesCid: string
-    tokenRulesContract: AcsContractEntry
 }
 
 /**
- * Self-transfers Alice's TestToken on app-synchronizer via TransferFactory_Transfer.
- * Both TokenRules (the factory) and Alice's Token have been reassigned to app-synchronizer
- * in step 11c, so the submission targets app-synchronizer.
+ * Bob self-transfers a portion of his remaining TestToken holding from global
```
jarekr-da: This is "hack" that allows to get away without use of sdk.ledger.internal reassign.
causes auto-reassign ot TokenRules to app-synchronizer
but really feels like hack

PR #1715, opened 2026-05-06
test: db migrations

### LINKED ISSUE #1655: Database testing
1. Create automated tests that validate changes to db schema and data during migrations for both postgres and sqlite.
2. Make Wallet Gateway E2E tests run against both postgres and sqlite.

pawelstepien-da:
In SQL stores I added tests for each migration that validates that expected effects are applied to schema, relations and data in case of up/down. They are run against SQLite and Postgres databases.

I made wallet gateway e2e tests with ping example run against both SQLite and Postgres database.

I added a mechanism validating that every migration has a corresponding test file (same file name), to assure every migration was tested against both SQL dialects before merging.

I added mechanism that validates immutability of migrations - each migration file gets hashed and put into migrations.lock of a store. If a migration gets unintentionally edited than CI will fail. To add a new migration, or potentially conciously edit existing, use `yarn migrations:update-lock`


mjuchli-da (approved):

mjuchli-da:
Great work, and I am generally in favor of the integrity check to verify that the migration's content was not changed.
However, I also think there are valid concerns laid out here:https://github.com/kysely-org/kysely/issues/984#issuecomment-2102387870

In a follow-up PR (low prio, #1798), we might want to feature-flag the integrity in case we need to alter a past migration.

PR #1692, opened 2026-04-30
feat: signMessage end-to-end and Sign-In-With-Canton over WalletConnect

ganchoradkov:
# feat(user-api + sdk): `signMessage` end-to-end and Sign-In-With-Canton over WalletConnect

## Summary

Adds first-class arbitrary message signing to the wallet stack and uses it to implement a Sign-In-With-X style auth flow ("Sign In With Canton") on the dApp SDK's WalletConnect adapter.

Two cohesive changes shipped together:

1. **`signMessage` on the signing driver + user-api** — wallets can now sign arbitrary UTF-8 messages with their Ed25519 key without going through the prepared-transaction pipeline.
2. **Sign In With Canton (SIWX) on the WalletConnect adapter** — when a dApp configures `signInWithCanton`, the adapter automatically issues a `canton_signMessage` request right after the WC session is approved, returning a SIWX-formatted message + signature to the dApp.

## End-to-end flow

```mermaid
sequenceDiagram
    autonumber
    participant dApp as dApp<br/>(examples/ping)
    participant Adapter as WalletConnectAdapter<br/>(sdk/dapp-sdk)
    participant WC as WalletConnect<br/>relay
    participant Wallet as Wallet<br/>(examples/walletconnect)
    participant Gateway as Gateway user-api<br/>(wallet-gateway/remote)
    participant Driver as WALLET_KERNEL driver<br/>(core/signing-internal)

    Note over dApp,Adapter: dApp configures signInWithCanton<br/>{ domain, uri, version, nonce, ... }

    dApp->>Adapter: connect()
    Adapter->>WC: signClient.connect()
    WC-->>Adapter: pairing URI
    Adapter-->>dApp: showUriInPopup(uri) (QR + URI)

    Wallet->>WC: pair(uri)
    WC->>Wallet: session_proposal
    Wallet-->>WC: approve(session)
    WC-->>Adapter: session approved

    Note over Adapter: signInWithCanton is set →<br/>compose SIWX message

    Adapter->>Adapter: composeSIWXMessage({domain, uri, nonce,<br/>account, chainId, ...})
    Adapter->>WC: request canton_signMessage<br/>{ message }
    WC->>Wallet: session_request canton_signMessage

    Wallet->>Gateway: callUserApi("signMessage",<br/>{ message, partyId? })
    Gateway->>Gateway: resolve wallet (primary or by partyId)<br/>assert signingProviderId === WALLET_KERNEL
[...]

PR #1718, opened 2026-05-07
feat: signMessage end-to-end and Sign-In-With-Canton over WalletConnect

### LINKED ISSUE #1754: Integration of WalletConnect
Establish WalletConnect as a first-class citizen. This integration broadens our ecosystem's reach, ensuring compatibility across a wider array of Wallet Providers.

mjuchli-da:
replaces #1692


sdk/dapp-sdk/src/sdk-controller.ts
```diff
@@ -153,13 +154,61 @@ export const dappSDKController = (provider: DappAsyncProvider) =>
...
+        ): Promise<SignMessageResult> => {
+            const response = await provider.request({
                 method: 'signMessage',
                 params,
-            }),
+            })
+            const { userUrl } = response
+            popup.open(userUrl)
+
+            const messageId = new URL(userUrl).searchParams.get('messageId')
```
pawelstepien-da, line 165: Could we get messageId from `response.messageId` instead of extracting it from URL query param?
Seems more type safe and flexible for wallets

examples/ping/src/hooks/useConnect.ts
```diff
@@ -14,7 +14,18 @@ const loopAdapter = new LoopAdapter({
...
 const wcProjectId = import.meta.env.VITE_WC_PROJECT_ID as string
 const wcAdapter = wcProjectId
-    ? WalletConnectAdapter.create({ projectId: wcProjectId })
+    ? WalletConnectAdapter.create({
+          projectId: wcProjectId,
+          signInWithCanton: {
+              domain: 'http://localhost:3000',
+              uri: 'http://localhost:3000/login',
+              version: '1.0.0',
+              nonce: '1234567890',
```
alexmatson-da: how would a dapp determine these values generally? Specifically version/nonce?

mjuchli-da: @ganchoradkov would be good to have your take on this here.

ganchoradkov: hey, these params were hardcoded for the example, but apps would be able to set them for their apps e.g. `uri`, `domain`, unique `nonce` per each SIWX etc

alexmatson-da (approved):

PHOL-DA (approved):

PR #1740, opened 2026-05-13
refactor(wallet-sdk): remove default synchronizer auto-selection

jarekr-da:
1. No automatic selection of synchonizer in wallet sdk code - if synchronizer is not explicitly provided by caller - we rely on canon ledger choice of synchronizer. There is no extra logic of choosing synchonizer in wallet sdk
2. tests code changed to explicitly set synchronizer

Note:

This `complicates` all existing tests. Code must now explicitly select synchronizer in many calls.

Alternative would be to start some tests in single synchronizer (but this also means, such test would fail randomly if run on multi synchronizer localnet).


jarekr-da:
Check - where syncId is really needed by canton.

sdk/wallet-sdk/src/wallet/namespace/party/external/service.ts
```diff
@@ -32,14 +32,18 @@ export class ExternalPartyNamespace {
             this.resolveParticipantUids(
                 options?.confirmingParticipantEndpoints ?? []
             ),
-            options?.synchronizerId || this.resolveSynchronizerId(),
+            options?.synchronizerId,
```
mateuszpiatkowski-da: I don't think this is needed here, as this is not async

sdk/wallet-sdk/src/wallet/namespace/party/external/service.ts
```diff
@@ -32,14 +32,18 @@ export class ExternalPartyNamespace {
...
         ]).then(
             ([
                 observingParticipantUids,
                 otherHostingParticipantUids,
                 synchronizerId,
-            ]) =>
-                this.ctx.ledgerProvider.request<Ops.PostV2PartiesExternalGenerateTopology>(
+            ]) => {
+                if (!synchronizerId)
```
mateuszpiatkowski-da: you can simply pass `options.synchronizerId`

sdk/wallet-sdk/src/wallet/namespace/party/external/signed.ts
```diff
@@ -242,9 +242,11 @@ export class SignedPartyCreationService {
         } = options
         const ledgerProvider = defaultLedgerProvider ?? this.ctx.ledgerProvider
         try {
-            const synchronizerId =
-                this.createPartyOptions?.synchronizerId ??
-                this.ctx.defaultSynchronizerId
+            const synchronizerId = this.createPartyOptions?.synchronizerId
+            if (!synchronizerId)
+                throw new Error(
```
mateuszpiatkowski-da: use `sdk.ctx.error.throw`

sdk/wallet-sdk/src/wallet/namespace/amulet/traffic.ts
```diff
@@ -46,6 +48,11 @@ export class TrafficNamespace {
     }): Promise<PreparedCommand> {
         const { buyer, ccAmount, inputUtxos } = params
         const migrationId = params.migrationId ?? 0
+        const synchronizerId = params.synchronizerId
+        if (!synchronizerId)
+            throw new Error(
```
mateuszpiatkowski-da: use `sdk.ctx.error.throw`

sdk/wallet-sdk/src/wallet/namespace/amulet/traffic.ts
```diff
@@ -12,9 +12,11 @@ export class TrafficNamespace {
     async status(
         params?: Partial<{ memberId?: string; synchronizerId?: string }>
     ) {
-        const synchronizerId =
-            params?.synchronizerId ||
-            this.sdkContext.commonCtx.defaultSynchronizerId
+        const synchronizerId = params?.synchronizerId
+        if (!synchronizerId)
+            throw new Error(
```
mateuszpiatkowski-da: use `sdk.ctx.error.throw`

sdk/wallet-sdk/src/wallet/namespace/amulet/preapproval.ts
```diff
@@ -116,12 +116,12 @@ export class PreapprovalNamespace {
         const { parties, inputUtxos, expiresAt } = args
         const preapprovalStatus = await this.fetchStatus(parties.receiver)
         const provider = parties?.provider ?? this.ctx.validatorParty
-        const synchronizerId =
-            args.synchronizerId ?? this.ctx.commonCtx.defaultSynchronizerId
+        const synchronizerId = args.synchronizerId
```
mateuszpiatkowski-da, line 119: nitpick: is this `const` needed? I think we can just write `args.synchronizerId`

docs/wallet-integration-guide/examples/snippets/setupTests.ts
```diff
@@ -95,12 +97,24 @@ async function beforeEachSetup() {
         asset: global.ASSET_CONFIG,
     })
 
+    // ========= Resolve the synchronizer parties are hosted on =========
+    const { connectedSynchronizers } = await sdk.ledger.connectedSynchronizers(
+        {}
```
mateuszpiatkowski-da, line 102: I've commented on this in the target branch's PR - I think that the argument can be optional, as `{}` doesn't say much to the reader

core/wallet-test-utils/src/otc-trade.ts
```diff
@@ -79,9 +79,18 @@ export class OTCTrade {
             PATH_TO_DAR_IN_LOCALNET
         )
 
+        // Resolve the global synchronizer explicitly: the SDK no longer
```
mateuszpiatkowski-da, line 82: nitpick: the comment refers to the prior status of SDK capabilities, but once this PR is merged this will no longer be valid as the changes will be overwritten. I suggest shortening this to:

```suggestion
        // Resolve the global synchronizer explicitly as the DAR upload cannot be autodetected when the
        // participant is connected to multiple synchronizers.
```

core/wallet-test-utils/src/otc-trade.ts
```diff
@@ -207,6 +216,20 @@ export class OTCTrade {
         }
     }
 
+    private async resolveGlobalSynchronizerId(): Promise<string> {
+        if (!this.sdk) throw new Error('SDK not initialized')
+
+        const { connectedSynchronizers } =
+            await this.sdk.ledger.connectedSynchronizers({})
```
mateuszpiatkowski-da, line 223: nitpick: same comment as in https://github.com/canton-network/wallet/pull/1740#discussion_r3488106567

docs/wallet-integration-guide/examples/scripts/utils/index.ts
```diff
@@ -39,6 +39,27 @@ export function resolveGlobalSynchronizerId(
...
 }
 
+/**
+ * Fetches connected synchronizers from the ledger API and returns the ID of the
+ * synchronizer aliased `'global'`.
+ *
+ * The wallet SDK no longer auto-selects a synchronizer, so client code (these
+ * examples) resolves it explicitly and passes it to SDK calls that require one.
+ */
+export async function getGlobalSynchronizerId(sdk: {
```
mateuszpiatkowski-da, line 49: One thing I don't understand - why can't `wallet-sdk` itself have the default behavior like described here to implement? I think it would be substantially easier for the end user to just let sdk do their own thing by default and have the possibility to optionally overwrite that behavior by passing either:
1. end value
2. callback incorporating appropriate business logic
3. value that would be used in sdk-fixed function

Is it possible for the global synchronizer id to change during the lifecycle of the sdk instance? Can the global synchronizer id even change? Intutively this should be a constant publically accessible, so in this case `synchronizerAlias` should always be equal to `global` which in turn should simplify both the logic and the user interface of sdk

mateuszpiatkowski-da (changes requested):
the code itself looks fine, but I don't understand why can't we simply make the default synchronizer id an optional parameter when instantiating sdk. This way the user would still have the ability to point to the right resource (as described in [this comment](https://github.com/canton-network/wallet/pull/1740#discussion_r3488120992)) while keeping the interface as simple as possible :)

PR #1743, opened 2026-05-13
chore: check before storing cache in CI

PHOL-DA:


scripts/src/start-localnet.ts
```diff
@@ -74,17 +73,24 @@ const env = { ...process.env, IMAGE_TAG: spliceVersion }
...
+        execFileSync(composeBase[0], [...composeBase.slice(1), 'pull'], {
+            stdio: 'inherit',
+            env,
+        })
+    } else {
+        execFileSync(composeBase[0], [...composeBase.slice(1), 'up', '-d'], {
+            stdio: 'inherit',
+            env,
+        })
+    }
```
pawelstepien-da: I think we could flatten this to single-level-depth if elses
```suggestion
if (command === 'pull') {
    execFileSync(composeBase[0], [...composeBase.slice(1), 'pull'], {
        stdio: 'inherit',
        env,
    })
} else if(command === 'start') {
    execFileSync(composeBase[0], [...composeBase.slice(1), 'up', '-d'], {
        stdio: 'inherit',
        env,
    })
```

pawelstepien-da (dismissed):

pawelstepien-da (approved):

PR #1769, opened 2026-05-13
chore: ensure aligned daml version script

PHOL-DA:
small script that ensure any daml.yaml is aligned with daml release version. Each variation of this adds ~ 700 mb to the dpm cache and ~ 2 min to ci.


mjuchli-da:
Close for now due to extended CI runtime. We can reopen if we really see the need.

PR #1773, opened 2026-05-15
fix(wallet-sdk): convert all urls

### LINKED ISSUE #1772: Wallet SDK bug during config
Config 1 will fail with a hard error while config 2 will work.

**Config 1:**
```
import { SDK, localNetStaticConfig } from '@canton-network/wallet-sdk';
async function main() {
    const auth = {
        method: 'self_signed',
        issuer: 'unsafe-auth',
        credentials: {
            clientId: 'ledger-api-user',
            clientSecret: 'unsafe',
            audience: 'https://canton.network.global',
            scope: '',
        },
    };
    const validatorBaseUrl = "http://127.0.0.1:2000";
    const sdk = await SDK.create({
        auth:auth,
        ledgerClientUrl: 'http://127.0.0.1:2975',
        token: {
            validatorUrl: `${validatorBaseUrl}/api/validator`,
            registries: [`${validatorBaseUrl}/api/validator/v0/scan-proxy`],
            auth: auth,
        },
        amulet: {
            validatorUrl: `${validatorBaseUrl}/api/validator`,
            scanApiUrl: 'https://scan.sv-1.global.canton.network.sync.global/api/scan',
            auth: auth,
            registryUrl: `${validatorBaseUrl}/api/validator/v0/scan-proxy`,
        },
        asset: {
            registries: [`http://localhost:2000/api/validator/v0/scan-proxy`],
            auth: auth,
        },
    });
}
main().catch(err => {
    console.error('Error initializing SDK:', err);
});
```

config 2:
```
import { SDK, localNetStaticConfig } from '@canton-network/wallet-sdk';
async function main() {
    const auth = {
        method: 'self_signed',
        issuer: 'unsafe-auth',
        credentials: {
            clientId: 'ledger-api-user',
            clientSecret: 'unsafe',
            audience: 'https://canton.network.global',
            scope: '',
        },
    };
    const validatorBaseUrl = "http://127.0.0.1:2000";
    const sdk = await SDK.create({
        auth:auth,
        ledgerClientUrl: 'http://127.0.0.1:2975',
        token: {
            validatorUrl: `${validatorBaseUrl}/api/validator`,
            registries: [`${validatorBaseUrl}/api/validator/v0/scan-proxy`],
            auth: auth,
        },
        amulet: {
            validatorUrl: `${validatorBaseUrl}/api/validator`,
            scanApiUrl: 'https://scan.sv-1.global.canton.network.sync.global/api/scan',
            auth: auth,
            registryUrl: `${validatorBaseUrl}/api/validator/v0/scan-proxy`,
        },
        asset: {
            registries: [new URL(`http://localhost:2000/api/validator/v0/scan-proxy`)],
            auth: auth,
        },
    });
}
main().catch(err => {
    console.error('Error initializing SDK:', err);
});
```

**only different being the asset.registries (one wrapped in url other as pure string).**

Config one fails with a hard error of invalid url type

mateuszpiatkowski-da:


PHOL-DA (changes requested):
looks good, except one small problem. These changes are not backwards compatible for anything that is exposed through the wallet sdk. This means we effectively can't put these in until a wallet SDK V2.

my suggested approach would be:

extend all the namespaces with `string | URL` where applicable (or make two variations one that take string and one that take URL and mark the string as @deprecated).

we make a under the hood conversion of string to URL.

alexmatson-da:
> extend all the namespaces with string | URL where applicable (or make two variations one that take string and one that take URL and mark the string as @deprecated).

My two cents would be to _not_ mark string variant as deprecated. The wallet-sdk is fairly flexible, and I can imagine someone might build a 3rd party app that reads SDK configuration values externally (a JSON file, environment variables, or even some API call).

In that case, it would pretty convenient to support plain string URLs so that the developer doesnt have to remap their external config into the SDK constructor and convert types

(We should still do URL validation, of course)

PHOL-DA (approved):

PR #1779, opened 2026-05-15
fix: transactions by id

rukmini-basu-da:
https://github.com/canton-network/wallet-gateway/issues/1778


PHOL-DA (approved):
LGTM, however i would suggest you update the migration docs and add it to atleast one of our test snippets or scripts so we have it proven working somewhere as well

PR #1814, opened 2026-05-19
feat: deliver CIP-103 push events through DappSyncProvider

### LINKED ISSUE #1815: DappSyncProvider drops all CIP-103 push events (txChanged, accountsChanged, statusChanged, connected)
## Summary

`DappSyncProvider` + `WindowTransport` (the postMessage path used by any CIP-103 browser-extension wallet via `ExtensionAdapter`) silently drops every push event. The dApp-side `provider.on('txChanged', listener)` contract that CIP-103 inherits from EIP-1193 is structurally non-functional on this path.

The Async path (`DappAsyncProvider` over HTTP+SSE) wires `eventSource.addEventListener(...)` correctly, and adapter-based providers like `LoopAdapter` work because they emit through their own in-process `EventEmitter`. The Sync/postMessage path is the only outlier.

## End-user symptom

`examples/ping` against any CIP-103-compliant browser extension wallet: user signs successfully, the transaction lands on the ledger, but the dApp UI ("Total transactions: 1") never updates because `sdk.onTxChanged(listener)` never fires.

## Root cause (verbatim from upstream code)

`core/rpc-transport/src/index.ts` lines 65-83 (`wallet-gateway/main`):

```ts
const listener = (event: MessageEvent) => {
    if (
        !isSpliceMessageEvent(event) ||
        event.data.type !== WalletEvent.SPLICE_WALLET_RESPONSE ||
        event.data.response.id !== message.request.id
    ) {
        return
    }
    window.removeEventListener('message', listener)
    // resolves pending promise
}
window.addEventListener('message', listener)
```

Two structural problems:

1. **Per-request, response-only**: the only `window.message` listener is installed inside `submit()`, filters for `SPLICE_WALLET_RESPONSE` with matching `id`, and removes itself after the first match.
2. **No always-on event listener**: between `submit()` calls, the dApp window has zero listeners for any `SPLICE_WALLET_*` traffic. Push events from the wallet arriving in that gap have nowhere to land.

## EIP-1193 precedent

CIP-103 names EIP-1193 as its design inspiration. EIP-1193 is deliberately silent on the inpage-to-wallet wire (`eip-1193.md:24`), but mandates that `provider.on(eventName, listener)` works per the Node.js `EventEmitter` API (`eip-1193.md:158-167`).

The downstream-implementation convention is uniform across browser-extension wallets (MetaMask, Rabby, Coinbase Wallet, Frame, Brave, OKX, Phantom's EVM mode): a **single permanent inpage listener** on `window.postMessage`, demuxing **JSON-RPC notifications (no `id`)** for push events vs **responses (with matching `id`)** for replies, on the same channel.

## The dapp-sdk already ships this pattern three other ways

| Wallet picker entry | Provider class | Event delivery | Works? |
|---|---|---|---|
| Wallet Gateway (localhost, dev) | `DappAsyncProvider` | `eventSource.addEventListener('txChanged', ...)` (`DappAsyncProvider.ts:51-105`) | ✅ |
| 5N Loop Wallet (Devnet) | `LoopSDKProvider` (in adapter) | In-process `this.emit('txChanged', ...)` (`loop-adapter.ts:406-433`) | ✅ |
| canton (injected) | `DappSyncProvider` | None — bug | ❌ |
| any CIP-103 browser-extension wallet | `DappSyncProvider` | None — bug | ❌ |

Fixing `DappSyncProvider` is doing for the postMessage path what `DappAsyncProvider` already does for SSE.

## CIP-103 spec gap

CIP-103 today defines the four push events (`txChanged`, `accountsChanged`, `statusChanged`, `connected`) as OpenRPC methods (pollable). The docs label them as "Events" in the Sync API but stop short of specifying a wire envelope for pushing them over postMessage. The HTTP/SSE path has a de-facto answer (SSE event-name). The postMessage path has no documented answer.

This issue treats the implementation gap as the primary deliverable. The spec ambiguity is a follow-up to surface to the CIP-103 maintainers (`canton-foundation/cips`) once the upstream design choice for the envelope is made — happy to drive that conversation when the time comes.

## Working reference implementation

Open as PR [`canton-network/wallet#1814`](https://github.com/canton-network/wallet/pull/1814) (against `main`) from `alleneubank:bb/dapp-sync-event-channel`. Minimal patch:

| File | What |
|---|---|
| `core/types/src/index.ts` | `WalletEvent.SPLICE_WALLET_EVENT` + matching `SpliceMessage` discriminated-union variant |
| `core/rpc-transport/src/index.ts` | Optional `RpcTransport.onEvent` + `WindowTransport.onEvent` with single shared `message` listener gated on `target` |
| `core/provider-dapp/src/DappSyncProvider.ts` | Constructor wires `transport.onEvent` → `AbstractProvider.emit(event, payload)`; adds `teardown()` |

Manually verified end-to-end against `examples/ping` with a CIP-103 browser extension: pre-patch UI never updates; post-patch UI renders `Total transactions: N` with all three lifecycle JSON payloads (pending → signed → executed).

## Design choice for upstream

The wire envelope is a bikeshed on top of the structural fix. Two viable options:

- **Option A** — extension envelope `{type: 'SPLICE_WALLET_EVENT', event, payload, target?}`. Symmetric with existing `SPLICE_WALLET_REQUEST` / `_RESPONSE` envelopes. Adopted by the working reference impl.
- **Option B** — JSON-RPC notification: `{type: 'SPLICE_WALLET_REQUEST', request: {jsonrpc: '2.0', method: '<eventName>', params: [...]}}` with no `id`. Reuses the existing request envelope; matches the EIP-1193 convention more directly. Trade-off: `WindowTransport.submit` needs a guard to treat `SPLICE_WALLET_REQUEST`-without-`id` as a notification rather than a request addressed to the dApp.

Either choice needs the same four pieces of plumbing (envelope type, `RpcTransport.onEvent`, `WindowTransport`'s always-on `installEventDispatcher`, and the `DappSyncProvider` constructor wiring). Maintainers pick the envelope; happy to author the upstream PR mirroring whichever shape lands.

## References

- EIP-1193 spec: <https://eips.ethereum.org/EIPS/eip-1193>
- CIP-103 spec: <https://github.com/canton-foundation/cips/blob/main/cip-0103/cip-0103.md>
- Working reference PR: <https://github.com/canton-network/wallet/pull/1814>
- Diagnosis doc on the PR branch: `docs/fork-only-dapp-sync-event-channel.md`
- Related upstream PR (JsCommands schema, unrelated scope): <https://github.com/canton-network/wallet/pull/1726>

alleneubank:
## Summary

CIP-103 requires wallets to deliver `txChanged` / `accountsChanged` / `statusChanged` / `connected` through `provider.on(event, listener)`. On the sync (postMessage) path used by `ExtensionAdapter`, no wire shape a wallet posts results in delivery: `WindowTransport` only installs per-request, response-only listeners, so every wallet-pushed event is silently dropped — tracked in #1815.

End-user symptom (reported in public Slack, repro'd against `examples/ping`):

> "I sign the transaction successfully, but the transaction never seems to go through."

The transaction *does* land on the ledger; the dApp's `sdk.onTxChanged(listener)` just never fires, so the UI never updates.

## Envelope: JSON-RPC 2.0 notification (revised)

An earlier revision of this PR introduced a new `SPLICE_WALLET_EVENT` message type and asked for feedback on the envelope shape. This revision settles on the shape proposed in #1815: a **JSON-RPC 2.0 notification (§4.1)** — the existing `SPLICE_WALLET_REQUEST` frame with no `id`, `method` = event name. Reusing the existing wire vocabulary means **no new message type and no `SpliceMessage` union change**; the `core/types` diff from the earlier revision is gone entirely.

Requests submitted by the transport always carry a uuid `id`, so an id-less request is unambiguously a wallet-originated notification — this also keeps the dApp's own outbound frames (posted to the same window) out of the event path.

## Changes (sync flow only; HTTP/SSE untouched)

| File | What |
| --- | --- |
| `core/rpc-transport/src/index.ts` | `WindowTransport.onNotification(handler)`: a **single always-on** window `message` listener (installed on first subscription, removed with the last unsubscribe) demuxing id-less `SPLICE_WALLET_REQUEST` frames to subscribers. Delivery is gated on the transport's `target` routing key — the announced provider's key from `requestAnnouncedProviders` — so multi-wallet pages don't cross-deliver events between providers. |
| `core/provider-dapp/src/DappSyncProvider.ts` | Constructor subscribes to transport notifications and forwards each `(method, params)` to `this.emit(method, params)`, making `provider.on(event, listener)` deliver as CIP-103 requires. |
| `core/provider-dapp/src/DappSyncProvider.push-events.test.ts` | Envelope semantics: the canonical id-less notification is delivered; frames in any other envelope are ignored (the demux cannot become a sink for arbitrary window traffic). |
| `core/provider-dapp/src/DappSyncProvider.notification-demux.test.ts` | Announced-provider path: targeted delivery, cross-wallet isolation (wrong-target and unstamped frames dropped by a targeted transport), and no echo of the dApp's own id-carrying requests. |

`RpcTransport` itself is unchanged — `onNotification` is a `WindowTransport` member and `DappSyncProvider` feature-detects it, so `HttpTransport` / `DappAsyncProvider` (which already deliver events via SSE) are unaffected.

## Test plan

- [x] Unit/integration: 12/12 in `core/provider-dapp` (7 pre-existing + 5 new), vitest browser mode (chromium). The new coverage is exactly what the earlier revision deferred pending envelope consensus.
- [x] End-to-end red/green at the dApp UI, against a production CIP-103 extension wallet (Send Connect) on a Canton localnet: a Playwright spec connects `examples/ping` through the **announced** provider, creates a Ping contract, and asserts the Post Events panel (rendered solely from `sdk.onTxChanged` subscriptions). With the stock dapp-sdk the wallet's notifications are observably posted into the window but the panel stays empty; with this patch the same spec passes and the panel renders the `txChanged` lifecycle (pending → signed → executed).
- [x] HTTP/SSE flow regression: no code change on `DappAsyncProvider` / `HttpTransport`.

## Links

- Closes #1815 (sync-path event delivery)
- CIP-0103 Provider API event contract (`provider.on`)

### DIFF
--- a/core/provider-dapp/src/DappSyncProvider.notification-demux.test.ts
+++ b/core/provider-dapp/src/DappSyncProvider.notification-demux.test.ts
@@ -1,0 +1,74 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from 'vitest'
+import { WalletEvent } from '@canton-network/core-types'
+import { WindowTransport } from '@canton-network/core-rpc-transport'
+import { DappSyncProvider } from './DappSyncProvider'
+
+// Companion to DappSyncProvider.push-events.test.ts: exercises the always-on
+// notification demux against the *announced-provider* construction path.
+//
+// Discovery hands the dApp a routing key, not a provider object
+// (requestAnnouncedProviders → AnnouncedProvider.target), and the SDK then
+// builds DappSyncProvider(new WindowTransport(window, { target })). In that
+// future every transport is targeted, so event delivery must honor the same
+// target gating as extension detection — otherwise a page with two wallets
+// cross-delivers one wallet's events to the other wallet's provider.
+describe('DappSyncProvider notification demux (announced-provider path)', () => {
+    const flushMessageQueue = () => new Promise((r) => setTimeout(r, 100))
+
+    const notification = (target?: string) => ({
+        type: WalletEvent.SPLICE_WALLET_REQUEST,
+        request: {
+            jsonrpc: '2.0',
+            method: 'txChanged',
+            params: { status: 'executed', commandId: 'cmd-demux-1' },
+        },
+        ...(target !== undefined ? { target } : {}),
+    })
+
+    it('delivers notifications stamped with the provider announced routing key', async () => {
+        const provider = new DappSyncProvider(
+            new WindowTransport(window, { target: 'wallet-a' })
+        )
+        const received: unknown[] = []
+        provider.on('txChanged', (payload: unknown) => received.push(payload))
+
+        window.postMessage(notification('wallet-a'), '*')
+        await flushMessageQueue()
+
+        expect(received).toHaveLength(1)
+    })
+
+    it('does not cross-deliver notifications addressed to another wallet', async () => {
+        const provider = new DappSyncProvider(
+            new WindowTransport(window, { target: 'wallet-a' })
+        )
+        const received: unknown[] = []
+        provider.on('txChanged', (payload: unknown) => received.push(payload))
+
+        window.postMessage(notification('wallet-b'), '*')
+        // An unstamped frame is also undeliverable to a targeted provider:
+        // with multiple announced wallets there is no way to attribute it.
+        window.postMessage(notification(), '*')
+        await flushMessageQueue()
+
+        expect(received).toHaveLength(0)
+    })
+
+    it('does not echo the dApp own id-carrying requests into the event path', async () => {
+        const provider = new DappSyncProvider()
+        const received: unknown[] = []
+        provider.on('ping', (payload: unknown) => received.push(payload))
+
+        // request() posts an id-carrying SPLICE_WALLET_REQUEST into the same
+        // window the demux listens on; it must be treated as an outbound call,
+        // not a wallet notification. No wallet responds in this harness, so
+        // the returned promise intentionally never settles.
+        void provider.request({ method: 'ping' } as never)
+        await flushMessageQueue()
+
+        expect(received).toHaveLength(0)
+    })
+})
--- a/core/provider-dapp/src/DappSyncProvider.push-events.test.ts
+++ b/core/provider-dapp/src/DappSyncProvider.push-events.test.ts
@@ -1,0 +1,73 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, expect, it } from 'vitest'
+import { WalletEvent } from '@canton-network/core-types'
+import { DappSyncProvider } from './DappSyncProvider'
+
+// Regression coverage for https://github.com/canton-network/wallet/issues/1815.
+//
+// The CIP-103 Provider API requires wallets to deliver txChanged /
+// accountsChanged / statusChanged / connected through
+// provider.on(event, listener). On the sync (postMessage) path the wire
+// shape for a wallet-pushed event is a JSON-RPC 2.0 notification — the
+// existing SPLICE_WALLET_REQUEST envelope with no `id`, method = event
+// name. WindowTransport demuxes those frames on an always-on listener and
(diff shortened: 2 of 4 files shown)


alexmatson-da (approved):
thanks!

PR #1821, opened 2026-05-20
fix: fix package.json with new repo

PHOL-DA:


mateuszpiatkowski-da (dismissed):

core/rpc-generator/src/components/client.ts
```diff
@@ -143,7 +143,7 @@ const hooks: IHooks = {
                 if (repositoryDirectory) {
                     updatedPkgObj.repository = {
                         type: 'git',
-                        url: 'git+https://github.com/canton-network/wallet-gateway.git',
+                        url: 'git+https://github.com/canton-network/wallet.git',
```
pawelstepien-da, line 146: This should work, but ideally we should also update here for the fallback and to keep it uniform [`repository`](https://github.com/canton-network/wallet-gateway/blob/146b9fa9d987a966ec78093957e4adb138ecca73/core/rpc-generator/templates/client/typescript/_package.json#L51)

PHOL-DA: done, had my renaming a bit to narrow :+1:

pawelstepien-da (approved):

PR #1872, opened 2026-05-22
feat: hide auth details from listNetwork and mint self-signed token server-side

### LINKED ISSUE #1802: Remove auth details from `listNetworks` response
Given that the endpoint is not protected, this is a potential security risk.

The `network` type comes with many properties such as `auth`, `adminAuth`. 
Instead of exposing the object directly to the user api, we should work with a dto which only comes with the properties effectively used by the User UI.

mjuchli-da:
Closes #1802

Since we want to hide sensitive information such as the `clientSecret` to be exposed via the API, we need to handle the self-signed login server-side. This PR therefore introduces a new endpoint `selfSignedAccessToken` to issue an access token.

In a follow-up, one could think of combining this call with the creation of a `Session` (currently the workflow is: `selfSignedAccessToken + addSession`) and return the session object to the frontend.


alexmatson-da (approved):
nice!

PR #1875, opened 2026-05-26
chore: fix readme versioning

PHOL-DA:


docs/dapp-building/wallet-gateway/deployment/index.md
```diff
@@ -91,7 +91,7 @@ The following config is incomplete, but highlights specific fields of note to co
 ```yaml
 kernel:
     # Set the publically accessible URL that users would use to connect to the deployed Wallet Gateway.
-    # Subpath routing is also supported as of v0.20.0
+    # Subpath routing is also supported as of v1.3.1
```
pawelstepien-da: Just to be sure - we don't want to put `<VERSION>` here?

pawelstepien-da (dismissed):

docs/dapp-building/wallet-gateway/deployment/index.md
PHOL-DA: just updated it to not mention any specific version

pawelstepien-da (approved):

mjuchli-da (approved):

PR #1877, opened 2026-05-26
test: wallet gateway front-end

pawelstepien-da:
partially implements https://github.com/canton-network/wallet/issues/1768


wallet-gateway/remote/vitest.config.ts
```diff
@@ -27,9 +30,27 @@ export default defineConfig({
...
+                    name: 'browser',
+                    include: ['src/web/frontend/**/*.test.ts'],
+                    setupFiles: ['./vitest.setup.browser.ts'],
+                    browser: {
+                        enabled: true,
+                        provider: playwright({
+                            trace: 'off',
+                            screenshot: 'off',
+                            video: 'off',
+                        }),
```
alexmatson-da, line 48: just curious to see if i got it right -- basically this is saying we're using playwright to run the unit tests through vitest?

wallet-gateway/remote/src/web/frontend/test-helpers.ts
alexmatson-da, line 1: I wonder if we should structure things within some kind `__tests__` directory? That way we can exclude our test files from being published in the NPM packages. Not a big deal right now, just something to think about 🤔

alexmatson-da (approved):
nice, thorough!

wallet-gateway/remote/vitest.config.ts
pawelstepien-da: Exactly. Tests are running in a headless browser with that settings. Traditional alternative is happy-dom or js-dom which are more like browser apis polyfills but still running in node.
This article inspired me to use that approach and so far it works well
https://www.epicweb.dev/why-i-won-t-use-jsdom

wallet-gateway/remote/src/web/frontend/test-helpers.ts
pawelstepien-da: A very valid point. I created an issue for that under tech debt https://github.com/canton-network/wallet/issues/1899

PR #1882, opened 2026-05-26
fix: normalize urls for config

PHOL-DA:
**Problem**
Some times people expect default behavior when inserting string urls in config, for instance they might insert localhost:8080 instead of http://localhost:8080 or they might insert my-domain.com instead of http://my-domain.com.

**Solution**
This pr addresses default behavior, defined as:
- if using http and no port, then use port 80
- if using https and no port, then use port 443
- if no transport is provided use http


mjuchli-da:
Closing for now due to complexity. We can reopen once we see the need / receive feedback.

PR #1889, opened 2026-05-27
chore: upgrade package versions

PHOL-DA:


alexmatson-da:
has file conflicts, we should probs close and reopen a fresh PR for dep bumps

PR #1890, opened 2026-05-27
feat: update wallet-sdk types, introduce sdk plugin in portfolio

mateuszpiatkowski-da:


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "examples/portfolio/wallet-sdk-utility"]
```
fayi-da: I wonder if it's simpler to just paste the code(just the file we need) in here rather than using submodules. we can easily delete it later

PR #1896, opened 2026-05-27
test: tx parser test

### LINKED ISSUE #1854: core-tx-parser unit tests

rukmini-basu-da:
Closes https://github.com/canton-network/wallet/issues/1854

Moves the tx parser tests from the token-standard-service module to the tx-parser module (seems like these got shifted when we moved these out of the core-ledger-client.


alexmatson-da (approved):
thx!

pawelstepien-da (approved):

PR #1898, opened 2026-05-27
feat: add support for additional postgres configuration fields

da-tanabe:
Expose the ability to use more fields than just those that are part of the schema. The main motivation for this is to enable SSL-based fields, but there may be other fields from the underlying postgres connection that people may also want to set.

I kept the original schema definitions in place so that we at least get validation for known fields; things like ports still need to be numbers, for example.

Signed-off-by: Davin K. Tanabe <davin.tanabe@digitalasset.com>


alexmatson-da (dismissed):

mjuchli-da:
# Tutorial: Wallet Gateway + Postgres over TLS (Docker)

## 0) Prereqs

- docker running
- openssl available

## 1) Create TLS files for Postgres (self-signed)

```shell
mkdir -p .dev/postgres-tls
cd .dev/postgres-tls

# server key + cert (CN=localhost)
openssl req -x509 -newkey rsa:2048 -nodes \
  -keyout server.key -out server.crt -days 365 \
  -subj "/CN=localhost"

# Postgres requires strict key perms
chmod 600 server.key

# pg_hba: allow local socket (for init), enforce TLS for TCP
cat > pg_hba.conf <<'EOF'

# TYPE  DATABASE  USER  ADDRESS       METHOD
local   all       all                 scram-sha-256
hostssl all       all   0.0.0.0/0     scram-sha-256
hostssl all       all   ::/0          scram-sha-256
EOF

cd ../..
```

## 2) Run Postgres with TLS enabled (Docker command)

```shell
docker rm -f local-postgres 2>/dev/null || true
```

```shell
docker run --name local-postgres \
  -e POSTGRES_USER=postgres \
  -e POSTGRES_PASSWORD=postgres \
  -e POSTGRES_DB=app_db \
  -e POSTGRES_INITDB_ARGS="--auth-host=scram-sha-256 --auth-local=scram-sha-256" \
  -p 5432:5432 \
  -v "$PWD/.dev/postgres-tls/server.crt:/var/lib/postgresql/server.crt:ro" \
  -v "$PWD/.dev/postgres-tls/server.key:/var/lib/postgresql/server.key:ro" \
  -v "$PWD/.dev/postgres-tls/pg_hba.conf:/var/lib/postgresql/pg_hba.conf:ro" \
  -d postgres:16 \
  -c ssl=on \
  -c ssl_cert_file=/var/lib/postgresql/server.crt \
  -c ssl_key_file=/var/lib/postgresql/server.key \
  -c hba_file=/var/lib/postgresql/pg_hba.conf
```

## 3) Verify TLS works

Verify via TCP + TLS (inside container)
```shell
docker exec -e PGPASSWORD=postgres local-postgres \
  psql "host=127.0.0.1 port=5432 user=postgres dbname=app_db sslmode=require" \
  -c "select current_setting('ssl') as ssl_on;"
```
Expected: ssl_on = on.

## 4) Configure Wallet Gateway store + signingStore to use TLS
In your wallet-gateway config file, set both connections like this (quick local mode, no cert verification):
```json
{
  "store": {
    "connection": {
      "type": "postgres",
      "host": "localhost",
      "port": 5432,
      "user": "postgres",
      "password": "postgres",
      "database": "app_db",
      "ssl": { "rejectUnauthorized": false }
    }
  },
  "signingStore": {
    "connection": {
      "type": "postgres",
      "host": "localhost",
      "port": 5432,
      "user": "postgres",
      "password": "postgres",
      "database": "app_signing_db",
      "ssl": { "rejectUnauthorized": false }
    }
  }
}
```

## Optional: verify the server cert locally (with ssl.ca)
If you generated a CA (the “verified” variant), copy the CA PEM into a JSON string (no literal newlines):

```shell
perl -0777 -pe 's/\n/\\n/g' .dev/postgres-tls/server.crt | pbcopy
```

Then:
```json
"ssl": {
  "rejectUnauthorized": true,
  "ca": "<PASTE HERE>"
}
```

mjuchli-da (approved):

fayi-da (approved):

Issue #1790, opened 2026-05-17
Implement dashboard page content


fayi-da:
Figma: https://www.figma.com/design/VzQePESEI6WLMD8L29lrSI/Splice-Portfolio-App?node-id=320-22933&t=0zZYcGaukZWTrDGS-4

- [ ] Show action required items
- [x] Link to offers
- [x] Show assets across all wallets
- [x] Add and link asset transfer button

PR #1908, opened 2026-05-28
feat: implement asset transfer for new portfolio dapp

fayi-da:
Part of #1790

https://github.com/user-attachments/assets/7e4a49b6-43cd-4e45-af7f-5e552aff4a48

### DIFF
--- a/examples/portfolio/src/components/next/transfer-form.tsx
+++ b/examples/portfolio/src/components/next/transfer-form.tsx
@@ -1,0 +1,514 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useMemo, useState } from 'react'
+import CloseIcon from '@mui/icons-material/Close'
+import {
+    Alert,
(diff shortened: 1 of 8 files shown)


examples/portfolio/src/components/next/transfer-form.tsx
```diff
@@ -0,0 +1,514 @@
...
+                            fullWidth
+                            disabled={
+                                !canSubmit ||
+                                isSubmitting ||
+                                disabled ||
+                                !primaryParty
+                            }
+                            sx={{ mt: 0.5, minHeight: 48 }}
+                        >
+                            {disabled ? (
```
pawelstepien-da, line 355: I'm not sure if that matters here, but maybe `isSubmitting === true` also should render the spinner?

pawelstepien-da (approved):

examples/portfolio/src/components/next/transfer-form.tsx
fayi-da: @pawelstepien-da it does implictily

```ts
    const disabled = createTransferMutation.isPending
```

the mutation is pending because submission is happening :)

PR #1910, opened 2026-05-29
fix: transform mixed esm modules

fayi-da:


alexmatson-da (approved):
if you havent yet, maybe worth testing the bundle by running some local web server (like `python -m http.server`) from the dist/ directory. If that works, then i'm reasonably confident the image should work

fayi-da:
> if you havent yet, maybe worth testing the bundle by running some local web server (like `python -m http.server`) from the dist/ directory. If that works, then i'm reasonably confident the image should work

@alexmatson-da I actually went further and built a docker image using a similar Dockerfile in the infra repo. Ran it on port 3333, added 3333 for cors to WG and tested a simple flow. All worked well

Issue #1915, opened 2026-05-29
Remove old code,  document new code


fayi-da:
- [x] Remove everything in /_legacy
- [x] Remove any unused utilities
- [ ] Document Code
  - [ ] Add readme with overview and deployment instructions
  - [ ] Document hooks/utils where necessary

PR #1916, opened 2026-05-29
chore: remove unused /old and all old components in portfolio

fayi-da:
These have been unused for a long time. See ya later :wave: 

Part of #1915


alexmatson-da (approved):
👏

PR #1927, opened 2026-06-01
fix: allow empty strings for optional values in config and remove scanProxyUrl

fayi-da:


alexmatson-da (approved):

examples/portfolio/src/config/portfolio-config.ts
alexmatson-da: I realized we may want to consider combining these...

alexmatson-da (changes requested):

examples/portfolio/src/config/portfolio-config.ts
fayi-da: if you know for sure we dont need both then yes. Was planning to look into it when switching those services for wallet sdk.

alexmatson-da (approved):

PR #1928, opened 2026-06-01
feat: filter out holdings by instrument

### LINKED ISSUE #1925: Token-standard-service: Add instrumentId when selecting utxos for transfer
https://github.com/canton-network/wallet/blob/954de7e4e657f018c854ac5b0b00dbae9066e814/core/token-standard-service/src/token-standard-service.ts#L104

When creating a transfer, we pass in the inputUtxos as a choiceArgument. This should be filtered by instrumentId + admin and passed in the createTransfer command rather than listing all holdings.

rukmini-basu-da:
Fixes https://github.com/canton-network/wallet/issues/1925

Currently, when creating a transfer, all the unlocked utxos are selected (using the acs) for a senderParty. The original workaround was to listHoldings and filter by the instrument, then supply that list of utxos to createTransfer. We are adding this filtering logic to the getINputCids function if the instrumentId/instrumentAdmin are defined (these are passed in the createTransfer function)

Additionally, this removes parseFloat and replaces with Decimal

### DIFF
--- a/core/amulet-service/package.json
+++ b/core/amulet-service/package.json
@@ -44,6 +44,7 @@
         "@canton-network/core-wallet-auth": "workspace:^",
         "bignumber.js": "^10.0.2",
         "dayjs": "^1.11.19",
+        "decimal.js": "^10.6.0",
         "openapi-fetch": "^0.17.0",
         "pino": "^10.3.1"
     },
--- a/core/token-standard-service/package.json
+++ b/core/token-standard-service/package.json
@@ -44,7 +44,6 @@
         "@canton-network/core-tx-parser": "workspace:^",
         "@canton-network/core-types": "workspace:^",
         "@canton-network/core-wallet-auth": "workspace:^",
-        "bignumber.js": "^10.0.2",
         "dayjs": "^1.11.19",
         "decimal.js": "^10.6.0",
         "openapi-fetch": "^0.17.0",
--- a/core/token-standard-service/package.json
+++ b/core/token-standard-service/package.json
@@ -46,6 +46,7 @@
         "@canton-network/core-wallet-auth": "workspace:^",
         "bignumber.js": "^10.0.2",
         "dayjs": "^1.11.19",
+        "decimal.js": "^10.6.0",
         "openapi-fetch": "^0.17.0",
         "pino": "^10.3.1",
         "typescript-lru-cache": "^2.0.0"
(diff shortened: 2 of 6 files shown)


core/token-standard-service/src/token-standard-service.ts
```diff
@@ -140,24 +143,46 @@ export class CoreService {
...
+
+        if (options.amount) {
             return CoreService.getInputHoldingsCidsForAmount(
-                amount,
-                unlockedSenderHoldings
+                options.amount,
+                unlockedHoldingsForInstrument
             )
         } else {
             return unlockedSenderHoldings.map((h) => h.contractId)
```
fayi-da: Curious why the else branch returns `unlockedSenderHoldings`, not `unlockedHoldingsForInstrument` I would have thought since we cant filter by amount we just return the unfiltered `unlockedHoldingsForInstrument`

core/token-standard-service/package.json
```diff
@@ -46,6 +46,7 @@
         "@canton-network/core-wallet-auth": "workspace:^",
         "bignumber.js": "^10.0.2",
         "dayjs": "^1.11.19",
+        "decimal.js": "^10.6.0",
```
fayi-da, line 48: We already have biugnumber just 2 lines above. We don't need 2 number libs surely

core/token-standard-service/src/token-standard-service.ts
rukmini-basu-da: good catch, thanks fixed it!

core/token-standard-service/package.json
rukmini-basu-da: Updated, looks like bignumber wasn't being used

fayi-da: Quick search shows we use both across our repo. I created a tech debt for it https://github.com/canton-network/wallet/issues/1930

core/amulet-service/package.json
```diff
@@ -44,6 +44,7 @@
         "@canton-network/core-wallet-auth": "workspace:^",
         "bignumber.js": "^10.0.2",
```
fayi-da, line 45: I see it here too lol. I've created a tech debt for this https://github.com/canton-network/wallet/issues/1930

fayi-da (approved):
Thanks

PR #1938, opened 2026-06-03
fix: asyncapi streamupdates request body is invalid

### LINKED ISSUE #1939: bug: async api request body is invalid
While working the token-standard-service unit tests, I noticed that the streamUpdates in the wallet-sdk/async client was collecting the following event: 

```
    0: {
      "code": "INVALID_ARGUMENT",
      "cause": "The submitted request has invalid arguments: Both update_format and filter are set. Please use either backwards compatible arguments (filter and verbose) or update_format, but not both.",
      "correlationId": null,
      "traceId": "d14a5fa3d4813bea829bc60ca486f37a",
      "context": {
        "participant": "app-user",
        "definite_answer": "false",
        "tid": "d14a5fa3d4813bea829bc60ca486f37a",
        "category": "8"
      },
      "resources": [],
      "errorCategory": 8,
      "grpcCodeValue": 3,
      "retryInfo": null,
      "definiteAnswer": null
    }
```

This wasn't caught in the integration tests because we weren't asserting the anything about the event content. However, when adding 
```

if (
    updateEvents.length === 0 ||
    updateEvents.filter((event) => event.update) === undefined
) {
    logger.error(
        'No command completion events received, something went wrong with the subscription'
    )
    updatesController.abort()
    process.exit(1)
}
```

This causes the test to fail on main.

rukmini-basu-da:
Fixes https://github.com/canton-network/wallet/issues/1939

The `streamUpdates` function in the asyncapi client throws the following error against canton versions 3.5.x:

```
{"code":"INVALID_ARGUMENT","cause":"The submitted request has invalid arguments: Both update_format and filter are set. Please use either backwards compatible arguments (filter and verbose) or update_format, but not both.","correlationId":null,"traceId":"789fe90b612f00ece03bc3e64f0d53fc","context":{"participant":"app-user","definite_answer":"false","tid":"789fe90b612f00ece03bc3e64f0d53fc","category":"8"},"resources":[],"errorCategory":8,"grpcCodeValue":3,"retryInfo":null,"definiteAnswer":null}
```

The test was only testing the length of the updates but not the content of the response, so this was collecting errors as an event, this fixes the problem by changing the request format.

### DIFF
--- a/core/asyncapi-client/src/websocket-client.ts
+++ b/core/asyncapi-client/src/websocket-client.ts
@@ -30,6 +30,8 @@ export type CompletionStreamRequest =
 export type GetUpdatesRequest = LedgerCommonSchemas['GetUpdatesRequest']
 
 export type JsGetUpdatesResponse = LedgerCommonSchemas['JsGetUpdatesResponse']
+export type CompletionsResponse =
+    LedgerCommonSchemas['CompletionStreamResponse']
 
 type UpdateSubscriptionOptions = {
     beginExclusive: number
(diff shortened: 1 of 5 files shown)


alexmatson-da (approved):
Nice!

PR #1940, opened 2026-06-04
docs: splice-portfolio deployment

### LINKED ISSUE #1932: Portfolio deployment docs

alexmatson-da:


docs/dapp-building/examples/portfolio/index.md
```diff
@@ -0,0 +1,66 @@
...
+    registries:
+        - url: 'https://registry.example.com' # @schema required
+          name: '' # (optional)
+          partyId: '' # (optional)
+```
+
+Then
+
+```sh
+helm upgrade --install portfolio ghcr.io/digital-asset/splice-portfolio/helm/splice-portfolio:1.5.0 -f values.yaml
```
fayi-da, line 65: Do you have a cluster where you tested this helm deployment? Just curious

docs/dapp-building/examples/portfolio/index.md
```diff
@@ -0,0 +1,66 @@
...
+```
+
+## Deployment
+
+### Docker
+
+To start the UI with Docker, create a `config.json` in your current directory and set your validatorUrl. Then, run the container:
+
+```sh
+docker run --rm \
```
fayi-da, line 35: I use `--rm` locally to avoid an extra deletion step when testing but not sure we want that by default in the docs

docs/dapp-building/examples/portfolio/index.md
```diff
@@ -0,0 +1,66 @@
...
+    -p 3333:80 \
+    -v "config.json:/usr/share/nginx/html:ro" \
```
fayi-da, line 37:
```suggestion
    -v "path/to/config.json:/usr/share/nginx/html:ro" \
```

docs/dapp-building/examples/portfolio/index.md
alexmatson-da: I imagine someone copy/pasting this directly from the docs to just try it out temporarily, so it seems like a nicety. In a more production-like case I imagine people would use a docker-compose or something...

maybe it'd be a good idea to also provide an example docker-compose containing everything:

- a localnet
- a wallet gateway
- splice portfolio

docs/dapp-building/examples/portfolio/index.md
alexmatson-da: just tested out with `k3d` locally, turns out i need a `oci://` protocol prefix for the chart but otherwise it works. also the chart had a bug 🤦 but I fixed that

docs/dapp-building/examples/portfolio/index.md
alexmatson-da: same as before, was assuming someone following along with the docs above it ("create a config.json in the current directory")

docs/dapp-building/examples/portfolio/index.md
fayi-da: nice :saluting_face: . You need to show me that k3s setup someday

docs/dapp-building/examples/portfolio/index.md
fayi-da: > In a more production-like case I imagine people would use a docker-compose or something...

Fair. 

> maybe it'd be a good idea to also provide an example docker-compose containing everything:

Good shout. tech debt ticket? Assign it to me

fayi-da (approved):
thanks!

docs/dapp-building/examples/portfolio/index.md
alexmatson-da: added #1941

PR #1950, opened 2026-06-08
fix: get signing key by pub key user check

mjuchli-da:


pawelstepien-da (approved):

pawelstepien-da (changes requested):
Changes look good, but it exposed an issue in wallet-kernel signing driver - it uses signing store without auth and therefore `assertConnected` fails when driver calls `getSigningKeyByPublicKey` during wallet creation.

core/signing-blockdaemon/src/index.ts
```diff
@@ -67,7 +67,7 @@ export default class BlockdaemonSigningDriver implements SigningDriverInterface
                         ...(params.internalTxId !== undefined && {
                             internalTxId: params.internalTxId,
                         }),
-                        userIdentifier: userId,
+                        userIdentifier: authContext?.email,
```
mjuchli-da, line 70: userId used to be the email as far as i can see.

please double-check @pawelstepien-da

pawelstepien-da: Yes we definitely used value from email here. In BD signing driver controller the argument was called userId, it was typed as `userId: AuthContext['userId']`, but it was actually supposed to be email, so it was quite confusing and is so much clearer now.

PR #1952, opened 2026-06-08
feat: service account support

### LINKED ISSUE #1866: Implement Service Account Support for Automations in Wallet Gateway
## Objective

Implement the minimum necessary service account functionality in the Wallet Gateway to support straight-through, machine-to-machine automation for clients, particularly for high-volume tasks such as token minting and burning. This is particularly required for automations that are not written in TypeScript and therefore cannot rely on a combination of `wallet-sdk` and signing driver functionality.

## Deadline & Performance Target

Critical Cut-off Date: The feature should be complete, released and production ready by June 24.

Performance: The implementation must support an average of at least one Transaction Per Second (1 TPS) through the Wallet Gateway, with ability to handle peak traffic during major market events.

## Implementation Details

**Authorization Detection:**
The Wallet Gateway must detect automation requests by checking the authorization method used.
Specifically, detect the use of a client ID/secret (client credential Auth method), which signifies a machine-to-machine channel and triggers the service account workflow.

**Straight-Through Signing and Submission:**
Upon detecting an automation request, the Wallet Gateway must perform straight-through signing without requiring a UI view or manual screen.
The Wallet Gateway must be responsible for automatically submitting the signed transaction directly to the validator/ledger API.

**Access Control (Configuration):**
[OPTIONAL] Add a configuration setting to the Wallet Gateway that lists the users who are allowed to be accessed by service accounts.

**Policy Engine (Non-Requirement):**
DO NOT build a policy engine within the Wallet Gateway.
The service account will inherit read/write permissions based on the user's ledger permissions. Policy enforcement remains the responsibility of the external signing provider (e.g., Fireblocks).

**Asynchronous Signing Handling:**
Implement an internal background job within the Wallet Gateway.
This job is necessary to poll external signing providers for transaction status updates, which accounts for potential long-running manual approvals that may take hours or days.
The orchestrator service (background processes on the client end) will monitor the transaction status updates using the existing DApp API event stream.

**API Integration:**
Utilize the existing DApp API endpoints for this functionality.
No new DApp API endpoints or changes to the DApp API standard should be introduced for this initial iteration.

## Side effect

- Needs external audit
- Reshuffle internal ressources

mjuchli-da:
Closes #1866


wallet-gateway/remote/src/signing/service-account-session.ts
```diff
@@ -0,0 +1,66 @@
...
+ * Service-account networks can run without a pre-existing session by minting
+ * an access token. Interactive networks still require a valid stored session.
+ */
+export async function resolveAutomationRunContext(
+    bootstrapStore: Store & AuthAware<Store>,
+    userId: string,
+    networkId: string,
+    logger: Logger
+): Promise<AutomationRunContext | undefined> {
+    const network = await bootstrapStore.getNetwork(networkId)
```
pawelstepien-da, line 33: Doesn't that fail because of `assertConnected` in `getNetwork`, when using store without auth context?

core/wallet-store-inmemory/src/store-internal.ts
```diff
@@ -432,6 +432,14 @@ export class StoreInternal implements Store, AuthAware<StoreInternal> {
         return storage.transactions.get(transactionId)
     }
 
+    async listAllPendingTransactions(): Promise<Array<Transaction>> {
+        this.assertConnected()
```
pawelstepien-da: SQL store counterpart doesn't have `assertConnected`. Should we have it in both/neither?

wallet-gateway/remote/src/signing/signing-worker.ts
```diff
@@ -0,0 +1,280 @@
...
+
+        const transactionService = new TransactionService(
+            store,
+            transactionLogger,
+            this.options.signingDrivers,
+            notifier
+        )
+
+        try {
+            const result = await transactionService.signAndExecute(
```
pawelstepien-da, line 202: I wonder how this works with our regular signing and executing flow in UI.
I'm thinking of such flow:
-User creates a TX with signing provider fireblocks
-WG shows a toast that says complete tx in fireblocks and click approve.
-User approves in fireblocks
-Worker tick does `singAndExecute`
-User clicks approve (but tx is already executed)

FE doesn't know about status change, so it calls `userApi.sign`, but it's already executed and I'm worried that unexpected effects could happen, like making store tx downgrade it's status from executed to signed (because query to fireblocks api will say that transaction status is signed), or trying to execute already executed tx and getting ledger api error.

I think we will need to add some safeguards in user api `sign` and `execute` that check tx status is actually 'pending' for `sign` and 'signed' for `execute` before proceeding. 

In later steps maybe we should rely fully on the worker to finish transactions with external signing providers, instead of requiring user to click approve second time. Something to discuss.

core/wallet-store-inmemory/src/store-internal.ts
alexmatson-da: yes, removed from both (the SigningWorker needs to call this before the scoped store is created)

wallet-gateway/remote/src/signing/service-account-session.ts
alexmatson-da: yes, good catch... I saw two options:

1. Create an authenticated store with a dummy context (empty access token)
2. Remove the assertConnected

I opted for 2 since 1 felt hacky, and I dont think the security properties change significantly.

(BTW had to do the same with `getIdp` method store too)

wallet-gateway/remote/src/signing/signing-worker.ts
alexmatson-da: added guards in `sign` and `execute` against transactions that have already been executed

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -245,6 +249,11 @@ export class TransactionService {
         if (!existingTx) {
             throw new Error(`Transaction not found with id: ${transactionId}`)
         }
+
+        if (existingTx.status === 'executed') {
```
pawelstepien-da: Should we also check for status `signed`?

alexmatson-da: switched to `!== 'pending'` which should also include `failed` (I think we also want that)

pawelstepien-da (dismissed):

docs/dapp-building/wallet-gateway/configuration/index.md
```diff
@@ -569,6 +569,26 @@ Store `adminAuth` secrets via `clientSecretEnv` and Kubernetes secrets (Helm `oa
 
 For participant-only signing with no external custody, leave the Helm chart `signing: {}` block empty — see [Deployment](../deployment/index.md#signing-chart-values-signing-).
 
+## Service account automation
+
+When a network uses **`client_credentials`** for `auth` (machine-to-machine), the Wallet Gateway treats DApp API `prepareExecute` calls as **service account automation**: it prepares, signs, and submits transactions without opening the approval UI. Clients should still use the existing DApp API (`prepareExecute`, `txChanged` events, `addSession`) — no new endpoints are required.
```
mjuchli-da: Outdated, it's `ApiKey` auth now.

docs/dapp-building/wallet-gateway/usage/index.md
```diff
@@ -102,6 +102,12 @@ Use the same host as `kernel.publicUrl` and the default `server.dappPath` (`/api
 - Calls `listNetworks()` / `listIdps()`, then `createWallet()` with desired network and signing provider.
 - Uses `listWallets()`, `sign()`, `execute()`, etc. as needed for your use case.
 
+**5. Service account automation (M2M)**
```
mjuchli-da: Update as well

docs/dapp-building/wallet-gateway/index.md
```diff
@@ -30,6 +30,7 @@ The Wallet Gateway enables transparent interaction between a dApp, Validator Nod
 
 - [Getting Started](getting-started/index.md)
 - [Configuration](configuration/index.md)
+- [Automations](automations/index.md) — service accounts and M2M transaction submission
```
mjuchli-da: remove m2m reference

examples/service-account/main.py
mjuchli-da, line 1: This needs to be accompanied with an extensive Readme.md file that explains the purpose of this example and what is does. It also needs to mention that we've purposely written it in python to demonstrate that the service-account feature (e.g. use of api keys) is language agnostic.

wallet-gateway/remote/src/dapp-api/controller.ts
```diff
@@ -211,6 +217,33 @@ export const dappController = (
...
+            if (actAs.length === 0) {
+                if (!primaryWallet) {
+                    throw new Error(
+                        'No primary wallet found. Create or sync a wallet and set it as primary before prepareExecute.'
+                    )
+                }
+                actAs = [primaryWallet.partyId]
+            }
+
+            for (const party of actAs) {
```
mjuchli-da: Extract to helper/assertion method

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -110,6 +115,10 @@ export class TransactionService {
         ledgerClient?: LedgerClient,
         network?: Network
     ): Promise<ExecuteResult> {
+        if (transaction.status === 'executed') {
+            throw new Error('Cannot execute an already executed transaction')
```
mjuchli-da:
```suggestion
            throw new Error('Transaction is already executed.')
```

docs/dapp-building/wallet-gateway/automations/index.md
```diff
@@ -0,0 +1,377 @@
+# Service account automations
```
mjuchli-da, line 1: this needs to be updated with the changes made from using the client credentials auth as the automation, to using the api key generated by the user and the make use of the serviceAccountAuth (client_credentials).

docs/dapp-building/wallet-gateway/configuration/index.md
mjuchli-da: implicitly client_credentials is being use via network.serviceAccountAuth

pawelstepien-da (approved):

PR #1953, opened 2026-06-08
test: token standard service test

### LINKED ISSUE #1839: core-token-standard-service unit tests

rukmini-basu-da:
Closes https://github.com/canton-network/wallet/issues/1839

### DIFF
--- a/core/token-standard-service/src/token-standard-service.test.ts
+++ b/core/token-standard-service/src/token-standard-service.test.ts
@@ -1213,6 +1213,37 @@ describe('Token standard service', () => {
             'http://registry1.com',
             'http://registry2.com',
         ])
-        expect(result).toHaveLength(2)
+        expect(result).toHaveLength(3)
+    })
+
+    it('blah', async () => {
+        const { service, provider } = makeService()
+        provider.request
+            .mockResolvedValueOnce({ participantPrunedUpToInclusive: 5 })
+            .mockResolvedValueOnce({ offset: 100 })
+            .mockResolvedValueOnce([])
+
+        vi.spyOn(service.core, 'toPrettyTransactions').mockResolvedValue({
+            nextOffset: 100,
+            transactions: [],
+        })
+        await service.listHoldingTransactions(senderParty)
+        const resources = provider.request.mock.calls.map(
+            (c: any) => c[0].params.resource
+        )
+        expect(resources).toContain('/v2/state/latest-pruned-offsets')
+    })
+
+    it('transaction by id', async () => {
+        const { service, provider } = makeService()
+        provider.request.mockResolvedValue({ transaction: {} })
+        vi.spyOn(service.core, 'toPrettyTransaction').mockResolvedValue({
+            id: 'tx-1',
+        } as any)
+
+        await service.getTransactionById('update-abc', senderParty)
+        const [call] = provider.request.mock.calls
+        expect(call[0].params.resource).toBe('/v2/updates/transaction-by-id')
+        expect(call[0].params.requestMethod).toBe('post')
     })
 })
--- a/core/token-standard-service/src/token-standard-service.test.ts
+++ b/core/token-standard-service/src/token-standard-service.test.ts
@@ -108,107 +108,6 @@ const makeHolding = (
 const instrumentAdmin =
     'DSO::1220c69732dd5f3b434c283f61cbc29d3bb492c50c56e306b436c3e1741cbc7be53e'
 const instrumentId = 'Amulet'
-function createHolding(
-    cId: string,
-    amount: string,
-    admin: string,
-    instrumentId: string,
-    lock?: { expiresAt?: string | null } | null
-) {
-    return {
-        contractId: cId,
-        activeContract: {
-            createdEvent: {
-                offset: 49,
-                nodeId: 3,
-                contractId: cId,
-                templateId:
-                    'a31be0483f3175647053f28965a4e6d97e3dbc433ea2338be303fae69bbcff6a:Splice.Amulet:Amulet',
-                contractKey: null,
-                contractKeyHash: '',
-                createArgument: {
-                    dso: 'DSO::1220c69732dd5f3b434c283f61cbc29d3bb492c50c56e306b436c3e1741cbc7be53e',
-                    owner: 'v1-01-alice::12206eee60f64d90be3f823007d1321dc6acc5f4f2c57d3dd6ac1f66148753bb65c5',
-                    amount: {
-                        initialAmount: amount,
-                        createdAt: {
-                            number: '1',
-                        },
-                        ratePerRound: {
-                            rate: '0.0038051800',
-                        },
-                    },
-                },
-                createdEventBlob:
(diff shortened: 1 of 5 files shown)


core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -1,47 +1,412 @@
...
+                )
+            expect(result.inputHoldingCids).toEqual(['cid1', 'cid2'])
+            expect(result.requestedAt).toBe(ts)
+        })
+    })
+
+    describe('createAllocationInstructionFromContext', () => {
+        const instrumentAdmin =
+            'DSO::1220c69732dd5f3b434c283f61cbc29d3bb492c50c56e306b436c3e1741cbc7be53e'
+        it('calls getInputHoldingCids and embeds resulting cids', async () => {
```
pawelstepien-da: Update description to match what is tested.

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -1,47 +1,412 @@
...
+        const { service } = makeService()
+        vi.spyOn(service.core, 'listContractsByInterface').mockResolvedValue([
+            lockedHolding,
+            unlockedHolding,
+        ])
+
+        const result = await service.core.getInputHoldingsCids({
+            sender: senderParty,
+        })
+        expect(result).toHaveLength(1)
```
pawelstepien-da, line 170: If by mistake a dev would invert condition like from `!holding.isLocked` to `holding.isLocked`, the test wouldn't catch it, because it would still be of length 1. Let's add one more assertion that validates that `result[0].id === '2'`

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+                    'splice-api-token-transfer-instruction-v1': 1,
+                    'splice-api-token-allocation-request-v1': 1,
+                    'splice-api-token-allocation-v1': 1,
+                    'splice-api-token-holding-v1': 1,
+                    'splice-api-token-allocation-instruction-v1': 1,
+                },
+            })
+
+        const response = await service.instrumentsToAsset(registryUrl)
+        expect(response).toHaveLength(2)
```
pawelstepien-da: Please also validate that objects have proper shape and values

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+                ],
+                nextPageToken: null,
+            })
+            .mockResolvedValueOnce({ adminId: 'admin-b' })
+
+        const result = await service.registriesToAssets([
+            'http://registry1.com',
+            'http://registry2.com',
+        ])
+        expect(result).toHaveLength(3)
```
pawelstepien-da: Also here let's check that items inside `result` have proper values

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+
+        vi.spyOn(service.core, 'toPrettyTransactions').mockResolvedValue({
+            nextOffset: 100,
+            transactions: [],
+        })
+        await service.listHoldingTransactions(senderParty)
+        const resources = provider.request.mock.calls.map(
+            (c: any) => c[0].params.resource
+        )
+        expect(resources).toContain('/v2/state/latest-pruned-offsets')
```
pawelstepien-da: I think we should also check that request to `/v2/updates/flats` happens and validate request body has proper values.

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+            },
+            contractId: 'proxy-cid',
+            templateId:
+                '#splice-util-featured-app-proxies:Splice.Util.FeaturedApp.DelegateProxy:DelegateProxy',
+        })
+    })
+
+    it('accepts delegate proxy transfer instruction accept', async () => {
+        const { service } = makeService()
+        vi.spyOn(service.transfer, 'createTransfer').mockResolvedValue([
```
pawelstepien-da: I think this mock is not used in that test

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+            { contractId: 'accept-cid', choiceArgument: {} } as any,
+            [],
+        ])
+
+        const ctx = makeChoiceContext()
+        vi.spyOn(
+            service.transfer,
+            'fetchAcceptTransferInstructionChoiceContext'
+        ).mockResolvedValue({ choiceContextData: ctx, disclosedContracts: [] })
```
pawelstepien-da: `makeChoiceContext` already creates `{ choiceContextData, disclosedContracts }`, so the mock double nests `choiceContextData`. I think you can just do `mockResolvedValue(ctx)` and update assertion below.

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+    })
+
+    it('create delegate proxy transfer', async () => {
+        const { service } = makeService()
+        vi.spyOn(service.transfer, 'createTransfer').mockResolvedValue([
+            { contractId: 'transfer-cid', choiceArgument: {} } as any,
+            [],
+        ])
+
+        const [exercise] = await service.createDelegateProxyTranfser(
```
pawelstepien-da: Looks like we have a typo in source code. This shouldn't be a breaking change to fix, right?

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -1,47 +1,412 @@
...
+                    ctx as any
+                )
+
+            expect(exercise.choice).toBe('AllocationFactory_Allocate')
+            expect(exercise.contractId).toBe('factory-id')
+        })
     })
+})
 
+describe('getInputHoldingsCidsForAmount', async () => {
```
pawelstepien-da: Probably doesn't break anything, but `describe` block are supposed to be sync, while insides like `beforeEach` or `it` can be async.

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -147,3 +512,732 @@ describe('getInputHoldingsCidsForAmount', async () => {
...
+
+        const [exerciseReject] =
+            await service.transfer.createRejectTransferInstruction(
+                'cid',
+                registryUrl,
+                ctx as any
+            )
+        expect(exerciseReject.choice).toBe('TransferInstruction_Reject')
+
+        //TODO: do all of these where it fetches from registry when no ctx is provided
```
pawelstepien-da, line 684: Just to make sure - this is for sometime in future, or you want to do this in that PR?

rukmini-basu-da: yeah, i'll do this in the future

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -1,47 +1,452 @@
...
-            },
-            lock: null,
-            meta: {
-                values: {},
-            },
-            amount,
+/* eslint-disable @typescript-eslint/no-explicit-any */
+const { mockAcsState, mockParseTransaction } = vi.hoisted(() => ({
+    mockAcsState: vi.fn().mockResolvedValue([]),
+    mockParseTransaction: vi.fn().mockResolvedValue({ offset: 10, events: [] }),
```
pawelstepien-da, line 16: I just realized it does nothing - actual TransactionParser from core-tx-parse is used and coincidently it worked, but values you mocked with `mockParseTransaction.mockResolvedValueOnce` changed nothing.

Add somewhere below
```
vi.mock('@canton-network/core-tx-parser', async (importActual) => {
    const actual =
        await importActual<typeof import('@canton-network/core-tx-parser')>()
    return {
        ...actual,
        TransactionParser: vi.fn(function () {
            return { parseTransaction: mockParseTransaction }
        }),
    }
})
```

You can verify that by commenting places where you do `mockParseTransaction.mockResolvedValueOnce` - with mocked import it fails as it should, without it - no effect.

core/token-standard-service/src/token-standard-service.test.ts
```diff
@@ -1,47 +1,452 @@
...
+                inputHoldingCids: ['cid1', 'cid2'],
+                requestedAt: '2026-01-01T00:00:00.000Z',
+            })
+        })
+    })
+
+    describe('createAllocationInstructionFromContext', () => {
+        const instrumentAdmin =
+            'DSO::1220c69732dd5f3b434c283f61cbc29d3bb492c50c56e306b436c3e1741cbc7be53e'
+        it('calls getInputHoldingCids and embeds resulting cids', async () => {
```
pawelstepien-da: Maybe something like

```suggestion
        it('builds an AllocationFactory_Allocate exercise command for the given factory id and context', async () => {
```

pawelstepien-da (commented):
2 more points and we are good to go

pawelstepien-da (approved):

Issue #1926, opened 2026-06-01
Configure WG with external/custom signing driver


mjuchli-da:
We currently predefine a set of signing drivers (those that reside in the repository are prepackaged) for the Wallet Gateway.
Instead, the Wallet Gateway **artifact** should support a configuration that allows en-/disabling pre-packaged signing drivers, as well as invoking signing drivers from an external package. E.g. a plug-in like experience.

If the approach with the artifact is inconvenient for some reason, one could also explore the option in which the WG itself is a dependency of a custom-WG project that includes different signing drivers.

PR #1984, opened 2026-06-11
chore: signing security disclaimer

mjuchli-da:


alexmatson-da (approved):
this probably points to needing a good way of enabling/disabling individual signing providers, with `participant` and `wallet-kernel` off by default in the Docker/Helm configs. 

Then its "enable at your own risk" rather than leaving the signing options open

Edit: Ah, we do have #1926

PR #1986, opened 2026-06-11
test: wallet sdk logging package and utils

### LINKED ISSUE #1961: Logger package unit tests

rukmini-basu-da:
closes https://github.com/canton-network/wallet/issues/1961

pino is a bit tricky to test, but it is a pretty small package and only missing 3 lines (not much of an impact in the long run)

image


sdk/wallet-sdk/src/wallet/logger/logger.test.ts
```diff
@@ -0,0 +1,115 @@
...
+
+        it('accepts CustomLogAdapter instance directly', () => {
+            const { adapter } = makeCustomAdapter()
+            expect(() => new SDKLogger(adapter)).not.toThrow()
+        })
+    })
+
+    describe('sdk logger logs correct levels based on node env', () => {
+        const isBrowserEnv = typeof process === 'undefined'
+        let log: ReturnType<typeof vi.fn<LogAdapter['log']>>
```
pawelstepien-da, line 85: LogAdapter is not defined

sdk/wallet-sdk/src/wallet/logger/logger.test.ts
```diff
@@ -0,0 +1,115 @@
...
+            expect(log).toHaveBeenCalled()
+        })
+
+        it.skipIf(isBrowserEnv)('debug', () => {
+            setNodeEnv('production')
+            logger.debug({ requestId: '123' }, 'should be supressed')
+            expect(log).not.toHaveBeenCalled()
+        })
+
+        it.skipIf(isBrowserEnv)('debug', () => {
```
pawelstepien-da: That one has confusing descriptions - `it` says debug, but .warn is called.
warn gets 'should be supressed', but it expectes log to be called

sdk/wallet-sdk/src/wallet/namespace/utils/utils.test.ts
```diff
@@ -0,0 +1,148 @@
...
+        expect(parsedUrlAsString.href).toBe('http://registry.com/path')
+        expect(parsedUrlAsString).toBeInstanceOf(URL)
+    })
+
+    it('tests ParsedURL with bad input', () => {
+        const urlAsString = 'registry.com'
+
+        try {
+            new ParsedURL(ctx, urlAsString)
+        } catch (e) {
```
pawelstepien-da, line 70: If somehow it wouldn't throw, the test wouldn't detect it, because only catch block has assertions.
You could try something like
```
let thrownError
try {
        new ParsedURL(ctx, urlAsString)
    } catch (e) {
        thrownError = e
    }

expect(thrownError).toBeInstanceOf(SDKError)
// and the rest of assertions
```

sdk/wallet-sdk/src/wallet/namespace/utils/utils.test.ts
```diff
@@ -0,0 +1,148 @@
...
+    })
+
+    it.skipIf(isBrowserEnv)('tests hash service', async () => {
+        const utils = new SDKUtilsNamespace({
+            logger: new SDKLogger('console'),
+            error: new SDKErrorHandler(new SDKLogger('console')),
+        })
+
+        const preparedTxHash = (
+            await utils.hash.preparedTransacation(pingTx)
```
pawelstepien-da: Looks like we have a typo in source code, but I believe fixing it is not worth making a breaking change.

alexmatson-da: heh, good catch... we can at least deprecate this method and introduce a properly named version to use

pawelstepien-da (approved):

PR #1987, opened 2026-06-11
feat: display list of suggested wallet extensions

### LINKED ISSUE #1886: Introduce trusted section wallet discovery
Add a `wallets.json` list to carry all trusted wallets, identified by 
- provider name
- extension id
- download url
- dapp url
- logo
- (or similar)

Document it so that wallet providers understand that they can open PR's to get listed.
Once listed, we will list the wallet in the picker. If the wallet is a browser extension and is not installed, we link to the download url (e.g. chrome store).

Image

alexmatson-da:
Fixes #1886 

This PR introduces "suggested wallets" for the wallet picker. These are a list of popular, default 3rd-party extensions one can use to setup a Canton wallet. 

how it works:

1. `wallets.json` within the dapp-sdk package maintains a canonical list of suggested extension entries, including supported browsers and relevant install URLs -- similar to `gateways.json`, but with a different schema
2. The wallet picker loads all the entries, including installed extensions that have announced themselves
3. The wallet picker filters out those suggested extensions that are already installed, by comparing `providerId`s
4. The rest are displayed under the "Suggested Wallets" header. If an extension does not have an install URL available for the user's browser, it is grayed out (but still shown)

# Before installing

Screenshot 2026-06-11 at 1 31 16 PM

# After installing

Screenshot 2026-06-11 at 1 31 39 PM

### DIFF
--- a/core/wallet-ui-components/src/components/wallet-picker.ts
+++ b/core/wallet-ui-components/src/components/wallet-picker.ts
@@ -4,7 +4,11 @@
 import { css } from 'lit'
 import { BaseElement } from '../internal/base-element'
 import { cssToString } from '../utils'
-import { WalletPickerEntry } from '@canton-network/core-types'
+import {
+    BrowserPlatform,
+    WalletPickerEntry,
+    WalletPickerSuggestedEntry,
+} from '@canton-network/core-types'
 
 export type {
     WalletPickerEntry,
@@ -106,6 +110,43 @@ const SUBSTITUTABLE_CSS = cssToString([
             flex: 1;
             overflow-y: auto;
             padding: 4px 12px 0;
+        }
+
+        .wallet-suggested-card {
+            display: flex;
+            align-items: center;
+            flex-wrap: wrap;
+            gap: 12px;
+            padding: 14px 16px;
+            border-radius: 8px;
+            border: 1px solid var(--wg-theme-border-color);
+            background: var(--wg-theme-surface-color);
+            transition: all 0.15s ease;
+            width: 100%;
+            text-align: left;
+            margin-bottom: 8px;
+        }
+
+        .wallet-suggested-card.wallet-suggested-card-disabled {
+            opacity: 0.8;
+            background: var(--wg-theme-border-color);
+        }
+
+        .btn-secondary.wallet-install-btn {
+            text-decoration: none;
+            padding: 2px 4px;
+            font-size: 12px;
+            background: var(--wg-theme-surface-color);
+        }
+
+        .wallet-install-buttons
+            > .btn-secondary.wallet-install-btn:not(:first-child) {
+            margin-left: 4px;
+        }
+
+        .wallet-install-btn:hover {
+            background: var(--wg-theme-surface-hover);
+            border-color: var(--wg-theme-accent-color);
         }
 
         .wallet-card {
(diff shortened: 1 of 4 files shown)


sdk/dapp-sdk/src/sdk.ts
```diff
@@ -58,6 +59,11 @@ export interface DappSDKConnectOptions<
 > {
     defaultAdapters?: TDefaultAdapter[]
     additionalAdapters?: ProviderAdapter[] | undefined
+    enableSuggestedWallets?: boolean
```
alexmatson-da, line 62: Defaults to true, but I figured dapp-sdk consumers should be able to opt out of showing this list

core/wallet-ui-components/src/components/wallet-picker.ts
```diff
@@ -700,6 +790,72 @@ export class WalletPicker extends HTMLElement {
...
+            if (!aIsCurrentBrowser && bIsCurrentBrowser) return 1
+
+            return a.platform.localeCompare(b.platform)
+        })
+
+        for (const { url, platform } of sortedInstallUrls) {
+            const badge = this.el('a', `Get for ${platform}`, {
+                class: 'btn-secondary wallet-install-btn',
+                href: url,
+                target: '_blank',
```
fayi-da, line 848: if we're opening in new tab then it might be worth having `rel: 'noopener'`

core/wallet-ui-components/src/components/wallet-picker.ts
```diff
@@ -700,6 +790,72 @@ export class WalletPicker extends HTMLElement {
...
+            const aIsCurrentBrowser = this.platform === a.platform
+            const bIsCurrentBrowser = this.platform === b.platform
+
+            if (aIsCurrentBrowser && !bIsCurrentBrowser) return -1
+            if (!aIsCurrentBrowser && bIsCurrentBrowser) return 1
+
+            return a.platform.localeCompare(b.platform)
+        })
+
+        for (const { url, platform } of sortedInstallUrls) {
```
fayi-da, line 843: Might be tripping but shouldn't we be going through a filtered list here so we dont show "Get for chrome" if you're on firefox and there's no extension there?

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -309,6 +316,17 @@ export class DappSDK {
             this.configuredAdapters = normalizeConnectOptions(options)
         }
 
+        const enableSuggestedWallets = defaultTrue(
+            this.configuredAdapters?.enableSuggestedWallets
+        )
+        if (enableSuggestedWallets) {
```
fayi-da, line 323: Should we clear that key in localstorage  in the else case here? Thinking of an instance where it's initially enabled and the dApp changes to false but we still have the list in localstorage

sdk/dapp-sdk/src/sdk.ts
fayi-da: i agree

core/wallet-ui-components/src/components/wallet-picker.ts
alexmatson-da: lol I agree but thats what we were discussing in standup yesterday, @mjuchli-da is concerned product people might be confused if their extension doesnt show up because its unsupported. 

I opted to move them to the bottom, grayed out

fayi-da: Ah yes, now that convo fully makes sense haha :see_no_evil: . Ok good with me

fayi-da (approved):
thanks!

Issue #1791, opened 2026-05-17
Implement offers features


fayi-da:
Figma: https://www.figma.com/design/VzQePESEI6WLMD8L29lrSI/Splice-Portfolio-App?node-id=476-10894&t=7LgMQXj0nNSowSfC-4

- [x] implement incoming and outgoing offers tab view
- [x] fetch and display all incoming offers
- [x] fetch and display all outgoing offers
- [x] view all incoming offer details(transfers and allocations)
- [x] accept/reject transfers
- [x] allocate for allocations 
    - [x] unable to allocate with no enough funds ([Figma](https://www.figma.com/design/VzQePESEI6WLMD8L29lrSI/Splice-Portfolio-App?node-id=319-20239&t=j2jQYJiDMw6Rs1vQ-4)) 
- [x] view/withdraw outgoing transfers
~- [ ] ability to search and filter offers by ????~
~- [ ] CSV export for offers~

fayi-da:
All items completed

PR #1989, opened 2026-06-12
feat: implement offers page in portfolio

fayi-da:
Part of #1791 

A lot of refactoring in this PR to ensure action required and offers pages use the same components and logic keeping it nice and DRY :wink: 

https://github.com/user-attachments/assets/79d03e14-e242-4970-8132-45ad5dcf044f


examples/portfolio/src/components/dashboard/action-required-section.tsx
```diff
@@ -68,26 +67,22 @@ export function ActionRequiredSection({
...
+                        key={offer.id}
+                        offer={offer}
+                        onClick={() => setSelectedOfferId(offer.id)}
                     />
                 ))}
             </Box>
             <ActionRequiredDialog
-                item={selectedItem}
-                onClose={() => setSelectedItemKey(null)}
+                item={selectedOffer?.source ?? null}
```
alexmatson-da, line 79: should we allow this prop to accept `undefined` instead of remapping to `null`? or is there a benefit to keeping it strict?

examples/portfolio/src/hooks/useOffers.ts
```diff
@@ -0,0 +1,140 @@
...
+function deriveOffers(items: ActionItem[]): OfferItem[] {
+    return items.flatMap((item) => {
+        if (item.kind === 'transfer') {
+            return deriveTransferOffer(item)
+        }
+        return deriveAllocationOffers(item)
+    })
+}
+
+function deriveTransferOffer(item: TransferActionItem): OfferItem[] {
```
alexmatson-da: Any reason for returning an array?

examples/portfolio/src/hooks/useOffers.ts
```diff
@@ -0,0 +1,140 @@
...
+    const directions = new Set<OfferDirection>()
+
+    for (const leg of item.transferLegs) {
+        if (isCurrentPartyReceiver(item.currentPartyId, leg)) {
+            directions.add('incoming')
+        }
+        if (isCurrentPartySender(item.currentPartyId, leg)) {
+            directions.add('outgoing')
+        }
+    }
```
alexmatson-da, line 97: I'm not familiar with the shape of allocation data, but I'm a bit confused -- this is basically just a set of at-most 2 items? 

Could you have `transferLegs` that are both incoming and outgoing? With a result of having 2 transfer offers based on the same allocation item?

examples/portfolio/src/hooks/useOfferItems.ts
```diff
@@ -0,0 +1,183 @@
...
+    )
+    const allocationRequests = useQuery(
+        useAllocationRequestsQueryOptions(primaryParty)
+    )
+    const allocations = useQuery(useAllocationsQueryOptions(primaryParty))
+
+    // Allocation requests and allocations are returned as separate contracts.
+    // Use the settlement reference plus transfer leg ID as a stable join key so
+    // each request leg can be decorated with any allocations already created.
+    const allocationKey = useCallback(
```
alexmatson-da: does this need a useCallback if there are no dependencies on react state? could even just be a function defined outside the hook

examples/portfolio/src/hooks/useOffers.ts
fayi-da: It's really because i'm using `flatMap` when calling this function :see_no_evil:   I might be able to achieve that same thing with null and retuirn [] on the caller site. I'll review.

examples/portfolio/src/hooks/useOffers.ts
fayi-da: > this is basically just a set of at-most 2 items?

Yeah. 

> Could you have transferLegs that are both incoming and outgoing? With a result of having 2 transfer offers based on the same allocation item?

Not sure I understand this question. An allocation request has multiple legs, each leg that has you as a counterparty is either incoming(you're getting something) or outgoing(you're sending something) each of which is of course just a transfer underneath.


My main issue with this actually is that allocation requests currently show in both the incoming and outgoing tab which is something I’ll be clarifying with product in our next meeting.

examples/portfolio/src/hooks/useOfferItems.ts
fayi-da: Yeah good catch. I'll move it outside the hook

examples/portfolio/src/components/dashboard/action-required-section.tsx
fayi-da: `undefined` scares me here as it could mean the prop was not provided. I find it a bit more intuitive and explicit to use null for the absense of the object to communicate "no item was selected".

examples/portfolio/src/hooks/useOffers.ts
alexmatson-da: I think what I was getting at: if the allocation request can _only_ have one direction, then you could just do some kind of `item.transferLegs.some(...)` to figure it out without the `directions` Set stuff. 

But if the same item could legitimately be both incoming and outgoing, then this approach would make sense 

> My main issue with this actually is that allocation requests currently show in both the incoming and outgoing tab

yep, basically also what im confused on

alexmatson-da (approved):
thanks!

examples/portfolio/src/hooks/useOffers.ts
fayi-da: I'll refine once i chat with Product

PR #1992, opened 2026-06-12
test: dapp-sdk unit tests

### LINKED ISSUE #1752: dApp SDK Unit Test Coverage
- 80% Coverage

### LINKED ISSUE #1861: dapp-sdk unit tests

pawelstepien-da:
Note:
dapp-sdk integration tests run together with units on yarn test / test:coverage, but only unit tests count to coverage.


sdk/dapp-sdk/vitest.config.ts
```diff
@@ -9,7 +9,11 @@ export default defineConfig({
         globalSetup: ['./vitest.global-setup.ts'],
         coverage: {
             include: ['src/**/*.ts'],
-            exclude: ['src/integration-test'],
+            exclude: [
+                'src/integration-test/**',
+                'src/**/*.test.ts',
```
alexmatson-da: does this exclude the unit tests from coverage reporting?

alexmatson-da (dismissed):

sdk/dapp-sdk/vitest.config.ts
```diff
@@ -20,8 +24,25 @@ export default defineConfig({
             },
         },
         environment: 'node',
-        include: [],
+        // include: [],
```
mateuszpiatkowski-da: I guess this is not needed anymore?

sdk/dapp-sdk/vitest.config.ts
```diff
@@ -9,7 +9,11 @@ export default defineConfig({
...
             include: ['src/**/*.ts'],
-            exclude: ['src/integration-test'],
+            exclude: [
+                'src/integration-test/**',
+                'src/**/*.test.ts',
+                'src/dapp-api/rpc-gen/**',
+            ],
             provider: 'v8',
             reporter: ['text', 'html', 'lcov'],
             thresholds: {
```
mateuszpiatkowski-da, line 15: You don't want to change the values here?

sdk/dapp-sdk/src/adapter/extension-adapter.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    StatusEvent,
+} from '@canton-network/core-wallet-dapp-rpc-client'
+import * as storage from '../storage'
+import { ExtensionAdapter } from './extension-adapter'
+
+type MockProvider = {
+    request: Mock<Provider<DappRpcTypes>['request']>
+}
+
+const connectedStatus = (): StatusEvent => ({
```
mateuszpiatkowski-da, line 26: does it need to be a function? I can see it only used in one place, so I'm not sure if we even need a const here tbh :D

sdk/dapp-sdk/src/adapter/extension-adapter.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+        expect(adapter.providerId).toBe('browser:ext:my-wallet')
+        expect(adapter.getInfo()).toEqual({
+            providerId: 'browser:ext:my-wallet',
+            name: 'My Extension',
+            type: 'browser',
+            description: 'Test extension',
+            icon: 'data:image/png;base64,abc',
+        })
+        expect(adapter.target).toBe('my-wallet')
```
mateuszpiatkowski-da, line 68: Since we're testing public props here, shouldn't we ensure that `name`, `type`, `icon` (and possibly `target`) are properly assigned to?

sdk/dapp-sdk/src/adapter/extension-adapter.test.ts
```diff
@@ -0,0 +1,155 @@
...
+    })
+
+    it('uses defaults when config is omitted', () => {
+        const adapter = new ExtensionAdapter()
+
+        expect(adapter.providerId).toBe('browser')
+        expect(adapter.getInfo().name).toBe('Browser Extension')
```
mateuszpiatkowski-da, line 75: I'd add the `description` property to test as well via `getInfo()`

sdk/dapp-sdk/src/adapter/extension-adapter.test.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const connectedStatus = (): StatusEvent => ({
+    provider: { id: 'browser:ext:test' },
+    connection: {
+        isConnected: true,
+        isNetworkConnected: true,
+    },
+})
+
+const makeMockProvider = (): MockProvider => ({
```
mateuszpiatkowski-da, line 34:
```suggestion
const mockProvider = vi.hosted(() => ({
```

This way you don't have to make a call to create a new object each time, as `vi.restoreAllMocks` will handle the reset of `vi.fn()` spy (it's nothing serious though :P )

sdk/dapp-sdk/src/storage.test.ts
```diff
@@ -0,0 +1,124 @@
...
+                url: 'https://gateway.example.com',
+            })
+
+            expect(getKernelDiscovery()).toEqual({
+                walletType: 'remote',
+                url: 'https://gateway.example.com',
+            })
+        })
+
+        it('round-trips extension discovery through localStorage', () => {
```
mateuszpiatkowski-da, line 63: It seems for me as this test does exactly the same as the one above it. Are you sure we need them both?

sdk/dapp-sdk/src/adapter/remote-adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+    clearAllLocalState: vi.fn(),
+}))
+
+vi.mock('@canton-network/core-wallet-ui-components', () => ({
+    popup: {
+        close: vi.fn(),
+    },
+}))
+
+const kernelSession = (): StatusEvent => ({
```
mateuszpiatkowski-da, line 93: I guess it doesn't necessarily need to be a function if it's just an object which we read (function is ok though)

sdk/dapp-sdk/src/adapter/remote-adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+        mockController.prepareExecute.mockResolvedValue(null)
+        mockController.prepareExecuteAndWait.mockResolvedValue({
+            tx: { commandId: 'cmd-1', status: 'executed' },
+        })
+        mockController.signMessage.mockResolvedValue({ signature: 'sig' })
+        mockController.ledgerApi.mockResolvedValue({ ok: true })
+    })
+
+    afterEach(() => {
+        vi.restoreAllMocks()
```
mateuszpiatkowski-da, line 127: Are you sure we need it? We have the clearing of mocks in `beforeEach` hence the question

sdk/dapp-sdk/src/adapter/remote-adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+        mockController.connect.mockResolvedValue(kernelSession().connection)
+        mockController.disconnect.mockResolvedValue(null)
+        mockController.isConnected.mockResolvedValue(kernelSession().connection)
+        mockController.listAccounts.mockResolvedValue([])
+        mockController.prepareExecute.mockResolvedValue(null)
+        mockController.prepareExecuteAndWait.mockResolvedValue({
+            tx: { commandId: 'cmd-1', status: 'executed' },
+        })
+        mockController.signMessage.mockResolvedValue({ signature: 'sig' })
+        mockController.ledgerApi.mockResolvedValue({ ok: true })
```
mateuszpiatkowski-da, line 123: if you want we can do sth like this:
```suggestion
mockController.ledgerApi.mockResolvedValue({ ok: true })

adapter = new RemoteAdapter({
  name: 'Gateway',
  rpcUrl: RPC_URL
})
``` 
and just declare the var via `let` above

this way you can omit the need to create the adapter each time you want to use it.

sdk/dapp-sdk/src/adapter/remote-adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+        const listener = vi.fn<EventListener<StatusEvent>>()
+
+        provider.on('statusChanged', listener)
+        provider.emit('statusChanged', kernelSession())
+
+        expect(mockDappAsyncProvider.on).toHaveBeenCalledWith(
+            'statusChanged',
+            listener
+        )
+        expect(mockDappAsyncProvider.emit).toHaveBeenCalledWith(
```
mateuszpiatkowski-da, line 245: we don't want to check if `listener` has beed called too? I'm not into the business logic as much so I'm just not sure about it tho :p

sdk/dapp-sdk/src/adapter/remote-adapter.test.ts
```diff
@@ -0,0 +1,340 @@
...
+        expect(clearAllLocalState).toHaveBeenCalledWith({ closePopup: true })
+    })
+
+    it('closes the popup on teardown', () => {
+        new RemoteAdapter({ name: 'Gateway', rpcUrl: RPC_URL }).teardown()
+
+        expect(popup.close).toHaveBeenCalled()
+    })
+
+    describe('restore', () => {
```
mateuszpiatkowski-da, line 294: I'd also add the test when `discovery.walletType !== 'remote'`

mateuszpiatkowski-da (commented):
To be continued 😝

sdk/dapp-sdk/vitest.config.ts
pawelstepien-da: Yes, but it wasn't needed, vitest/v8 excludes .test files automatically from the coverage. I removed it.

sdk/dapp-sdk/src/adapter/walletconnect-adapter.test.ts
```diff
@@ -0,0 +1,292 @@
...
+) =>
+    WalletConnectAdapter.create({
+        projectId: 'test-project-id',
+        ...overrides,
+    })
+
+describe('WalletConnectAdapter', () => {
+    beforeEach(() => {
+        vi.clearAllMocks()
+        mockSignClient.session.getAll.mockReturnValue([])
```
mateuszpiatkowski-da, line 76: I think that we can basically put `.mockReturnValue` or `.mockResolvedValue` next to `vi.fn()` above (although I'm not sure about that 100% and it's not a deal breaker)

sdk/dapp-sdk/src/adapter/walletconnect-adapter.test.ts
```diff
@@ -0,0 +1,292 @@
...
+        mockSignClient.connect.mockResolvedValue({
+            uri: 'wc:test-uri',
+            approval: vi.fn().mockResolvedValue(mockSession),
+        })
+    })
+
+    afterEach(() => {
+        vi.restoreAllMocks()
```
mateuszpiatkowski-da, line 84: I don't think we need it if we have `vi.clearAllMocks` above already (in beforeEach)

sdk/dapp-sdk/src/adapter/walletconnect-adapter.test.ts
```diff
@@ -0,0 +1,292 @@
...
+
+const makeAdapter = (
+    overrides: Partial<
+        Pick<
+            WalletConnectAdapterConfig,
+            'onUri' | 'onSignInWithCanton' | 'signInWithCanton'
+        >
+    > = {}
+) =>
+    WalletConnectAdapter.create({
```
mateuszpiatkowski-da, line 68: I'm thinking if it's sensible to use .create since the constructor itself is public. I think that this is a tech debt and .create should be marked as deprecated (unless there is a reason for it's existence which I believe should be documented then)

OFC this is not on this PR, but I'm thinking about:
```suggestion
    new WalletConnectAdapter({
```

sdk/dapp-sdk/src/announce-discovery.test.ts
```diff
@@ -0,0 +1,80 @@
...
+
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import {
+    CANTON_ANNOUNCE_PROVIDER_EVENT,
+    CANTON_REQUEST_PROVIDER_EVENT,
+} from '@canton-network/core-types'
+import { requestAnnouncedProviders } from './announce-discovery'
+
+describe('requestAnnouncedProviders', () => {
+    afterEach(() => {
```
mateuszpiatkowski-da, line 12:
```suggestion
    beforeEach(() => {
        vi.useFakeTimers()
    })

    afterEach(() => {
        vi.useRealTimers()
```

This way we can omit waiting for real time to pass by when passing timeout

sdk/dapp-sdk/src/client.test.ts
```diff
@@ -0,0 +1,210 @@
...
+
+const makeProvider = (): MockDappProvider => ({
+    request: vi.fn(),
+    on: vi.fn(),
+    removeListener: vi.fn(),
+})
+
+// MockDappProvider when calling vitest mocking methods
+// Provider<DappRpcTypes> when passing to source code methods
+const asProvider = (mock: MockDappProvider): Provider<DappRpcTypes> =>
```
mateuszpiatkowski-da, line 42: I think we can simplify the logic by connecting `makeProvider` with the type assertion here. This way we can omit this function

mateuszpiatkowski-da: Also, we can bring the assignment of provider to `beforeEach` clause to reduce code duplication (this way we don't have to worry about assigning/creating mock provider for each test as it would be done by us)

mateuszpiatkowski-da (approved):
Basically 99% of the comments (if not all of them) are improvements, hence the ✅ . I suppose it would be nice to look into them though in a separate improvement task :)

mateuszpiatkowski-da:
All comments included in https://github.com/canton-network/wallet/issues/2027

PR #1994, opened 2026-06-12
test(wallet-sdk): add party namespace tests

### LINKED ISSUE #1975: Party namespace unit tests

mateuszpiatkowski-da:


alexmatson-da (dismissed):
thanks!

sdk/wallet-sdk/src/wallet/namespace/party/party.test.ts
```diff
@@ -0,0 +1,762 @@
...
+                        {
+                            method: 'ledgerApi',
+                            params: {
+                                resource: '/v2/parties/participant-id',
+                                requestMethod: 'get',
+                            },
+                        }
+                    )
+                })
+                expect(ledgerProvider.request).toHaveBeenCalledTimes(6)
```
pawelstepien-da: I think this could potentially be a bit flaky, because:
-observing and confirming participant endpoints calls request 5 times
-resolveSynchronizerId 1 request call
and if their resolution wins the race before this expect, `Promise.all().then` executes and there would be one more call to `/v2/parties/external/generate-topology`.

Maybe the concern is invalid with how event loop works, but feels brittle and I would suggest checking that sixth call is to `/v2/state/connected-synchronizers` instead that number of calls equals 6.

sdk/wallet-sdk/src/wallet/namespace/party/party.test.ts
```diff
@@ -0,0 +1,762 @@
...
+                                party: partyTransaction.partyId,
+                            },
+                        ],
+                    } satisfies LedgerCommonSchemas['GetPartiesResponse'])
+                    // Mock grantRights call - fails to grant rights
+                    .mockResolvedValueOnce({
+                        newlyGrantedRights: undefined,
+                    } satisfies LedgerCommonSchemas['GrantUserRightsResponse'])
+
+                await expect(signedParty.execute()).rejects.toThrow()
```
pawelstepien-da: Could you also check what error is thrown - we throw custom errors, but if we changed the implementation and introduced a bug that would throw error we don't expect, i.e. something like "Can't access x of undefined", the test could potentially give us a false positive.
Same for other 3 test cases with `.rejects.toThrow()` in that file.

pawelstepien-da (approved):

PR #2000, opened 2026-06-15
feat: show insufficient funds alert for allocations

fayi-da:
Part of #1791 

When there are not enough funds for allocation:
- disable allocate button
- show error message 

screenshot-2026-06-15_13-32-57

### DIFF
--- a/examples/portfolio/src/components/dashboard/allocation-action-dialog-content.tsx
+++ b/examples/portfolio/src/components/dashboard/allocation-action-dialog-content.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useState, type ReactNode } from 'react'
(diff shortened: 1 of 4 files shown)


examples/portfolio/src/components/dashboard/allocation-action-dialog-content.tsx
```diff
@@ -163,6 +178,9 @@ export function AllocationActionDialogContent({
                             isExpanded={expandedLegIds.has(leg.transferLegId)}
                             isLoading={isLoading}
                             isLegLoading={loadingLegId === leg.transferLegId}
+                            hasInsufficientFunds={insufficientLegIds.has(
+                                leg.transferLegId
+                            )}
                             error={
```
alexmatson-da, line 184: any reason to add a new prop rather than pass the error message to the `error` prop?

alexmatson-da (approved):

examples/portfolio/src/components/dashboard/allocation-action-dialog-content.tsx
fayi-da: na not really. I can certainly combine. Will fix in a follow up 👍

PR #2010, opened 2026-06-15
fix: minor error cleanup in portfolio dapp

fayi-da:
As requested by @alexmatson-da  in  https://github.com/canton-network/wallet/pull/2000


alexmatson-da (approved):
thanks!

PR #2030, opened 2026-06-17
test: token namespace tests

### LINKED ISSUE #1979: Token namespace unit tests

rukmini-basu-da:
Closes https://github.com/canton-network/wallet/issues/1979
image

### DIFF
--- a/sdk/wallet-sdk/src/wallet/namespace/token/allocation/allocation.test.ts
+++ b/sdk/wallet-sdk/src/wallet/namespace/token/allocation/allocation.test.ts
@@ -1,0 +1,310 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { describe, it, vi, beforeEach, expect } from 'vitest'
+import { mock } from '../../../__test__/mocks'
+import { TokenNamespaceConfig } from '../index'
+import { ParsedURL } from '../../utils/url'
+import { AllocationNamespace } from './service'
+import {
+    AllocationContextParams,
+    AllocationInstructionCreateParams,
+} from './types'
+import { ALLOCATION_REQUEST_INTERFACE_ID } from '@canton-network/core-token-standard'
+/* eslint-disable @typescript-eslint/no-explicit-any */
+
+const { ctx, mockLogger } = mock
+
+const mockTokenStandard = {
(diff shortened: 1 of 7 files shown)


sdk/wallet-sdk/src/wallet/namespace/token/allocation/allocation.test.ts
```diff
@@ -0,0 +1,310 @@
...
+import { mock } from '../../../__test__/mocks'
+import { TokenNamespaceConfig } from '../index'
+import { ParsedURL } from '../../utils/url'
+import { AllocationNamespace } from './service'
+import {
+    AllocationContextParams,
+    AllocationInstructionCreateParams,
+} from './types'
+import { ALLOCATION_REQUEST_INTERFACE_ID } from '@canton-network/core-token-standard'
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
mateuszpiatkowski-da, line 14: I've seen that for some packages where we have tests we use `MockedObject<Entity>`. Have you tried using this? Maybe this way we could get rid of `any`s ;)

mateuszpiatkowski-da: same in other test files ;)

sdk/wallet-sdk/src/wallet/namespace/token/utxos/utxos.test.ts
```diff
@@ -0,0 +1,323 @@
...
+    registryUrls: [new ParsedURL(ctx, 'http://registry.com')],
+    tokenStandardService: mockTokenStandard as unknown as any,
+    validatorParty: 'validatorParty::123',
+}
+
+function makeUtxos(batchSize: number) {
+    const utxos = []
+    for (let i = 0; i < batchSize; i++) {
+        const utxo = {
+            contractId: v4(),
```
mateuszpiatkowski-da, line 46: it's more of a bonus, but I don't think we need to use v4 actually. I think we could simply pass a static string

mateuszpiatkowski-da (approved):
LGTM, the only thing that I'd change is the removal of `any`s if possible to get the most out of Typescript :) 

Thank you!

sdk/wallet-sdk/src/wallet/namespace/token/allocation/allocation.test.ts
rukmini-basu-da: will resolve this in a follow up pr as discussed

PR #2034, opened 2026-06-18
fix: remove auth requirement for registry in portfolio

fayi-da:


examples/portfolio/src/components/preapproval-contract-settings.tsx
```diff
@@ -113,13 +115,19 @@ export function PreapprovalContractSettings() {
             return
         }
 
-        console.log('registryUrl', registryUrl)
         const normalizedRegistryUrl = new URL(registryUrl).toString()
-        console.log('normalizedRegistryUrl', normalizedRegistryUrl)
-        console.log('sdk.asset.list', sdk.asset.list)
```
fayi-da, line 119: lol not sure how i didnt remove these previously

examples/portfolio/src/components/preapproval-contract-settings.tsx
```diff
@@ -57,6 +58,7 @@ const formatRegistryOption = (partyId: string, registryUrl: string) =>
 
 export function PreapprovalContractSettings() {
     const registryUrls = useRegistryUrls()
+    const instruments = useInstruments()
```
fayi-da, line 61: because the sdk uses auth so switching to the caveman service

examples/portfolio/src/services/resolve.ts
```diff
@@ -55,7 +55,7 @@ const createTokenStandardService = async ({
     const tokenStandardService = new TokenStandardService(
         provider,
         logger,
-        accessTokenProvider ?? defaultAccessTokenProvider({ logger }), // access token provider
+        accessTokenProvider ?? noAuthAccessTokenProvider,
```
alexmatson-da: same here, just drop

examples/portfolio/src/services/resolve.ts
```diff
@@ -39,7 +39,7 @@ const createTokenStandardClient = async ({
     return new TokenStandardClient(
         registryUrl,
         logger,
-        accessTokenProvider ?? defaultAccessTokenProvider({ logger }) // access token provider
+        accessTokenProvider ?? noAuthAccessTokenProvider
     )
```
alexmatson-da, line 41: can just drop these, even from the function params

fayi-da: dropped from the func but still need to pass an empty provider to the client

alexmatson-da (approved):
thx!

PR #2013, opened 2026-06-16
feat: add backend support for API keys

alexmatson-da:
Fixes #1999

PR #2039, opened 2026-06-18
feat: support API key mgmt in frontend

### LINKED ISSUE #2001: Implement API keys frontend
Requirements:

- Add a "Service Accounts" menu item + page to the WG UI
- Display a button to generate a new key. 
    - Allow the user to specify a human readable name to the key
    - After generating, the UI should show the key to the user only once and not store it anywhere. 
- Display all existing SA keys
- Display a button to revoke a key
- In the `Activities` page, specially mark transactions that were executed by a SA

alexmatson-da:
Fixes #2001 

Depends on #2013 

https://github.com/user-attachments/assets/d508b3bc-c0db-4ed5-b0ad-269215061406


wallet-gateway/remote/src/web/frontend/api-keys/add/index.ts
```diff
@@ -0,0 +1,164 @@
...
+                    ></wg-copy-button>
+                </div>
+
+                <div class="form-actions">
+                    <button
+                        class="btn btn-primary btn-sm rounded-pill"
+                        type="button"
+                        @click=${this.onConfirm}
+                    >
+                        Continue
```
fayi-da, line 158: This button(from the video) doesn't seem to match the other buttons in the app. some sizing and padding tweaks needed

fayi-da (approved):
🫡

fayi-da (changes requested):
We need a toast after revoking and perhaps a confirmation dialog before actually revoking

core/wallet-ui-components/src/components/api-key-form.ts
```diff
@@ -0,0 +1,241 @@
...
+                border-color: var(--wg-input-border-focus);
+                box-shadow: 0 0 0 3px rgba(var(--wg-accent-rgb), 0.12);
+            }
+
+            .select-wrap {
+                position: relative;
+            }
+
+            .select-wrap .field-control {
+                padding-right: 40px;
```
mateuszpiatkowski-da, line 82: just a nitpick - shouldn't we use rems instead? I mean I've seen two approaches (one for pxs, and another for rems) so this question is more like for me to get an idea of how we approach the units here ;P

wallet-gateway/remote/src/user-api/controller.ts
mateuszpiatkowski-da, line 1083: nitpick - I think Promise<void> is enough ;)

mateuszpiatkowski-da (dismissed):
LGTM, left a few nitpicks ;)

alexmatson-da:
> We need a toast after revoking and perhaps a confirmation dialog before actually revoking

added a toast, but planning to follow up with a confirmation dialog separately: #2043

wallet-gateway/remote/src/user-api/controller.ts
alexmatson-da: I think GH glitched out, this file wasn't touched in this PR

fayi-da (approved):

core/wallet-ui-components/src/components/api-key-form.ts
alexmatson-da: good question, I think it still depends on the exact use case. rems are good for responsive design (like spacing between sections or different UI components), but sometimes you'll may still want the precision of px for specific things (e.g. like "these two icons should always be exactly 8px apart")  

I dont think we have any formal approach in the repo though

PR #2044, opened 2026-06-18
feat: accept ApiKey authorization headers

### LINKED ISSUE #2038: Allow ApiKey based authentication
In our auth middleware, support `Authorization: ApiKey ...` type auth headers

alexmatson-da:
Fixes #2038 

depends on #2039 

Adds `examples/service-accounts` to showcase a Python service account using API Keys


wallet-gateway/remote/src/init.ts
```diff
@@ -332,15 +333,13 @@ export async function initialize(opts: CliOptions, logger: Logger) {
...
-    app.use(
-        '/api/*splat',
-        jwtAuth(authService, logger.child({ component: 'JwtHandler' }))
-    )
-    app.use('/api/*splat', postAuthRateLimit)
     app.use(
         '/api/*splat',
+        express.json(),
+        preAuthRateLimit,
+        apiKeyAuth(store, logger.child({ component: 'ApiKeyHandler' })),
```
mjuchli-da: We have to exclude user-api here.

alexmatson-da: done

wallet-gateway/remote/src/middleware/apiKeyAuth.ts
```diff
@@ -0,0 +1,97 @@
...
+
+            const context: AuthContext = {
+                isApiKey: true,
+                userId: matchingKey.userId,
+                accessToken: apiKey,
+                email: matchingKey.email || undefined,
+            }
+
+            // automatically initiate a session for the API key user
+            store.withAuthContext(context).setSession({
```
fayi-da: Checked and `setSession` returns a promise. Any reason we're not awaiting? Surprised out checks didn't catch that

fayi-da (dismissed):
👍

wallet-gateway/remote/src/middleware/apiKeyAuth.ts
alexmatson-da: ooh thanks, fixed

> Surprised out checks didn't catch that

seems like we might need to switch to [recommended-type-check](https://typescript-eslint.io/users/configs/#recommended-type-checked) rather than plain recommended for our ts-eslint config, to include `no-floating-promises`

fayi-da (approved):

PR #2045, opened 2026-06-18
feat: support service account auth config for ledger API access

alexmatson-da:
depends on #2044 

this PR adds a new network config option for `serviceAccountAuth`, which must be a m2m oauth client corresponding to the ledger user to use for submitting transactions on behalf of any automations using API keys.


wallet-gateway/remote/src/middleware/apiKeyAuth.ts
```diff
@@ -62,20 +66,60 @@ export function apiKeyAuth(
...
                 userId: matchingKey.userId,
-                accessToken: apiKey,
-                email: matchingKey.email || undefined,
-            }
+                accessToken: 'unused',
+            })
 
             // automatically initiate a session for the API key user
-            store.withAuthContext(context).setSession({
+            authStore.setSession({
```
fayi-da: `await` or deal with race issues later 😛

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -676,11 +675,16 @@ export const userController = (
...
+                    executeParams,
+                    transaction,
+                    wallet,
+                    userId: connectedContext.userId,
+                },
             })
 
             const response = await transactionService.execute(
-                connectedContext,
+                connectedContext.userId,
```
fayi-da, line 687: we use the `userId` here but the `ledgerUserId` in the dapp-api. same thing?

fayi-da (dismissed):

wallet-gateway/remote/src/user-api/controller.ts
alexmatson-da: this is the `user` API controller, which won't be used/called by service accounts -- so the userId should be consistent

fayi-da (dismissed):

fayi-da (approved):

PR #1622, opened 2026-04-14
feat: multi sync example

Viktor-Kalashnykov-da:
## Multi-Synchronizer DvP Example - On ledger API - Part 1

Showcases automatic reassignment of Token from private synchronizer to global one, used on a realistic trade scenario with use of a custom Token.

New test: 
https://github.com/canton-network/wallet/pull/1622/changes#diff-37db720ea457076fe6d04f6938d19c44e5dde3d7e1a5815953ff6b83199a1052

Design principles:
1. !!! We remove handling of synchronizers inside wallet-sdk.
Client code should be responsible of selecting `synchronizerId` in multi-sync scenarios.
If not provided - synchronizer choice is left for canton participant logic.
Existing tests are updated (#1740).

2. We try to show "realistic scenario" - we use multiple parties (Alice, Bob, TokenAdmin, TradingApp) we also 
distribute dars to selected synchronizers (private dars to private). This contributes to relative complexity of the test code

image

source:
https://docs.google.com/presentation/d/1q6LpHi-wC_MO_mzf7wp5D-15j4lNeKjCMW5xmBggaTY

3. Example works with Token Standard V1
We have experimented with token standard V2 -> but as V2 is not yet merged to splice use of V2 means even more code

4. All tests run on multi-sync

**This is a first PR in a series** 
Showcases only on-ledger part - and ensures code works.
**There are 2 follow up PRs (wip)**
- ~~off ledger api for tokens standard Implementation https://github.com/canton-network/wallet/pull/1782~~
(code extracted and merged here)
- rework of synchronizerId https://github.com/canton-network/wallet/pull/1740/ - (choice of synchronizer removed from wallet sdk)

**Technical limitations:**
In this PR we introduce separate tests for multi-sync (otherwise regular  tests are flaky on multi-sync) - that split will be removed in a separate PR (above)

**New example script:**
docs/wallet-integration-guide/examples/scripts/15-multi-sync-trade.ts

## Notes
1. explicit reassignment
we experimented with automatic reassignment of contracts - and it basically worked BUT.

It worked as long as `Bob` is owner of `TokenRules` contract which seems unrealistic.
We introduced TokenAdmin as additional party that is issuer of `Token` on `app-synchronizer` (private one).
But this in fact forces us to use explicit reassignment.

2.  TestToken on global synchronizer. The original intention seemed to be that TestToken dar is only vetted on private synchronizer.
But since the settlement actually happens on global - TestToken related models must be also on global synchronizer

PR #2046, opened 2026-06-19
[Draft] Auto reassignment fail

jarekr-da:


docs/wallet-integration-guide/examples/scripts/15-multi-sync/REASSIGN.md
```diff
@@ -0,0 +1,125 @@
...
+When the allocation command is prepared on the **global** synchronizer, Canton
+must co-locate every input/used contract on that synchronizer, so it triggers an
+**automatic reassignment**. Canton's rule is that **the submitter must always be
+a stakeholder of every contract it reassigns** (`SUBMITTER_ALWAYS_STAKEHOLDER`).
+
+The blocker is **not** Bob's own `Token`: Bob is the `owner`, hence a
+stakeholder, and indeed the _explicit_ `reassign` submitted by Bob always
+worked. The blocker is the **`TokenRules` factory**, whose only stakeholder is
+`admin`. With a **separate `TokenAdmin` party**, the submitter (Bob) is **not** a
+stakeholder of `TokenRules`, so Canton refuses to reassign it automatically.
```
meiersi-da: Yes, the unreassignability of the `TokenRules` is why I'm pushing for guidelines where the workflows tell the token standard registry what the target synchronizer is. See this conversation: https://github.com/canton-network/wallet/pull/1622#pullrequestreview-4521776090

For that to work cleanly we'll need a small amendment to the token standard off-ledger APIs. However, until then we can funnel the information in an additional field in the `choiceArgument` submitted to the factory endpoints; e.g. a top-level `_targetSynchronizerId` field, which you then parse in the `TestTokenV1` registry handlers.

jarekr-da: I do not understand this :-(
I tried to play around - but do not get how would this additional field help in fundamental problem, that contract cannot be reassigned automatically

meiersi-da: The trick is that the `TestTokenV1` registry maintains `TokenRules` contract on every synchronizer it supports and selects the right one based on the communicated target synchronizer. This guarantees that `TokenRules` contracts never need to be reassigned by the client submitting a token standard action.

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_setup.ts
```diff
@@ -208,35 +231,28 @@ export async function setupMultiSyncTrade(
...
-    const amuletAdmin = (amuletRules.payload as Record<string, unknown>)[
-        'dso'
-    ] as string
+    const { admin: amuletAdmin } = await appAliceSdk.asset.find('Amulet')
     logger.info(`Amulet asset discovered — admin: ${amuletAdmin}`)
 
     return {
-        appUserSdk,
-        appProviderSdk,
+        appAliceSdk,
```
meiersi-da: side question: why is this "app"AliceSdk? What is "app'y" about this? I'd have expected "aliceSdk".

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_token_transfer.ts
```diff
@@ -46,18 +46,18 @@ export async function aliceSelfTransferToApp(
 
     // The settled holding lands on the global synchronizer; move it to the
     // app-synchronizer before self-transferring there (mirrors Bob's flow).
```
meiersi-da, line 48: app-synchronizer vs testTokenSynchronizerId ... seems inconsistent...

docs/wallet-integration-guide/examples/scripts/15-multi-sync/index.ts
```diff
@@ -131,7 +140,18 @@ const testTokenAllocationDisclosed = {
     synchronizerId: '',
 }
```
meiersi-da, line 138: This does not match what we would expect to happen in reality. The trading app would be listening on their participant node to observe the allocation and use it only then. I'd suggest to implement the code here such that you wait for it to appear on the participant node used for the submission of the settlement.

docs/wallet-integration-guide/examples/scripts/15-multi-sync/index.ts
```diff
@@ -152,20 +172,20 @@ try {
     await bobSelfTransferToApp(setup, logger)
```
meiersi-da: What is that? Why do we have to wait for a self-transfer?

docs/wallet-integration-guide/examples/scripts/15-multi-sync/index.ts
```diff
@@ -152,20 +172,20 @@ try {
...
     logger.info('Contracts after settlement failure (compensation applied):')
```
meiersi-da: Why this compensation?

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_trade_settle.ts
meiersi-da: This does not match what we'd expect to happen on a settlement failure. The default would be that the settlement venue tries again, and if they fail to do so, they would cancel the allocations that exist.

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_trade_settle.ts
```diff
@@ -113,9 +113,9 @@ export async function settleOtcTrade(
     logger: Logger
 ): Promise<void> {
     const {
-        svSdk,
-        tokenNamespaceAppUser,
-        tokenNamespaceAppProvider,
+        tradingAppSdk,
+        aliceTokenNamespace,
+        bobTokenNamespace,
```
meiersi-da: using these two SDK's feels wrong. It should be only the trading app that uses their access to their participant node to build the command!

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_trade_settle.ts
```diff
@@ -129,27 +129,27 @@ export async function settleOtcTrade(
...
+    const amuletExecCtx = await aliceTokenNamespace.allocation.context.execute({
+        allocationCid: amuletAllocation.contractId,
+        registryUrl: localNetStaticConfig.LOCALNET_REGISTRY_API_URL,
+    })
 
     // Fetch Bob's TestToken execute-transfer choice context from the registry's
     // allocation-v1 API (instead of hard-coding an empty context).
-    const testTokenExecCtx =
-        await tokenNamespaceAppProvider.allocation.context.execute({
+    const testTokenExecCtx = await bobTokenNamespace.allocation.context.execute(
```
meiersi-da: here as well: use the tradingApp's SDK to fetch this information.

meiersi-da (commented):
Some comment from my review to get context on what might be failing,

docs/wallet-integration-guide/examples/scripts/15-multi-sync/_trade_settle.ts
jarekr-da: ~~there is only one sdk "handle" used (tradingAppSdk)~~ 
I see. xyzTokenNamespace are actually nod needed here

Issue #2006, opened 2026-06-15
Settings Page


fayi-da:
- [x] Amulet pre-approvals
- [x] DA Utility pre-approvals
- [x] Registry management
- [x] CC Tap

PR #2049, opened 2026-06-22
feat: registry management in new portfolio dApp

fayi-da:
Part of #2006 

https://github.com/user-attachments/assets/2a8031ea-50b6-47ca-8cef-3483d355bf6b


alexmatson-da (approved):
looks good! not sure if this is already in the plans, but we should consider keeping track of network status per registry, in case there's any registry URL that is unreachable, and display that somewhere (could be just as a 🟢 / 🔴 , or something more). 

some reasons that could happen

- The user put in the wrong URL
- The server hosting the registry API is having an outage
- The client cant access the network

extra bonus: if we catch this then we can also stop the UI from repeatedly polling a broken registry URL too

PR #2050, opened 2026-06-22
feat: devnet tap in new portfolio dApp

fayi-da:
Part of #2006 

https://github.com/user-attachments/assets/47bb850f-5d7a-4aaa-9107-891d1aaa63a5

### DIFF
--- a/examples/portfolio/src/components/settings/devnet-tap-dialog.tsx
+++ b/examples/portfolio/src/components/settings/devnet-tap-dialog.tsx
@@ -1,0 +1,526 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { useMemo } from 'react'
+import CloseIcon from '@mui/icons-material/Close'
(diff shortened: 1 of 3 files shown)


examples/portfolio/src/components/settings/devnet-tap-dialog.tsx
```diff
@@ -0,0 +1,526 @@
...
+                    >
+                        {(field) => {
+                            const selectedKey = field.state.value
+                                ? instrumentKey(field.state.value)
+                                : ''
+                            const selectDisabled =
+                                disabled || availableInstruments.length === 0
+
+                            return (
+                                <FieldBlock label="Select instrument">
```
alexmatson-da, line 225: is it even possible to tap anything besides Amulet?

alexmatson-da (approved):

examples/portfolio/src/components/settings/devnet-tap-dialog.tsx
fayi-da: Honestly, no and we talked about his today actually. Will see if we can determine the asset and always map to amulet later but need product's input

PR #2051, opened 2026-06-22
chore: cleanups for devnet tap and registry management

fayi-da:
- minro refactoring to devnet tap dialog
- prevent users from saving an invalid registry combo

Follow ups for #2006 

https://github.com/user-attachments/assets/e00fc069-c332-4495-9543-d7f712211ca3


alexmatson-da (approved):
nice

PR #2053, opened 2026-06-23
fix: enable non-primary wallet submission for prepareExecute method

### LINKED ISSUE #2031: Allow submitting commands with actAs separate from primary wallet account

alexmatson-da:
Fixes #2031


fayi-da (approved):
Is this what i think it is? First step in getting rid of primary party?

alexmatson-da:
> Is this what i think it is? First step in getting rid of primary party?

well, not getting rid of it, but supporting submissions with non-primary parties :D

PR #2059, opened 2026-06-25
fix: vaults select for fireblocks

### LINKED ISSUE #2060: Allow vault selection in wallet creation
There is hardcoded vault name that is always used when creating a fireblocks wallet "Canton Party".
It should be configurable during creation.

pawelstepien-da:
fixes:  #2060

### DIFF
--- a/wallet-gateway/remote/src/web/frontend/parties/add/index.ts
+++ b/wallet-gateway/remote/src/web/frontend/parties/add/index.ts
@@ -16,6 +16,7 @@ import { SigningProvider } from '@canton-network/core-signing-lib'
 import { createUserClient } from '../../rpc-client'
 import { setLocationHref } from '../../navigation.js'
 import { stateManager } from '../../state-manager'
+import { showToast } from '../../utils.js'
 import '../../index'
 import { WALLET_CREATION_STATUS_CODE } from '../index'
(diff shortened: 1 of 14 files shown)


wallet-gateway/remote/src/web/frontend/parties/add/index.ts
```diff
@@ -65,12 +71,49 @@ export class UserUiAddParty extends BaseElement {
...
+
+        try {
+            const userClient = await createUserClient(
+                stateManager.accessToken.get()
+            )
+            const result = await userClient.request({
+                method: 'listSigningProviderVaults',
+                params: { signingProviderId },
+            })
+            this.vaults = result.vaults
```
alexmatson-da: there could be a lot of results, at the least we should sort these by name so people can find a particular vault more easily

alexmatson-da (dismissed):
nice!

wallet-gateway/remote/src/web/frontend/parties/add/index.ts
pawelstepien-da: Good idea. Added sort() and adjusted test. Maybe in future we could something like autocomplete combobox for finding results easier.

alexmatson-da (approved):

PR #2064, opened 2026-06-27
ci: pin action versions

PascalHonegger:
According to the GitHub guide on using third-party actions, it's good practice to pin GitHub Actions using their full-length commit SHA: https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions

I also updated the versions and removed some duplicate node-version specification - but this does update the Node.js runtime from `v22.16.0` to `v24.9.0` for the publish task.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
             mainnet_splice_version: ${{ steps.version.outputs.mainnet_splice_version }}
         steps:
             - name: Checkout
-              uses: actions/checkout@v6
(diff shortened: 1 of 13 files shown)


.github/workflows/build.yml
PascalHonegger, line 111: For some reason this variable was empty on the previous CI run, not sure if that has to do with permissions, bad timing or an actual side-effect form the other updates (which would surprise me)

mjuchli-da: It's a permission thing unfortunately. Quickly need to check if there's another option than someone from DA opening a new PR

mjuchli-da: #2071 is attempt to resolve this.

mjuchli-da (approved):

PR #2065, opened 2026-06-29
feat: make coin type configurable with backward-compatible

HelmiTrabelsi:
## Motivation

`CC_COIN_TYPE` was hardcoded to `6767` inside `FireblocksHandler`, making it
impossible to configure a different derivation path coin type without forking
the package. This change makes it configurable while keeping full backward
compatibility.

## Changes

- Add optional `coinType?: number` to `FireblocksConfig` and `FireblocksHandler`
  constructor. It defaults to `CC_COIN_TYPE` (6767) when omitted
- `CC_COIN_TYPE = 6767` remains exported from `core-signing-lib` unchanged
- All existing call sites work without any modification

## Usage

```ts
// existing — no change needed, still uses 6767
new FireblocksSigningDriver({ userApiKeys })

// new — opt in to a custom coin type
new FireblocksSigningDriver({ userApiKeys, coinType: 1234 })


mjuchli-da (approved):
Thanks!

PR #2067, opened 2026-06-29
feat: make validator url optional in token namespace

### LINKED ISSUE #1704: wallet-sdk: Extend token namespaces to interact with CIP-56 with only the ledger api
Currently, we need a valid validator url to initialize the token namespace. However, we should be able to interface with CIP-56 only using the ledger api. We should create a variation of the token namespace where we don't need to provide the validator url

rukmini-basu-da:
Closes https://github.com/canton-network/wallet/issues/1704


alexmatson-da (approved):
nice one!

PR #2072, opened 2026-06-30
fix(wallet-sdk): export custom log adapter types

kyken:
## Summary

The Wallet SDK already accepts a custom `logAdapter` when creating SDK instances, but the types needed to construct and type a custom logger are not exposed from the package root.

This PR exports `CustomLogAdapter` and the logger adapter types from `@canton-network/wallet-sdk`, so SDK users can provide their own logging implementation without importing from internal paths.

## Changes

- Export `CustomLogAdapter` from the wallet SDK package root.
- Export logger adapter types used by custom `logAdapter` implementations.
- Extend the package validation script to verify these exports from the packed package.


rukmini-basu-da (approved):
thanks!

PR #2073, opened 2026-06-30
feat: create amulet service with only scan client

### LINKED ISSUE #2035: Drop scan-proxy from AmuletService
`AmuletService` currently requires a ScanProxyClient in its constructor: https://github.com/canton-network/wallet/blob/6a748990dd8e6f7d718d6809750b2a13c7460196/core/amulet-service/src/amulet-service.ts#L26

this should be unnecessary, everything should rely on a `ScanClient`

rukmini-basu-da:
Closes https://github.com/canton-network/wallet/issues/2035


alexmatson-da (approved):
thanks!

Issue #1902, opened 2026-05-28
Don't rely on store interfaces and zod in core-wallet-ui-components


pawelstepien-da:
`core-wallet-ui-components` relies on interfaces from `core-wallet-store`. Besides types it uses network zod for validating form inputs values against shape expected in the store.
We should:
-Keep `core-wallet-ui-components` agnostic to the store layer, no dependency to `core-wallet-store`
-Create zods for user api interfaces. Use them in to validate user api request payloads and frontend forms.
-Use different package(s) as source of types, like rpc api client packages, or re-export through `core-types`.
-(debatable) Either have user api zods in wallet gateway and validate form values one level higher - in wallet gateway frontend components instead of `core-wallet-ui-components`, or put user api zods in shared package (core-types?) and then it can be used in `core-wallet-ui-components` without need for dependency to wallet-gateway.

singhyash05:
@pawelstepien-da I’d like to take this issue. I recently completed and merged #2173.

Before implementation, I’ll map current `core-wallet-ui-components` dependencies and propose where user API Zod schemas should live, then proceed after confirming direction.

Could you assign #1902 to me?

singhyash05:
@alexmatson-da Please can you look into this?

mjuchli-da:
@singhyash05 apologies, just saw this now. I've assigned you. Looking forward to the contribution!

singhyash05:
@pawelstepien-da @mjuchli-da

Thanks for the assign.

Before I start writing code: the ticket leaves Zod placement open. Two options:

A) Keep Zod in wallet-gateway and validate forms there (above `core-wallet-ui-components`).
B) Put user-API Zod in a shared package (`core-types` or similar) so `core-wallet-ui-components` can validate forms without depending on store or gateway.

I'm leaning B, because the network form still lives in `core-wallet-ui-components` after #2077, and A means moving that validation up.

Does B work for you, or do you want A?
I'll map the current store/Zod usages next and follow whatever you pick.

PR #2077, opened 2026-07-02
feat: add service account auth UI, fix existing issues in networks

### LINKED ISSUE #2066: Fix WG Network add and review pages ledgerApi baseUrl and auth errors
UI expects `network.ledgerApi.baseUrl: string` but api returns `network.ledgerApi: string`.
Related to https://github.com/canton-network/wallet/issues/1902

Add network page doesn't render auth inputs, which prevents user from adding network. It should render auth fields for both add and edit, while not exposing secrets.

adminAuth inputs never renders

### LINKED ISSUE #2080: Add serviceAccountAuth inputs in network form

pawelstepien-da:
Closes: https://github.com/canton-network/wallet/pull/2077
Added serviceAccountAuth inputs and handled it in user api. 

Closes: https://github.com/canton-network/wallet/issues/2066
Fixed issues on network page:
-Add network not displaying auth inputs, making it impossible to add an network
-`adminAuth` inputs never displaying
-Existing network ledgerApi not displaying in the input and impossible to edit

Refactored network page with following changes to auth fields:
-Created reusable auth editor component and used it for auth, adminAuth and serviceAccountAuth
-Optional auths now start in readonly mode, you can add, edit it or delete it, or restore state before remove or edit
-Required auth renders inputs always as it was before
-`clientSecret` now won't show in the input. Instead if secret is set show asterisks and info that secret is hidden, otherwise just empty input with required
-Reduce conversions between network shapes used by api and store - rely almost entirely on api shape, with only exception of using zod validation from store, where ledger api shape is changed from `string` to `{ baseUrl: string }`
-Removed UI setting `adminAuth` to default but empty state with only `method`, instead user have control over it existing or not. It will result in errors 'No admin auth configured' when executing action possible only with `adminAuth` instead of more cryptic errors because of lack of other properties, especially that it was impossible to edit `adminAuth` with UI

image
image
image
image

### DIFF
--- a/core/wallet-ui-components/src/components/network-form.ts
+++ b/core/wallet-ui-components/src/components/network-form.ts
@@ -70,6 +70,7 @@ export class NetworkForm extends BaseElement {
     accessor network: NetworkFormData = {
         ledgerApi: '',
         auth: {},
+        serviceAccountAuth: {},
     } as NetworkFormData
 
     @state() private _error = ''
@@ -306,6 +307,7 @@ export class NetworkForm extends BaseElement {
                                         (authObj as SelfSignedAuth)
                                             .clientSecret ?? '',
                                 } satisfies SelfSignedAuth)
+                                // TODO should I enforce client credientials for network.serviceAccountAuth?
                             } else if (select.value === 'client_credentials') {
                                 Object.assign(authObj, {
                                     method: 'client_credentials',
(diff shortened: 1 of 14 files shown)


core/wallet-ui-components/src/components/network-form.ts
```diff
@@ -228,12 +228,15 @@ export class NetworkForm extends BaseElement {
...
+        AuthEditor.styles,
     ]
 
     handleSubmit(e: Event) {
         e.preventDefault()
 
-        const parsedData = networkSchema.safeParse(this.network)
+        const parsedData = networkSchema.safeParse(
+            this.toStoreNetworkForValidation(this.network)
+        )
```
alexmatson-da, line 239: arguably we'd actually want to implement zod schemas directly for the API models rather than the store models... but we can leave that for a future improvement / tech-debt

alexmatson-da (approved):
thanks!

core/wallet-ui-components/src/components/network-form.ts
pawelstepien-da: I totally agree. I think `wallet-ui-components` should be divorced from the store. Here is a tech debt ticket for that https://github.com/canton-network/wallet/issues/1902

PR #2093, opened 2026-07-08
chore: wallet-sdk release notes 1.4.0

rukmini-basu-da:


docs/wallet-integration-guide/src/release-notes/index.rst
```diff
@@ -3,6 +3,96 @@ Wallet SDK Release Notes
...
+
+*For certain functions that require a validator operator party, we have provided a method to retrieve given the validator URL*
+
+.. code-block:: javascript
+
+    import { getValidatorParty } from '@canton-network/wallet-sdk'
+
+    const validatorParty = await getValidatorParty(
+        new URL('http://validatorurl'),
+        {
```
mateuszpiatkowski-da, line 76: Nitpick: I think that, for simplicity, we can gather this into a single const value (like `authObject`) and reuse it so that user can focus on what actually matters, i.e. the validator function itself

mateuszpiatkowski-da (approved):
LGTM, left one nitpick (I'm totally fine if you ignore it though :D)

docs/wallet-integration-guide/src/release-notes/index.rst
rukmini-basu-da: Yeah, that makes sense, I have the authObject in the docs examples but for the release notes I wanted to make the values as explicit as possible.

PR #2096, opened 2026-07-08
fix: addSession verify that token matches network

pawelstepien-da:
In `userApi.addSession` add verification that accessToken claims `iss`, `aud`, `sub`, `scp` and `scope` match their counterparts in auth of the network passed as argument.
Adjust mock-oauth2 hardcoded `scp` and `scope` claims to match values defined in WG config default networks.
In login-callback if addSession fails show error toast.


wallet-gateway/remote/src/web/frontend/callback/login-callback.ts
```diff
@@ -75,9 +78,11 @@ export class LoginCallback extends LitElement {
...
                     tokenResponse.access_token,
                     stateManager.networkId.get() || ''
-                ).then(() => {
-                    redirectToIntendedOrDefault()
-                })
+                )
+                    .then(() => {
+                        redirectToIntendedOrDefault()
+                    })
+                    .catch(handleErrorToast)
```
pawelstepien-da, line 85: Potentially we could also call `stateManager.clearAuthState` and redirect to login, but that would clear the error toast from HTML before user can see it, additional workaround would be needed. Open for discussion how to handle it UX-wise.
That situation happens when there are issues with accessToken that make addSession fail. Before it just stayed at this page forever displaying "Logged in!", with my change it at least indicates that there was some error.

rukmini-basu-da (approved):

PR #2111, opened 2026-07-09
fix: dfns execute tx

pawelstepien-da:
This aligns DFNS transaction execution with other signing providers and removes obsolete comments and updates signing providers docs to reflect current state.


pawelstepien-da:
The comments regarding Dfns auto-signing / broadcasting txes to canton on sign were obsolete. It was initial idea of creator, but final version ended up with regular raw signing, like other signing providers.

mjuchli-da (approved):

PR #2122, opened 2026-07-11
feat(wallet-sdk,core-token-standard): add codegen, create ts interfaces for commands

### LINKED ISSUE #2107: [Ref Token] Add Test Token contracts' interfaces to wallet-sdk

mateuszpiatkowski-da:
Closes #2107

### DIFF
--- a/core/token-standard/src/types.ts
+++ b/core/token-standard/src/types.ts
@@ -3,9 +3,6 @@
 
 import { Splice } from '@daml.js/token-standard-models-1.0.0'
 import { PartyId } from '@canton-network/core-types'
-import * as testTokenCodegen from '@daml.js/test-token-v1'
-
-export const TestTokenV1 = testTokenCodegen.Splice.Testing.Tokens.TestTokenV1
 
 export * from './interface-ids.const.js'
 
--- a/core/token-standard/src/types.ts
+++ b/core/token-standard/src/types.ts
@@ -3,6 +3,9 @@
 
 import { Splice } from '@daml.js/token-standard-models-1.0.0'
 import { PartyId } from '@canton-network/core-types'
+import * as testTokenCodegen from '@daml.js/test-token-v1'
+
+export const TestTokenV1 = testTokenCodegen.Splice.Testing.Tokens.TestTokenV1
 
 export * from './interface-ids.const.js'
 
@@ -23,13 +26,11 @@ export type {
 
 export type {
     Transfer,
-    TransferInstruction,
     TransferInstructionView,
     TransferInstruction_Accept,
     TransferInstruction_Reject,
     TransferInstruction_Withdraw,
     TransferInstruction_Update,
-    TransferFactory,
     TransferFactoryView,
     TransferFactory_PublicFetch,
     TransferFactory_Transfer,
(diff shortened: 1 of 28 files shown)


scripts/src/lib/version-config.json
```diff
@@ -7,9 +7,9 @@
                 "hash": "3debf0d5f3a79ea5e29ab682e31428f9f853a5236d56bf9bd1d3af45673a6ca7"
             },
             "splice": {
-                "version": "0.6.1",
+                "version": "0.6.12",
```
mateuszpiatkowski-da, line 10: We can't access specific dar file  from .localnet without the upgrade

sdk/wallet-sdk/src/plugins/testToken.ts
```diff
@@ -0,0 +1,118 @@
...
+    TransferFactory,
+    TransferInstruction,
+    TestTokenV1,
+    AllocationSpecification,
+} from '@canton-network/core-token-standard'
+import { SDKContext, SDKPlugin } from '../wallet'
+import { WrappedCommand } from '@/wallet/namespace/ledger'
+import { PartyId } from '@canton-network/core-types'
+
+export default class WalletSDKTestTokenPlugin extends SDKPlugin {
```
mateuszpiatkowski-da: I'm still torn whether this should be a plugin. Practically it doesn't use `ctx` except for the `super()` call, but in theory we might wanna use it later (idk for what purpose however)...

I didn't want to add test-token related stuff directly inside wallet-sdk as I imagine it's not necessary for each instance to contain this piece of code and now it's just possible to extend it on demand.

core/token-standard/src/types.ts
```diff
@@ -3,6 +3,9 @@
 
 import { Splice } from '@daml.js/token-standard-models-1.0.0'
 import { PartyId } from '@canton-network/core-types'
+import * as testTokenCodegen from '@daml.js/test-token-v1'
+
+export const TestTokenV1 = testTokenCodegen.Splice.Testing.Tokens.TestTokenV1
```
mjuchli-da: I wonder if we shouldn't just create a separate core package that contains the test-token codegen and possibly other helpers. The package does not necessarily need to be published to NPM but it could.

core/token-standard/tsconfig.json
```diff
@@ -23,6 +23,9 @@
             ],
             "@daml.js/daml-stdlib-DA-Time-Types-1.0.0/*": [
                 "../../damljs/token-standard-models/daml-stdlib-DA-Time-Types-1.0.0/*"
+            ],
+            "@daml.js/test-token-v1": [
```
mjuchli-da: See comment above.

scripts/src/generate-test-token.ts
```diff
@@ -0,0 +1,33 @@
...
+import * as path from 'path'
+import { getRepoRoot } from './lib/utils.js'
+import { installDPM } from './install-dpm.js'
+import { runDamlCodegen } from './lib/daml-codegen.js'
+import { existsSync } from 'fs'
+import { execSync } from 'child_process'
+
+const repoRoot = getRepoRoot()
+
+async function main() {
```
mjuchli-da, line 23: Add a comprehensive comment to the method that describes that localnet is fetch and test-token is part of it; then codgen is being generated and is to be exported for further use.

sdk/wallet-sdk/src/plugins/testToken.ts
mjuchli-da: I think the `plugin` folder makes things a bit more convoluted.
The fact that this is in our repository makes it really a first-class citizen and not a plugin.
I hat the same intuition and would move it under a `test-token` namespace. 
The namespace can be lazy loaded, so that solves the "...I imagine it's not necessary for each instance..."

mjuchli-da (commented):
Thank you. I would refrain from defining a plugin and instead create a first class citizen (namespace). In addition, I would consider creating a separate core module for the test token.

mateuszpiatkowski-da:
Split into other PRs

PR #2123, opened 2026-07-13
feat: add securosys signing driver

MikolajSzargut:
# @canton-network/core-signing-securosys

Signing driver for integrating the Canton Wallet Gateway with Securosys TSB.

## Features

- `createKey` creates a TSB SKA key with a hardcoded empty policy.
- `getKeys` enumerates TSB keys and returns Wallet Gateway-compatible public
  keys.
- `signTransaction` creates a TSB sign request and returns the TSB request ID as
  the provider `txId`.
- `getTransaction` maps TSB request status/result into Wallet Gateway
  transaction status/signature fields.
- `getTransactions` fetches by provider transaction IDs. Public-key-only
  filtering is supported from this driver's in-memory transaction cache.
- Runtime configuration can be inspected and changed through
  `getConfiguration` / `setConfiguration`.

### DIFF
--- a/core/signing-securosys/src/signing-api-sdk.ts
+++ b/core/signing-securosys/src/signing-api-sdk.ts
@@ -1,0 +1,819 @@
+// SPDX-FileCopyrightText: Copyright 2026 Securosys SA
+// SPDX-License-Identifier: Apache-2.0
+
+import { readFileSync } from 'node:fs'
+import type {
+    CreateKeyParams,
+    GetTransactionParams,
+    GetTransactionsParams,
+    Key,
+    KeyIdentifier,
+    SignTransactionParams,
+    SigningStatus,
+    Transaction,
+} from '@canton-network/core-signing-lib'
+import { Agent, type Dispatcher } from 'undici'
+
+export type TsbRequestStatus =
+    | 'PENDING'
+    | 'APPROVED'
+    | 'EXECUTED'
+    | 'FAILED'
+    | 'EXPIRED'
+    | 'REJECTED'
+    | 'CANCELLED'
+
+export type TsbSignatureAlgorithm =
+    | 'EDDSA'
+    | 'SHA256_WITH_ECDSA'
+    | 'SHA512_WITH_ECDSA'
+    | 'NONE_WITH_EC_SCHNORR_BIP0340'
+    | string
+
+export interface TsbCreateKeyAttributes {
+    encrypt?: boolean
+    decrypt?: boolean
+    verify?: boolean
+    sign?: boolean
+    wrap?: boolean
+    unwrap?: boolean
+    derive?: boolean
+    bip32?: boolean
+    slip10?: boolean
+    extractable?: boolean
+    modifiable?: boolean
+    destroyable?: boolean
+    sensitive?: boolean
+    copyable?: boolean
+    rollover?: boolean
+    [key: string]: unknown
+}
+
+export interface TsbCreateKeyRequest {
+    label: string
+    password?: string
+    id?: string
+    algorithm?: string
+    algorithmOid?: string
+    curveOid?: string
+    keySize?: number
+    attributes: TsbCreateKeyAttributes
+    policy: TsbCreateKeyPolicy
+    [key: string]: unknown
+}
+
+export interface TsbCreateKeyPolicy {
+    ruleUse: null
+    ruleBlock: null
+    ruleUnblock: null
+    ruleModify: null
+    keyStatus: {
+        blocked: false
+    }
+}
+
+export interface TsbKeyAttributes {
+    label: string
+    id?: string
+    uuid?: string
+    algorithm?: string
+    algorithmOid?: string
+    curveOid?: string
+    publicKey?: string
+    [key: string]: unknown
+}
+
+export interface TsbSignedKeyAttributes {
+    xml?: string
+    json: TsbKeyAttributes
+    xmlSignature?: string
+    attestationKeyName?: string
+}
+
+export interface TsbSignRequest {
+    payload: string
+    payloadType?: string
+    signKeyName: string
+    keyPassword?: string
+    metaData?: string
+    metaDataSignature?: string
+    signatureAlgorithm: TsbSignatureAlgorithm
+    signatureType?: string
+    context?: string
+    auxiliaryRandomData?: string
+    taprootTweakData?: string
+    merkleRootData?: string
+    [key: string]: unknown
+}
+
+export interface TsbSignedSignRequest {
+    signRequest: TsbSignRequest
+}
+
+export interface TsbSignRequestResponse {
+    signRequestId: string
+}
+
+export interface TsbRequestStatusResponse {
+    id: string
+    status: TsbRequestStatus
+    executionTime?: string
+    approvedBy?: string[]
+    notYetApprovedBy?: string[]
+    rejectedBy?: string[]
+    result?: string | null
+    inputOfflineHsm?: unknown
+    [key: string]: unknown
+}
+
+export interface TsbRequestByStatus {
+    id: string
+    status: TsbRequestStatus
+}
+
+export interface TsbRequestIdsByStatusResponse {
+    requests: TsbRequestByStatus[]
+}
+
+export interface SecurosysTSBClientConfig {
+    baseUrl: string
+    keyManagementApiKey?: string
+    keyOperationApiKey?: string
+    bearerToken?: string
+    mtlsP12Path?: string
+    mtlsP12Password?: string
+    keyPassword?: string
+    signatureAlgorithm?: TsbSignatureAlgorithm
+}
+
+type ApiRole = 'key-management' | 'key-operation'
+
+const WALLET_KEY_ALGORITHM = 'ED'
+const WALLET_KEY_CURVE_OID = '1.3.101.112'
+const WALLET_CREATE_KEY_ATTRIBUTES: TsbCreateKeyAttributes = {
+    decrypt: false,
+    sign: true,
+    verify: true,
+    unwrap: false,
+    extractable: false,
+    modifiable: true,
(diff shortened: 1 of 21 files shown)


wallet-gateway/remote/src/init.ts
```diff
@@ -301,6 +312,24 @@ export async function initialize(opts: CliOptions, logger: Logger) {
             apiKey: Env.BLOCKDAEMON_API_KEY(''),
             caip2: Env.BLOCKDAEMON_CAIP2('canton:testnet') as CantonCaip2,
         }),
+        [SigningProvider.SECUROSYS]: new SecurosysSigningProvider({
+            baseUrl: Env.SECUROSYS_TSB_BASE_URL('http://localhost:8080'),
```
mjuchli-da: I'd prefer not to set a default url, and instead register the driver only if at least the URL is set. (See DFNS example below).

core/signing-securosys/src/signing-api-sdk.ts
```diff
@@ -0,0 +1,806 @@
...
+            cached?.metadata &&
+            typeof cached.metadata === 'object' &&
+            !Array.isArray(cached.metadata)
+                ? (cached.metadata as Record<string, unknown>)
+                : undefined
+        const signatureAlgorithm =
+            typeof cachedMetadata?.signatureAlgorithm === 'string'
+                ? cachedMetadata.signatureAlgorithm
+                : this.signatureAlgorithm
+        const signature =
```
mjuchli-da: Possibly empty. However, possible status `signed` expects a signature.
Needs another check and in said case switch status back to `pending` (or possibly `cancelled`?).

mjuchli-da (commented):
Thank you for the contribution! The PR looks very solid, and the most fundamental parts were provided: 
- Driver
- Allocator extension
- TransactionService extension
- Tests

There are a few CI blockers and a couple of robustness items before merge. In addition, my manual testing has resulted in an error when attempting to sign a transaction.

## Blockers
1. **CI**
   - `test-static`: Prettier failures under `core/signing-securosys/`
   - `test-unit`: branch coverage 75.38% < 80% threshold for `@canton-network/core-signing-securosys`
2. **Rebase / build tooling** — package uses `tsup` + `tsup.base`; `main` has moved to `tsdown`. Please rebase and align with other `core/signing-*` packages.
3. **`EXECUTED` without signature** — `mapTsbStatus` maps `EXECUTED` → `signed`, but signature is only populated when `response.result` is present. Returning signed-without-signature hard-fails in the gateway. Fall back to `pending` (or error) until a usable signature exists.


## Manual user testing

I've tested the signing driver with the sent credentials.

**Allocate a party/wallet** ✅
This worked perfectly by using the asynchronous workflow:
1. `createKey`
2. `getTransaction`
3. allocate with received signature

**Sign a transaction** ❌
The creation of a `Ping` contract failed at the signing step:
```
"Error from signing driver: TSB API call to /v1/key/attributes failed (500): {\"errorCode\":701,\"reason\":\"res.error.in.hsm\",\"message\":\"HSM error: status: PKCS#11: KEY_FUNCTION_NOT_PERMITTED\"}",
```

wallet-gateway/remote/src/init.ts
MikolajSzargut: Fixed:
1. Env no longer takes a fallback/default: /wallet-gateway/remote/src/env.ts:
`static SECUROSYS_TSB_BASE_URL = () => Env.get('SECUROSYS_TSB_BASE_URL')`
2. Driver registration is conditional on securosysBaseUrl: wallet-gateway/remote/src/init.ts:
```
const securosysBaseUrl = Env.SECUROSYS_TSB_BASE_URL()

if (securosysBaseUrl) {
    drivers[SigningProvider.SECUROSYS] = new SecurosysSigningProvider({
        baseUrl: securosysBaseUrl,
        ...
    })
}
```

core/signing-securosys/src/signing-api-sdk.ts
MikolajSzargut: Fixed:
Resolved in core/signing-securosys/src/signing-api-sdk.ts:
```
const rawSignature =
    typeof response.result === 'string' && response.result.trim()
        ? response.result
        : undefined

const signature = rawSignature
    ? normalizeSignature(rawSignature, signatureAlgorithm)
    : undefined

const mappedStatus = mapTsbStatus(response.status)
const status =
    mappedStatus === 'signed' && !signature ? 'pending' : mappedStatus
```
So EXECUTED maps to signed only when a non-empty usable signature exists. Otherwise it is downgraded to pending.

MikolajSzargut:
I fixed errors on pipeline related with securosys signing driver - prettier env.ts file and add node to tsdown config. Other errors looks like there is some issues with playwright.

core/signing-securosys/src/signing-api-sdk.ts
```diff
@@ -0,0 +1,911 @@
...
+            if (isSkippableKeyAttributesError(result.reason)) {
+                return []
+            }
+            throw result.reason
+        })
+    }
+
+    public async createKey(params: CreateKeyParams): Promise<Key> {
+        const request = compact({
+            label: params.name,
```
pawelstepien-da: Party hint is used as label here. Key label has to be unique, while party hint is not. 
For example on a single Wallet Gateway instance 2 different users connected to 2 different networks could try to create a party with same hint, and the one who does it later gets error from Securesys API:
```{\"errorCode\":608,\"reason\":\"res.error.key.already.existing\",\"message\":\"Could not create key. The key name is already in use.\"}"```

I see 2 possible approaches:
-Make key labels something unique. This would probably be problematic when using `signTransaction` with `keyIdentifier.id` and warrant using only `keyIdentifier.publicKey` and I see from the code that acquiring key label by publicKey from API is not straightforward. 
-Accept it as driver's limitation and document it in README.md. I confirmed with @mjuchli-da that this is also fine.

File: core/signing-securosys/src/signing-api-sdk.ts [hunk 4]
```diff
@@ -0,0 +1,911 @@
...
+            signature,
+            publicKey: cached?.publicKey,
+            metadata: {
+                tsbStatus: response.status,
+                executionTime: response.executionTime,
+                approvedBy: response.approvedBy,
+                notYetApprovedBy: response.notYetApprovedBy,
+                rejectedBy: response.rejectedBy,
+                inputOfflineHsm: response.inputOfflineHsm,
+                ...(cached?.metadata ?? {}),
```
pawelstepien-da: Cached metadata is spread last, so it it could overwrite metadata properties from response.

core/signing-securosys/src/signing-api-sdk.ts
```diff
@@ -0,0 +1,911 @@
...
+
+        this.transactionCache.set(transaction.txId, transaction)
+        return transaction
+    }
+
+    private async resolveKeyIdentifier(
+        keyIdentifier: KeyIdentifier
+    ): Promise<Key> {
+        if (keyIdentifier.id && keyIdentifier.publicKey) {
+            return this.cacheKey({
```
pawelstepien-da: The cached key doesn't come from Securosys API response, but from arbitrary parameters. This could  potentially cache wrong data. I suggest first checking if entry matching `keyIdentifier` is already in cache, if yes return it, in not use `getKeyAttributes` and validate that response matches `keyIdentifier` before caching.

mjuchli-da (commented):
Congrats!, I verified that the current state works.

I'll leave it to @pawelstepien-da to point out to a few minor ones.

pawelstepien-da (commented):
Looks very good. Leaving 3 minor comments. 
Fix for playwright errors you saw in CI is already merged to main.

core/signing-securosys/src/signing-api-sdk.ts
MikolajSzargut: Good catch, agreed. I changed the merge order so cached metadata is applied first and fresh TSB response metadata is applied afterwards. That means fields coming from the latest TSB status response, such as `tsbStatus`, `executionTime`, `approvedBy`, `notYetApprovedBy`, `rejectedBy`, and `inputOfflineHsm`, cannot be overwritten by stale cached metadata.
I also added a regression test covering this case.

core/signing-securosys/src/signing-api-sdk.ts
MikolajSzargut: Agreed. The previous shortcut was too trusting because it cached a key constructed from caller-provided `keyIdentifier.id` and `keyIdentifier.publicKey`.
The flow is now:

1. First check whether a cached key matches the supplied keyIdentifier.
2. If it matches, return it.
3. If not, call `getKeyAttributes` using the key id.
4. Validate that the TSB response matches the requested id/public key.
5. Cache the key only after that validation succeeds.

For public-key-only lookup, the driver still has to enumerate TSB labels first because TSB attributes lookup requires a key label. That path only caches keys returned by TSB.
Added tests for cache reuse, mismatched id/public key rejection, and metadata precedence.

core/signing-securosys/src/signing-api-sdk.ts
MikolajSzargut: Thanks, good catch. I changed this so the party hint is no longer used as the TSB key label.
The driver now creates the key with a temporary unique label (`wallet-{uuid}`), because TSB requires a label at creation time and the public key is not known yet. After TSB returns the generated public key, the driver immediately renames the key through `/v1/key/changeAttributes` to a deterministic label derived from the normalized public key.
The Wallet Gateway signing path now derives the Securosys key id from wallet.publicKey instead of wallet.hint, so two users/networks reusing the same party hint will not collide in TSB anymore. I also documented this in the Securosys README and added tests for the create-and-rename flow.

pawelstepien-da (approved):

mjuchli-da (approved):
Thank you!

PR #2129, opened 2026-07-13
fix: transaction history and parser fixes

fayi-da:


core/tx-parser/src/test-data/expected/txs.json
```diff
@@ -48,7 +48,10 @@
                 "lockedHoldingsChangeSummary": {
                     "amountChange": "0",
                     "inputAmount": "0",
-                    "instrumentId": { "admin": "", "id": "" },
+                    "instrumentId": {
+                        "admin": "",
+                        "id": ""
+                    },
```
alexmatson-da, line 54: huh, wonder why this happened

alexmatson-da (approved):
lgtm

rukmini-basu-da (approved):

core/tx-parser/src/test-data/expected/txs.json
fayi-da: prettier i believe. Maybe have been my vim using a different prettier version. I'll check

PR #2137, opened 2026-07-14
chore: add community-contributed shell.nix

jaspervdj-da:
Some of the other folks in SDK were working on this repository recently and complained they were unfamiliar with setting up the TS tooling (`yarn`, `nx`, etc.).  This provides a simple `shell.nix` to make drive-by contributions a bit easier.

### DIFF
--- a/docs/development/shell.nix
+++ b/docs/development/shell.nix
@@ -9,19 +9,11 @@ let
   java = pkgs.openjdk;
 in
 pkgs.mkShell rec {
-  # Some of the scripting uses tsup, but we rely on nix instead to provision
-  # this, so replace it by a stub.
-  tsup = pkgs.writeShellScriptBin "tsup" ''
-    #!/usr/bin/env bash
-    exec true
-  '';
-
   buildInputs = [
     java
     pkgs.corepack # This provides yarn.
     pkgs.nodejs_24
     pkgs.typescript-language-server
-    tsup
   ];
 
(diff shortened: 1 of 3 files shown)


Viktor-Kalashnykov-da (dismissed):
LGTM!

mjuchli-da (changes requested):
Thank you for the suggestion. We currently don't see the need to make use nix, and more importantly maintain this file. External contributors so far were able to work against the current setup. However, we do acknowledge that some contributors may want to use nix. For that reason, my current stance is that we would be open to have the contents of the file in a readme.

I'll put it to a vote team-internally again and will report here if things changed.

docs/development/shell.nix
```diff
@@ -0,0 +1,30 @@
...
+    #!/usr/bin/env bash
+    exec true
+  '';
+
+  buildInputs = [
+    java
+    pkgs.corepack # This provides yarn.
+    pkgs.nodejs_24
+    pkgs.typescript-language-server
+    tsup
```
alexmatson-da: why not rely on the tsup installed by yarn?

jaspervdj-da: Good catch.  I'm not sure, I added this back in December or so, maybe we were invoking `tsup` outside of yarn back then.  I tested again without it and it worked, so I removed it.

alexmatson-da (approved):
nice!

mjuchli-da (approved):

PR #2143, opened 2026-07-14
chore: add shared eslint config package

### LINKED ISSUE #374: Improve eslint setup
Currently we have a root eslint config that applies to every directory, but this can give different results if running eslint at an individual workspace level, since it does not resolve the same config. i.e. compare 

- `yarn eslint` (from the root), vs,
- `yarn workspaces foreach -A run eslint`

We should have a more consistent setup here by putting the shared eslint config in a package and installing it to all workspaces

singhyash05:
Move root eslint setup into core/eslint-config and wire workspaces so root and per-workspace lint resolve the same config. Fixes #374.

## Summary

Fixes inconsistent ESLint resolution described in #374: root `yarn eslint` and per-workspace `yarn workspaces foreach -A run eslint` were not always loading the same config.

Adds a private shared package `@canton-network/core-eslint-config` at `core/eslint-config` that owns the monorepo flat config (`createConfig({ rootDir })`). The root `eslint.config.mjs` and workspace configs become thin wrappers that import it, so root and workspace lint resolve one source of truth.

Wired and thinned:
- Root package + lockfile
- `core/wallet-dapp-rpc-client`, `core/wallet-dapp-remote-rpc-client`, `core/wallet-user-rpc-client` (shared config + local overrides: ignore `docs/**`, header rule off, `any` / unused as warn on `src`)
- Examples `ping`, `portfolio`, `walletconnect` (shared config + React where needed; header format off; walletconnect gets a new `eslint.config.js` so Nx/CI lint picks it up)

Also aligns a few shared-config details that were missing from the old root-only story: include `*.tsx` in files globs and set `tsconfigRootDir` from the caller’s `rootDir`.

Closes #374

## Test plan

- [x] `yarn lint` (Nx CI path) — all projects green
- [x] `yarn build:all` — green
- [x] `yarn playwright install` then `yarn test:all` — green

### DIFF
--- a/core/eslint-config/index.mjs
+++ b/core/eslint-config/index.mjs
@@ -1,0 +1,80 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import js from '@eslint/js'
+import globals from 'globals'
+import tseslint from 'typescript-eslint'
+import { defineConfig } from 'eslint/config'
+import { includeIgnoreFile } from '@eslint/compat'
+import { join } from 'node:path'
+import headers from 'eslint-plugin-headers'
+import nxeslint from '@nx/eslint-plugin'
+
+/**
+ * Shared ESLint flat config for the monorepo.
+ * @param {{ rootDir: string }} options
+ */
+export function createConfig({ rootDir }) {
+    const gitignorePath = join(rootDir, '.gitignore')
+    const headerFilePath = join(rootDir, 'header.txt')
+
+    return defineConfig([
+        includeIgnoreFile(gitignorePath),
+        {
+            ignores: [
+                '**/dist',
+                '**/build',
+                '**/_proto',
+                '**/.venv',
+                '**/vite-env.d.ts',
+                '.yarn/**',
+                '.commitlintrc.js',
+                '.pnp.*',
+                'core/wallet-dapp-rpc-client',
+                'core/wallet-dapp-remote-rpc-client',
+                'core/wallet-user-rpc-client',
+                'core/ledger-client/src/generated-clients',
+                'core/ledger-client-types/src/generated-clients',
+                'damljs/**',
+                'docs/wallet-integration-guide/examples/**',
+                'examples/ping/playwright-report/**',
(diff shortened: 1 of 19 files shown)


core/wallet-dapp-remote-rpc-client/eslint.config.mjs
```diff
@@ -1,67 +1,29 @@
-import eslint from '@eslint/js'
```
alexmatson-da, line 1: unfortunately these changes won't stick because the `core/wallet-*-rpc-client` packages are generated wholesale from code. The generator code lives in `core/rpc-generator` and is called for each OpenRPC spec through the root-level yarn task: `yarn generate:all` 

You'll probably see these changes revert if you regenerate the clients, therefore, I suggest

- add a `_eslint.config.mjs` file under `core/rpc-generator/templates/client/typescript` (you should also see an analogous `_package.json` in that dir) with the changes we want
- modify the [`afterCopyStatic` hook](https://github.com/canton-network/wallet/blob/7c2cce365d4c3085deb677113b9e645f2101aac8/core/rpc-generator/src/components/client.ts#L106) in the generator to overwrite the default eslint config with the modified one (`_eslint.config.mjs`)
- re-run `yarn generate:all` to validate the resulting packages include the same config 

LMK if you have any questions or run into problems

alexmatson-da: you'll also want to add `"@canton-network/core-eslint-config": "workspace:^",` to the devDependencies of that `_package.json` too

singhyash05: Thanks, that makes sense. I’ll move the RPC client eslint/package changes into the rpc-generator TypeScript client template, update the afterCopyStatic hook to emit eslint.config.mjs, then run yarn generate:all and push the regenerated output.

singhyash05:
Pushed updated with template, afterCopyStatic and generate:all verified.

core/wallet-dapp-remote-rpc-client/eslint.config.mjs
singhyash05: Pushed updated code @alexmatson-da

alexmatson-da (approved):
thanks!

PR #2150, opened 2026-07-15
fix: improve auth callback page

alexmatson-da:
small fix to make the callback page UI fit better with the overall app. The callback usually flashes only for a moment before it redirects you, but you can still catch a glimpse of the completely unstyled `"<h2>Logged in!</h2>"` for a second. 

Also the page could get stuck in the event of an error calling `addSession`, so the page would look a bit nicer in that case too

https://github.com/user-attachments/assets/5516d69b-c8e1-4e00-a6d1-e345f661f1e5


rukmini-basu-da (approved):
thanks!

PR #2161, opened 2026-07-16
feat: add multi-session support

alexmatson-da:


core/wallet-store-sql/src/migrations/015-add-origin-field-session.ts
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { Kysely } from 'kysely'
+import { DB } from '../schema.js'
+
+export async function up(db: Kysely<DB>): Promise<void> {
+    console.log('Adding origin column to networks table')
+
```
mjuchli-da, line 10: Session currently doesn't have a primary key. We should consider making the `id` the pk, as well as a unique constraint around `origin, user_id`. Then there's also the question of the `access_token` where i'm not sure if it's a good idea to make this unique, although it might very well be valid.

mjuchli-da (approved):
Great work @alexmatson-da !

PR #2163, opened 2026-07-17
feat: add BitGo TSS MPC signing driver

ravibitgo:
**Summary**

Adds a new signing driver (@canton-network/core-signing-bitgo) that integrates BitGo's TSS MPC custodial wallets as a Canton signing provider.

  - Key model: A BitGo custodial wallet maps 1:1 to a Canton key. walletId is used as the stable key identifier (publicKey) throughout — wallet creation is the createKey flow, and the walletId is passed as keyIdentifier.id on every signing request.
  - Async signing: signTransaction submits a message signing request (POST /msgrequests) and returns immediately with status: pending. The gateway polls getTransaction until messages[0].state === 'signed', at which point the Ed25519 signature and Canton signer fingerprint are extracted from the signed txHash (a hex-encoded JSON blob).
  - Restart resilience: An in-memory txRequestId → walletId map is populated at sign time for fast lookups. On process restart, the enterprise txrequests endpoint (GET /api/v2/enterprise/{id}/txrequests) serves as a fallback — no Redis required, consistent with other drivers.

**Changes**

  - core/signing-bitgo/ — new package: BitGoHandler (API client) + BitGoSigningDriver (implements SigningDriverInterface), 61 tests, test scripts, README
  - core/signing-lib/src/config/schema.ts — adds BITGO = 'bitgo' to SigningProvider enum
  - wallet-gateway/remote/src/env.ts + init.ts — wires up BitGo env vars (BITGO_ACCESS_TOKEN, BITGO_API_URL, BITGO_ENTERPRISE_ID, BITGO_COIN) and registers the driver

**Design notes for reviewers**

  - publicKey === walletId is intentional. Unlike Fireblocks (which returns a real Ed25519 public key), BitGo's MPC model doesn't expose a single public key upfront. The walletId is the stable identifier used to route signing requests.
  - apiVersion=full and latest=true are both required on txrequest fetches. Without full, messages[0].state is absent; without latest, you get a stale version of the txRequest.
  - Message-level state takes precedence: Canton signing is complete when messages[0].state === 'signed', even if the txRequest itself is still in pendingDelivery. This was discovered empirically during integration testing.
  
**Known limitations**
  
  - getTransactions with publicKeys fetches only the first page of txrequests per wallet (no cursor/pagination yet). Fine for current usage; can be followed up.
  
**Test plan**
  
  - [ ] yarn test in core/signing-bitgo — 61 tests pass
  - [ ] yarn build in core/signing-bitgo
  - [ ] Manual connectivity: BITGO_ACCESS_TOKEN=<token> BITGO_ENTERPRISE_ID=<id> npx tsx scripts/test-connectivity.ts
  - [ ] Manual sign round-trip: npx tsx scripts/test-sign.ts (submit mode, then check mode)


core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,302 @@
...
+        const messageState = txReq.messages?.[0]?.state
+        const mappedStatus: SigningStatus =
+            messageState === 'signed'
+                ? 'signed'
+                : (BITGO_STATE_TO_CANTON[txReq.state] ?? 'pending')
+        // Only extract when signed status comes from message-level state (txRequest still in
+        // pendingDelivery). When the txRequest itself is terminal (delivered/signed), trust it.
+        const signedFromMessage =
+            messageState === 'signed' && mappedStatus === 'signed'
+        const signedData = signedFromMessage
```
mjuchli-da: Potential bug:
- txReq.state is delivered / signed → mappedStatus = 'signed' 
- But messageState !== 'signed' or txHash missing → signedFromMessage = false → signedData = {}
- Pending fallback only runs when signedFromMessage is true
- Result: { status: 'signed' } with no signature

Fix 
Whenever mappedStatus === 'signed', extract signature; if extraction fails, return pending (which keeps polling).

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,302 @@
...
+        txHash: string
+        walletId: string
+        /** Override type detection — pass 'CANTON_SIGN_TRANSACTION' or 'CANTON_SIGN_TOPOLOGY' directly to skip the sdk-coin-canton import. */
+        messageStandardType?: string
+    }): Promise<{ txId: string }> {
+        let messageStandardType = params.messageStandardType
+        if (!messageStandardType) {
+            const { detectCantonSigningPayloadType } =
+                await import('@bitgo/sdk-coin-canton')
+            messageStandardType = detectCantonSigningPayloadType(params.tx)
```
mjuchli-da: Any chance this detection can be local to avoid pulling the full BitGo Canton SDK into the WG dependency graph?

core/signing-bitgo/src/index.ts
```diff
@@ -0,0 +1,232 @@
...
+                params: SetConfigurationParams
+            ): Promise<SetConfigurationResult> => {
+                const validated = BitGoConfigUpdateSchema.safeParse(params)
+                if (!validated.success) {
+                    return {
+                        error: 'bad_arguments',
+                        error_description: validated.error.message,
+                    }
+                }
+                this.config = { ...this.config, ...validated.data }
```
mjuchli-da: Potentially strip undefined keys before merge.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,302 @@
...
+                type: 'custodial',
+                multisigType: 'tss',
+                ...(this.enterpriseId && { enterprise: this.enterpriseId }),
+            }
+        )
+        // walletId is the stable key identifier for Canton — used in signTransaction and getTransaction.
+        return {
+            id: wallet.id,
+            name: wallet.label,
+            publicKey: wallet.id,
```
mjuchli-da: `publicKey === walletId` won’t work for Canton party allocation. The signing provider should return a real Ed25519 pubkey from the API; the party allocator will fingerprint that and passes it to `generateTopologyTransactions`.

mjuchli-da (commented):
Driver package looks solid; end-to-end BitGo support is incomplete. Please address the blockers below.

## Blockers

1. **Gateway never uses the driver** 
`init.ts` registers `SigningProvider.BITGO`, but `TransactionService.sign()` / `execute()` and WalletAllocationService have no BITGO cases; they throw "Unsupported signing provider".
TODO: mirror Dfns implementation and create a `BitGoWalletAllocator`, `signWithBitgo`, switch cases, and tests.

2. **publicKey === walletId won’t work for Canton party allocation**
Other signing providers return a real Ed25519 pubkey from the custody API; the allocator fingerprints that and passes it to `generateTopologyTransactions`.
BitGo returns publicKey: wallet.id. A future allocator (still needed) cannot use `walletId` alone as topology/fingerprint need key material.
Keep id = walletId for routing; resolve the real Ed25519 pubkey from BitGo keychains and return that as publicKey (same split as Dfns: id + publicKey).

3. **formatTxRequest can return signed without a signature**
When `txReq.state` is terminal (delivered/signed) but messages[0].txHash is missing/unparseable, status stays `signed` with **no signature**. 
The pending fallback only covers message-level signed. Gateway treats signed-without-signature as fatal.

**Also**
- setConfiguration should not spread undefined optional fields over required config.
- @bitgo/sdk-coin-canton is a heavy dep for detectCantonSigningPayloadType alone

## Details on driver is registered but not used

`TransactionService.sign()` and `TransactionService.execute()` (transaction-service.ts) have no BITGO case, resulting in `Unsupported signing provider`; the same goes for wallet creation `WalletAllocationService.createWallet()` in wallet-allocation-service.ts.

Even with env vars set, signingProviderId: "bitgo" fails at create/sign/execute. Registration alone is not enough, every external provider has a dedicated path (see [signWithDfns](https://github.com/canton-network/wallet/blob/main/wallet-gateway/remote/src/ledger/transaction-service.ts#L586), [DfnsWalletAllocator](https://github.com/canton-network/wallet/blob/main/wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/dfns-wallet-allocator.ts), [allocators folder](https://github.com/canton-network/wallet/tree/main/wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers)).

- bitgo-wallet-allocator.ts (copy Dfns pattern)
- case SigningProvider.BITGO in allocation + transaction services
- Tests alongside [transaction-service.test.ts](https://github.com/canton-network/wallet/blob/main/wallet-gateway/remote/src/ledger/transaction-service.test.ts) / wallet-allocation tests

core/signing-bitgo/src/bitgo.ts
ravibitgo: fixed

core/signing-bitgo/src/index.ts
ravibitgo: handled

core/signing-bitgo/src/bitgo.ts
ravibitgo: yes missed this, fixed now

core/signing-bitgo/src/bitgo.ts
ravibitgo: Done - @bitgo/sdk-coin-canton is no longer a dependency of this package at all. Detection is now local: topology payloads are base64(JSON array) (constructed by the gateway allocator), transaction payloads are proto binary - a simple JSON-parse check distinguishes them reliably. The call site also passes messageStandardType: 'CANTON_SIGN_TOPOLOGY' explicitly so detection only runs as a fallback.

The one net-new @bitgo package added to the WG graph is `@bitgo/sdk-lib-mpc`, which is needed for a different purpose: deriving the real Ed25519 public key from the BitGo TSS commonKeychain at m/0. That's unavoidable - it's the only way to get a usable public key for Canton party allocation.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,302 @@
...
+            messageState === 'signed' && mappedStatus === 'signed'
+        const signedData = signedFromMessage
+            ? this.extractSignedData(txReq.messages?.[0]?.txHash)
+            : {}
+        // Fall back to pending if message claimed signed but signature extraction failed —
+        // prevents signed-without-signature from reaching the gateway (hard error vs. retry).
+        const status: SigningStatus =
+            signedFromMessage && signedData.signature === undefined
+                ? 'pending'
+                : mappedStatus
```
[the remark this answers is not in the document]
ravibitgo: handled

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+                ? isTerminalState
+                    ? 'failed'
+                    : 'pending'
+                : mappedStatus
+        const { signature, signedBy } = status === 'signed' ? signedData : {}
+        return {
+            txId,
+            status,
+            ...(signature !== undefined && { signature }),
+            publicKey: walletId,
```
mjuchli-da: Still returns `walletId` instead of public key

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+        }
+        const wallet = await this.request<BitGoWallet>(
+            'POST',
+            `/api/v2/${this.coin}/wallet`,
+            {
+                label: name,
+                coin: this.coin,
+                type: 'custodial',
+                multisigType: 'tss',
+                ...(this.enterpriseId && { enterprise: this.enterpriseId }),
```
pawelstepien-da: There is an early check for falsy `this.enterpriseId`
```suggestion
                enterprise: this.enterpriseId.
```

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+                    )
+                }
+            }
+        }
+    }
+
+    async getKeys(): Promise<Key[]> {
+        const response = await this.request<{ wallets: BitGoWallet[] }>(
+            'GET',
+            `/api/v2/wallets?coin=${this.coin}&type=custodial`
```
pawelstepien-da: I think this returns paginated results with next page cursor `nextBatchPrevId`. If that's true then it should concatenate results from all pages.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+                await this.getKeys()
+            }
+            for (const publicKey of params.publicKeys!) {
+                const walletId = this.keyMap.get(publicKey)
+                if (!walletId) continue
+                const txsResponse = await this.request<{
+                    txRequests: BitGoTxRequest[]
+                }>(
+                    'GET',
+                    `/api/v2/wallet/${walletId}/txrequests?apiVersion=full&latest=true`
```
pawelstepien-da: Also here in API docs it's stated that it returns paginated results. We need to traverse all pages.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+
+    async getKeys(): Promise<Key[]> {
+        const response = await this.request<{ wallets: BitGoWallet[] }>(
+            'GET',
+            `/api/v2/wallets?coin=${this.coin}&type=custodial`
+        )
+        const keys: Key[] = []
+        for (const w of response.wallets) {
+            if (!w.keys[0]) {
+                // Non-TSS or incomplete wallet — walletId serves as both id and publicKey.
```
pawelstepien-da: I think it would be better to omit wallets without public key in response and cache, instead of using walletId as public key fallback.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,425 @@
...
+            }
+            const keychain = await this.request<BitGoKeychain>(
+                'GET',
+                `/api/v2/${this.coin}/key/${w.keys[0]}`
+            )
+            const publicKey = await this.derivePublicKey(
+                keychain.commonKeychain
+            )
+            this.keyMapSet(publicKey, w.id)
+            this.keyMapSet(w.id, w.id)
```
pawelstepien-da: I would remove adding walletId -> walletId mapping from cache everywhere to save half of space and only query the map by publicKey. walletId and publicKey should not be interchangable.

yarn.lock
pawelstepien-da, line 1: We switched from yarn to pnpm, `yarn.lock` should be removed.

pnpm-workspace.yaml
pawelstepien-da, line 1: Could you remove reformatting? I think `pnpm prettier . --write` should format it in the same way as on main.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,458 @@
...
+                } while (prevId)
+            }
+        }
+    }
+
+    async getKeys(): Promise<Key[]> {
+        const keys: Key[] = []
+        let prevId: string | undefined
+        do {
+            const qs = `coin=${this.coin}&type=custodial${prevId ? `&prevId=${encodeURIComponent(prevId)}` : ''}`
```
pawelstepien-da: What do you think about adding a higher `limit` param in qs? We could reduce amount of sequential requests if we fetched more items per page than default 25. Same for `getTransactions`

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,458 @@
...
+    // In-memory store mapping txRequestId → walletId for fast lookups.
+    // Not HA: restarts lose this cache; getTransaction can recover via the enterprise txrequests
+    // endpoint when enterpriseId is configured.
+    private txStore = new Map<string, string>()
+    // In-memory map of Ed25519 base64 publicKey → BitGo walletId.
+    // Populated by createKey and getKeys; used by getTransactions and signTransaction.
+    private keyMap = new Map<string, string>()
+    // Reverse map: BitGo walletId → Ed25519 base64 publicKey.
+    // Kept in sync with keyMap; used by formatTxRequest to return the real public key.
+    private walletKeyMap = new Map<string, string>()
```
pawelstepien-da, line 95: Those caches are useful for reducing amount of network traffic and handling indexing between wallet identifiers in Wallet Gateway (publicKey) and BitGo (walletId). But I can see a couple of scenarios where cache can be a breaking point.
1. `keyMap` is capped at 10k entries. With more wallets, `getKeys()` can discover a key and subsequently evict it during the same traversal. This could make `signTransaction` return `key_not_found`, even when it was present in API response. 
2. After process restart or `setConfiguration()`, `walletKeyMap` is empty, so `getTransaction()` recovers the walletId through the enterprise endpoint, but formatTxRequest() omits publicKey, because it's not in cache.
3. If paginated getKeys() fails in the middle of traversing pages, cache is left with incomplete state. This would make `*getTransactions` think it has complete state in cache, as it checks only that `cache.size !== 0`.
There may be more potential issues that I missed.

I would rework caches to follow those principles:
1. BitGo API should be the source of truth and cache should be performance optimization. Cache miss should fallback to API request.
2. Process restart or cache eviction should not return in different results than with cache.
3. Bulk cache refreshes should be atomic - write to cache should happen only after all pages have been traversed.
4. Non-empty map should not be treated as having complete state.
5. Forward and reverse mappings should be updated and evicted together, so they remain in sync.

core/signing-bitgo/src/index.test.ts
```diff
@@ -0,0 +1,581 @@
...
+const TX_REQUEST_ID = 'aaaabbbb-cccc-dddd-eeee-ffffffffffff'
+const FINGERPRINT =
+    '12206::122066c3b0ed7e4879579d53a0f04cabf54895fc4b410877adf88e61b68999b6e38d'
+
+const SIGNED_TX: BitGoTransaction = {
+    txId: TX_REQUEST_ID,
+    status: 'signed',
+    signature:
+        'RCS+Qh5w9VHk6Ih14jYTBcTB30RG5arj3ZmUr8mSTO7N6+zNPrJdXHUL13zE3LGfVDqvLEI76PKOczC5EL3OAw==',
+    publicKey: WALLET_ID,
```
pawelstepien-da, line 24: publicKey and walletId are not equal. Let's make tests reflect this.

core/signing-bitgo/src/index.test.ts
```diff
@@ -0,0 +1,581 @@
...
+                .getTransaction({ txId: TX_REQUEST_ID })
+            throwWhenRpcError(result)
+            expect(result.txId).toBe(TX_REQUEST_ID)
+            expect(result.status).toBe('signed')
+            expect(result.signature).toBe(SIGNED_TX.signature)
+            expect(result.publicKey).toBe(WALLET_ID)
+            expect(result.metadata?.signedBy).toBe(FINGERPRINT)
+        })
+
+        it('omits signature, publicKey, and metadata when absent', async () => {
```
pawelstepien-da, line 299: Test description states that publicKey is omited when absent, but in test it is present. Please align description and implementation of the test.

wallet-gateway/remote/src/ledger/transaction-service.test.ts
```diff
@@ -529,6 +529,128 @@ describe('TransactionService', () => {
             })
         })
 
+        describe('bitgo', () => {
+            it('returns the driver signature when signing completes', async () => {
```
pawelstepien-da, line 533: I think BitGo driver would never return `status: 'signed'` as response from `signTransaction`, it's implemented in such a way that it always returns `status: 'pending'`. 
It is fine that this case is handled in transaction service.

I only wonder if that is always the case when signing with BitGo that it always comes back unsigned? Or is there something like auto-sign? I assume it should still not break in such case, just require 2 clicks in gateway UI on "approve" button.

core/signing-bitgo/README.md
```diff
@@ -0,0 +1,97 @@
...
+| `baseUrl`      | No       | API base URL. Defaults to `https://app.bitgo.com` (prod). Use `https://app.bitgo-test.com` for testnet.                                |
+| `enterpriseId` | No       | BitGo enterprise ID. Required for `createKey`. Enables restart-safe `getTransaction` fallback via the enterprise txrequests endpoint.  |
+| `coin`         | No       | Canton coin identifier. Auto-detected: `tcanton` for `bitgo-test.com` URLs, `canton` for everything else (prod, proxies, custom URLs). |
+
+## Wallet Gateway configuration (`config.json`)
+
+```json
+{
+    "signingProvider": "bitgo",
+    "signingConfig": {
```
pawelstepien-da: We don't have these properties in config.json. Readme should describe environment variables instead.

core/signing-bitgo/README.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+Optional env vars for both modes: `BITGO_API_URL`, `BITGO_COIN`, `BITGO_MSG_TYPE` (`CANTON_SIGN_TOPOLOGY` default, or `CANTON_SIGN_TRANSACTION`).
+
+## Development
+
+```bash
+yarn build          # compile
+yarn test           # run tests
+yarn test:coverage  # with coverage report
```
pawelstepien-da: Please update to `pnpm x` instead of yarn.

core/signing-bitgo/scripts/test-connectivity.ts
```diff
@@ -0,0 +1,46 @@
...
+
+    if (!enterpriseId) {
+        console.log(
+            '\nSkipping createKey — set BITGO_ENTERPRISE_ID to test wallet creation'
+        )
+        return
+    }
+
+    console.log('\n=== 2. createKey (create a new TSS canton wallet) ===')
+    const key = await handler.createKey(`bitgo-canton-test-${Date.now()}`)
```
pawelstepien-da: File name implies it just tests the connection, but creating a wallet is a lasting write operation. I would prefer to have only a read operation in this test, just to verify connection, and createKey could be moved to `test-sign.ts`.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,458 @@
...
+        return txBase64
+    }
+    if (!Array.isArray(decoded)) return txBase64
+    if (decoded.length !== 1) {
+        throw new Error(
+            `BitGo signing driver only supports single-item topology transaction batches; received ${decoded.length}`
+        )
+    }
+    return decoded[0] as string
+}
```
pawelstepien-da: @mjuchli-da do you think this may be a problem, or something to document? Testing manually I only see single topology transaction generated, and what ultimately matters for wallet gateway side is signed multiHash. But I don't know if something could cause multiple topology transactions to be generated for an external party, and what consequences it could have if only first one would be sent to BitGo for key/wallet creation.

core/signing-bitgo/scripts/test-connectivity.ts
pawelstepien-da: Actually don't mind changing the 2 test scripts. Please remove them instead. We won't be able to use them for testing in CI, so it's better remove unused code.

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -723,6 +731,7 @@ export class TransactionService {
         if (tx.externalTxId) {
             signingResult = await driver
                 .getTransaction({
+                    userId,
```
pawelstepien-da: This goes to Securosys signing driver not Bitgo. Please remove that addition.

wallet-gateway/remote/src/ledger/wallet-allocation/wallet-allocation-service.ts
```diff
@@ -183,6 +195,17 @@ export class WalletAllocationService {
                     throw new Error('Securosys signing driver not available')
                 }
                 return this.securosysAllocator.createWallet(
+                    authContext.userId,
+                    authContext.email,
+                    partyHint,
+                    primary,
+                    vaultName
```
pawelstepien-da: This changes Securosys and even breaks the build.

core/signing-bitgo/src/bitgo.ts
```diff
@@ -0,0 +1,367 @@
...
+            }
+        )
+        this.txStoreSet(response.txRequestId, params.walletId)
+        return { txId: response.txRequestId }
+    }
+
+    async getTransaction(txId: string): Promise<BitGoTransaction | undefined> {
+        const walletId = this.txStore.get(txId)
+        if (walletId) return this.fetchTxRequest(txId, walletId)
+        if (!this.enterpriseId) return undefined
```
pawelstepien-da, line 239: If cache is cold, or txStore gets over 10001 entries and the one looked for was first (cache is FIFO and limited to 10k entries), then if enterprise ID is not set, the transaction is lost forever. 
Enterprise ID is also required to create a wallet.
I wonder if it even makes sense for wallet gateway to register BitGo sining driver in case user doesn't have, or forgets to set enterprise ID env var. WDYT @mjuchli-da?

wallet-gateway/remote/src/ledger/transaction-service.test.ts
ravibitgo: It will always comes back as `pending`. For signing driver we only allow `custody` wallets which require signing to be done by bitgo side.

wallet-gateway/remote/src/ledger/transaction-service.ts
ravibitgo: my bad, updated

wallet-gateway/remote/src/ledger/wallet-allocation/wallet-allocation-service.ts
ravibitgo: done

pawelstepien-da:
CI fails on prettier check. Please run `pnpm  prettier . --write` and it should fix the issue.

core/signing-bitgo/src/bitgo.ts
mjuchli-da: I think it is ok for now. If I remember correctly, at the moment we need only a single topology tx to allocate a party. Potentially a note in the documentation/readme would certainly make sense.

core/signing-bitgo/src/bitgo.ts
mjuchli-da: Not supporting the creating of a wallet is imho ok, as long as we throw an appropriate error and as long as `getKeys` is provided to allow assign exiting keys to a wallet.

mjuchli-da (approved):

pawelstepien-da (approved):

PR #2166, opened 2026-07-17
fix: add pagination to list tx

rukmini-basu-da:
This PR updates the `listTransactions` user-api and store interface to accept an optional parameter that consists of {limit: number, cursor: string}). 

The sql store and in memory implementations where both updated to query the same way with `createdAt desc nulls last`, using the `id desc` as the tiebreaker. The reason why the transactions table `createdAt` column is nullable (rather than something like currentTimestamp if not provided) is this was implemented for an audit trail and `currentTimestamp` is not necessarily accurate (hence the additional paging logic). The store requests pageSize+1 items, so if an additional item is returned, we know there is a next page (and this element is removed before returning the payload and also determines the `nextCursor`). 

Lastly, the `user-ui-activities` in the remote wallet-gateway used to `listTransactions` and get all the elements then implement pagination on the client side, so this was modified to use `this.transactions` as a paginated stream provided by the server. The `pageCursors` is sequentially cached where page 1 maps to undefined and as it's navigated forward, the `nextCursor` is sequentially cached, which helps with backwards/forwards traversal of the transactions pages.


wallet-gateway/remote/src/web/frontend/activities/index.ts
```diff
@@ -67,9 +72,11 @@ export class UserUiActivities extends BaseElement {
         `,
     ]
 
-    private get pagedTransactions() {
-        const start = (this.currentPage - 1) * this.pageSize
-        return this.transactions.slice(start, start + this.pageSize)
+    private get computedTotal() {
+        if (this.hasNextPage) {
+            return this.pageSize * this.pageSize + 1
```
pawelstepien-da: This doesn't seem correct to have pageSize squared. Did you mean `this.pageSize * this.currentPage + 1`?

pawelstepien-da: Nah, current page is not good either, because it would make total decline if you moved a page back.

I think with the next-page-cursor approach, maybe we should not show total at all?

alexmatson-da: maybe we can put the total pages in the API response?

core/wallet-store-sql/src/store-sql.ts
```diff
@@ -714,16 +716,18 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
                 ])
             )
             .orderBy('createdAt', 'desc')
-            .orderBy('id', 'desc')
```
pawelstepien-da, line 717: If it didn't cause any problems, then I would leave it just to be sure sure it's deterministic.
I'm also a bit worried about `orderBy('createdAt', 'desc')` if it's null - I'm not sure if it would be sorted before or after txes that have createdAt and could potentially differ between postgres and sqlite.

I think we should make sure everywhere where we sort by createdAt, that nulls are considered last.

rukmini-basu-da: good catch, updated!

alexmatson-da (approved):
thx!

PR #2183, opened 2026-07-21
feat(core-origin-manager): add origin check manager

mateuszpiatkowski-da:


core/provider-dapp/src/DappAsyncProvider.ts
```diff
@@ -127,7 +127,11 @@ export class DappAsyncProvider extends AbstractProvider<DappAsyncRpcTypes> {
 
         // Listen for the auth success event sent from the WK UI popup to the SDK running in the parent window.
         window.addEventListener('message', async (event) => {
-            if (!isSpliceMessageEvent(event)) return
+            if (
+                event.origin !== window.location.origin ||
+                !isSpliceMessageEvent(event)
+            )
```
alexmatson-da: hmm aren't these always non-matching though? supposing the WG popup is loaded on `https://wg.com` origin and the dapp is loaded on `https://mydapp.com` origin. A message emitted from the popup is always a separate origin from the dapp (and vice versa) 

I ran into basically the same issue with the multi-session PR, and came up with a sort of two-way handshake (request + ack back) to establish the expected origin: https://github.com/canton-network/wallet/pull/2161/changes#diff-996a6861dc1de45e9046316868d78c3584267529d4a02e513e9151da2a2b7e13R141

maybe we need to extend something like this so that the popup maintains a list of known dapp origins in LocalStorage, and reject if the event comes from an origin outside that list

mateuszpiatkowski-da: sounds good! I'll make the changes :)

internal/origin-check/package.json
```diff
@@ -0,0 +1,25 @@
+{
+    "name": "origin-check",
```
alexmatson-da: package should go into `core/` (as you mentioned)

alexmatson-da (dismissed):
looks good

core/test-token/README.md
```diff
@@ -2,6 +2,10 @@
 
 TypeScript wrapper package for the Test Token DAML codegen.
 
+## Requirements
```
mateuszpiatkowski-da: @mjuchli-da as per you request I'm throwing in a comment that you asked about :)

core/origin-manager/src/manager.ts
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { OriginHandshake, OriginHandshakeMessage } from './types'
+
+abstract class OriginManager {
+    protected allowedOrigins: Set<Location['origin']> = new Set()
+    protected abstract readonly messageToReceive: OriginHandshakeMessage
+    protected abstract readonly listenerCallback: (event: MessageEvent) => void
```
alexmatson-da: nitpick, maybe rename to `handshakeCallback` for clarity

core/origin-manager/src/manager.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+    private intervalMap = new Map<
+        Location['origin'],
+        ReturnType<typeof setInterval>
+    >()
+
+    /**
+     * Starts periodically broadcasting handshake messages to the child origin.
+     */
+    public poll(origin: Location['origin'], intervalMs = 500) {
```
alexmatson-da: another nitpick, I would maybe name this `connect` -- the polling is an implementation detail of connecting via the handshake

core/origin-manager/README.md
```diff
@@ -0,0 +1,105 @@
...
+originManager.poll(childOrigin)
+
+// Send a message using the safe postMessage method
+// This will only succeed if the handshake is complete
+originManager.postMessage({ type: 'greeting', data: 'hello' }, childOrigin)
+
+// Or manually check before sending
+if (originManager.assert(childOrigin)) {
+    window.postMessage(data, childOrigin)
+}
```
alexmatson-da, line 70: I think ideally the polling / handshake process is made invisible to the user, so they don't need to call `.poll` manually. `ParentWindowOriginManager` can probably track the connected status per origin, then

```ts
const originManager = new ParentWindowOriginManager()
const childOrigin = 'https://child.example.com'

// first time -- call `.poll` under the hood. throws any connection errors 
originManager.postMessage({ type: 'greeting', data: 'hello' }, childOrigin)

// second time -- already connected, msg goes straight through
originManager.postMessage({ type: 'greeting': data: 'hello-2' }, childOrigin)
```

core/origin-manager/README.md
```diff
@@ -0,0 +1,105 @@
...
+const originManager = new ChildWindowOriginManager()
+
+// The handshake is automatic; once complete, listener is removed
+
+// Send a message using the safe postMessage method
+// This will only succeed if the handshake is complete
+const parentOrigin = window.opener?.location.origin
+if (parentOrigin) {
+    originManager.postMessage({ type: 'response', data: 'world' }, parentOrigin)
+}
```
alexmatson-da: I would just move this check inside the implementation as well. The constructor could take an optional `parentOrigin` if the user wants to override it, but `window.opener` should be the default

alexmatson-da (dismissed):
nice stuff!

core/origin-manager/src/manager.ts
alexmatson-da: (nvm, ignore this comment if you decide to make `.poll` an internal detail instead of public, as per the comment below)

core/origin-manager/README.md
alexmatson-da: actually `window.opener.location` is problematic because browsers block access to this field across origins. The child manager should just use the origin it got from the handshake as the parent origin

alexmatson-da (approved):
thanks! 🚀

Issue #1795, opened 2026-05-17
wallet-sdk replacement


fayi-da:
This is a placeholder ticket for any existing usage of non wallet-sdk apis that were not easily replaceable while developing the feature. I'll update this with a list so it can be tracked and implemented afterwards.

PR #2185, opened 2026-07-21
chore: make wallet SDK connection and registry aware

fayi-da:
Another set of refactoring for #1795 . Most of the changes/cleanups are to deal with the fgact that a wallet-sdk instance is immutable. 

 - Clear provider-backed queries and legacy services when connections change
 - Recreate the Wallet SDK when authentication tokens or reachable registries change
 - Preserve unrelated wallet data when registry configuration changes
 - Moved transaction-history service ownership into React Query (will come in handy for reuse in extension :wink:)
 - Ensure transaction history remains bound to the provider that created it.
 - Updated mutation invalidation to use the new wallet-connection query hierarchy (invalidate `walletConnection` when anything changes and all child queries go poof :dash: )

### DIFF
--- a/examples/portfolio/src/contexts/ConnectionProvider.tsx
+++ b/examples/portfolio/src/contexts/ConnectionProvider.tsx
@@ -1,7 +1,7 @@
 // Copyright (c) 2025 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { useCallback, useEffect, useState } from 'react'
+import { useCallback, useEffect, useRef, useState } from 'react'
 import { useQueryClient } from '@tanstack/react-query'
 import * as sdk from '@canton-network/dapp-sdk'
 import { WalletConnectAdapter } from '@canton-network/dapp-sdk'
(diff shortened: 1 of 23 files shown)


examples/portfolio/src/contexts/ConnectionProvider.tsx
```diff
@@ -25,32 +25,67 @@ export const ConnectionProvider: React.FC<{ children: React.ReactNode }> = ({
...
+        (status: sdk.dappAPI.StatusEvent) => {
+            const nextToken = status.session?.accessToken
+            if (nextToken && nextToken !== currentSessionToken.current) {
+                clearResolvedServices()
+                currentSessionToken.current = nextToken
+                setSessionTokenVersion((version) => version + 1)
+            }
+        },
+        []
+    )
```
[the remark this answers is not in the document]
fayi-da: This is intentional. The benefit is that the user will not see a flicker/skeleton UI while the data is refetched.

examples/portfolio/src/lib/submit.ts
```diff
@@ -0,0 +1,27 @@
...
+    await provider.request({
+        method: 'prepareExecuteAndWait',
+        params: {
+            commands: [command],
+            commandId: v4(),
+            actAs: [actAs],
+            disclosedContracts,
+        },
+    })
+}
```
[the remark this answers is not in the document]
fayi-da: no thanks. nothing burger

examples/portfolio/src/contexts/ConnectionProvider.tsx
```diff
@@ -25,32 +25,67 @@ export const ConnectionProvider: React.FC<{ children: React.ReactNode }> = ({
...
+    const updateSessionToken = useCallback(
+        (status: sdk.dappAPI.StatusEvent) => {
+            const nextToken = status.session?.accessToken
+            if (nextToken && nextToken !== currentSessionToken.current) {
+                clearResolvedServices()
+                currentSessionToken.current = nextToken
+                setSessionTokenVersion((version) => version + 1)
+            }
+        },
+        []
```
mateuszpiatkowski-da, line 46: I'm not a react expert, but I think it would be beneficial to put `currentSessionToken` here right?

examples/portfolio/src/contexts/ConnectionProvider.tsx
```diff
@@ -25,32 +25,67 @@ export const ConnectionProvider: React.FC<{ children: React.ReactNode }> = ({
...
+                clearWalletConnectionQueries()
+                clearResolvedServices()
+                currentSessionToken.current = undefined
+                setAccounts([])
+            }
+            updateSessionToken(status)
+            setConnectionStatus(status)
+            setError(undefined)
+        },
+        [clearWalletConnectionQueries, updateSessionToken]
```
mateuszpiatkowski-da, line 61: same as above

mateuszpiatkowski-da (approved):
LGTM, left a few comments as nitpicks

examples/portfolio/src/contexts/ConnectionProvider.tsx
fayi-da: Usually, yes but in this case `currentSessionToken` is a stable value (it's defined as useRef) so it doesnt need to be there. The linter would have screamed if it wasn't a stable value

PR #2173, opened 2026-07-20
chore(deps): migrate monorepo library builds from tsup to tsdown

singhyash05:
## Summary
- Replaces `tsup` with `tsdown` for library builds across ~35 `core/*` and `sdk/*` packages.
- Shared config is now `tsdown.base.ts` (old `tsup.base.ts` and per-package `tsup.config.ts` files removed). Scripts and `devDependencies` point at `tsdown@^0.22.9`, including root and `yarn.lock`.
- Special cases kept working:
  - **browser:** `provider-dapp`, `provider-ledger`, `rpc-transport`, `splice-provider`
  - **node:** `signing-internal`, `signing-participant`, `wallet-test-utils`
  - **ledger-proto:** separate ESM/CJS outputs, CJS relative-import rewrite, CJS sourcemaps off (avoids `SOURCEMAP_BROKEN` spam)
- RPC generator TypeScript client template updated (`_package.json` + `tsdown.config.ts`) so `yarn generate:*` copies a config that resolves `tsdown.base.ts` from `core/<client>/`, not from the template directory.
- Source tweaks for rolldown: `export type` in `ledger-client`, `signing-lib`, and `splice-client`. Mixed default+named export packages set `outputOptions.exports: 'named'`.
- Root does **not** get `"type": "module"` — that would break `.commitlintrc.js`.
- **Follow-up fix:** `ParsedURL` no longer uses constructor parameter properties. Under `tsdown`/`oxc`, those emitted `this` before `super()` → `ReferenceError` in wallet-sdk e2e. Validate URL first, call `super()`, then assign fields.

Fixes #1965

## Test plan
- [x] Clean full rebuild (`yarn clean:all && yarn build:all`) — 45/45
- [x] `rg tsup` — no leftovers in source/config
- [x] `yarn workspace @canton-network/wallet-sdk build` — emit has `super(validateUrl(...))` before `this.ctx`
- [x] `yarn nx playwright:e2e @canton-network/example-portfolio` — 25 passed (localnet)
- [ ] CI e2e jobs green on PR

PR #2186, opened 2026-07-22
fix: add headers to tsdown configs

fayi-da:
Looks like leftover from https://github.com/canton-network/wallet/pull/2173


alexmatson-da (approved):
thx!

PR #2192, opened 2026-07-23
feat: migrate portfolio allocations to the wallet-sdk

fayi-da:
Part of #1795


examples/portfolio/src/utils/allocation.ts
```diff
@@ -0,0 +1,62 @@
...
+import type { Instruments, PortfolioInstrument } from '../types/instruments'
+import { normalizeRegistryUrl } from './registry'
+
+export const resolveAllocationAsset = ({
+    instrumentId,
+    instruments,
+    registryUrls,
+    reachableRegistryUrls,
+}: {
+    instrumentId: { admin: PartyId; id: string }
```
mateuszpiatkowski-da, line 14: nitpick: the naming looks off. Theoretically this should be `instrument` as per the definition in Metadata daml file.
```suggestion
    instrument: { admin: PartyId; id: string }
```

mateuszpiatkowski-da (approved):
LGTM :)

examples/portfolio/src/utils/allocation.ts
fayi-da: fair. I'll do another pass to rename this in similar places

Issue #2149, opened 2026-07-15
QOL Fixes


fayi-da:
Insights from manual testing

- [x] We currently drop registries that cannot be connected to and never retry until page refresh. 
- [x] The fields in the transfer dialog have varying inner padding. Use same inner padding for all fields
- [x] Outgoing transfers should not be shown in "Action Required"
- [x] Use date util for dates in Transfer Summary modal
- [x] Withdraw button for allocation dialog should be a secondary button not a warning red button
- [x] If one of the legs for an allocation request with you as the sender has been allocated, status should be "Partially Allocated" not Pending
- [x] Remove extra padding bottom from sidebar [here](https://github.com/canton-network/wallet-gateway/blob/46d7d186f4a61b2e426bf566aca178598c4f225c/examples/portfolio/src/routes/next/dashboard.tsx?plain=1#L167)
- [x] Set Allocate/Withdraw button size to small [here](https://github.com/canton-network/wallet-gateway/blob/46d7d186f4a61b2e426bf566aca178598c4f225c/examples/portfolio/src/components/dashboard/allocation-action-dialog-content.tsx?plain=1#L394)
- [x] Show the number of offers (transfers & allocation requests) on the sidebar badge
- [ ] Discuss and adjust font sizes with @pmunot-da
- [x] We can improve activity type in tx history by also checking the `reason` field
- [x] Ensure Tap modal is closed after successful op.
- [x] Rejection currently breaks transfer dialog. https://github.com/canton-network/wallet/issues/1867
- [x] Primary badge overlaps a long party hint on the sidebar
- [x] Sometimes unable to select an instrument for transfer even though i have a +ve balance (potential cache issue or tweak to `selectableInstruments`)
- [x] Validate expiry date for transfers, can only be in the future
- [x] Show expiry HH:mm in transfer offer dialog
- [ ] Change unlocked coin icon to new variant

PR #2200, opened 2026-07-24
feat: move isDevnet check to wallet-sdk

fayi-da:
Part of #2149 

- move isDevNet check to wallet-sdk
- make Tap amulet only operation
- delete unused services and dependencies


sdk/wallet-sdk/src/wallet/init/initializedSDK.ts
```diff
@@ -59,17 +59,10 @@ const createNamespace: {
             false
         )
 
-        const amuletService = config.validatorUrl
```
rukmini-basu-da, line 62: I think we left this in for backwards compatibility if people were already supplying the validatorURL in the amulet config (otherwise would've just removed the validatorURL completely from the config)

fayi-da: roger that, reverted @rukmini-basu-da 

I added a comment there too

examples/portfolio/src/hooks/query-options.ts
```diff
@@ -149,16 +148,29 @@ export const useAllocationsQueryOptions = (party: string | undefined) => {
     })
 }
 
-export const useIsDevNetQueryOptions = (sessionToken: string | undefined) => {
-    const { isDevNet } = usePortfolio()
+export const useIsDevNetQueryOptions = () => {
+    const { sdk } = useWalletSdk()
     const {
         token: { validatorUrl },
     } = usePortfolioConfig()
```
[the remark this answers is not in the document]
fayi-da: oops

rukmini-basu-da (approved):
thanks!

PR #2212, opened 2026-07-30
feat(example-test-token-v1-registry): switch from koa to express

### LINKED ISSUE #2205: switch from Koa to Express
Use Express instead of Koa in example-test-token-v1-registry.

mateuszpiatkowski-da:


mjuchli-da (approved):
Thanks!

PR #2239, opened 2026-08-05
Adding custom wallet display + copy in WG ui

### LINKED ISSUE #2234: Button to copy dApp API Url
The remote WG should display a button (settings / header bar) to copy the dApp API URL of the given WG instance. 

The Wallet Picker will not list every possible WG instance. If a user then wants to connect to its WG, the dApp API URL needs to be entered. While the host can be observed in the address bar, the actual API suffix is not necessarily known to the user.

katiepang-da:
fixes #2234 

1. Added custom wallet url in menu dropdown
2. Added url copy alongside it 

Screenshot 2026-08-05 at 16 32 14


mjuchli-da:
Implemented via #2240

PR #2249, opened 2026-08-07
feat(core-splice-codegen,example-test-token-v1-registry): create new core package, move test token

mateuszpiatkowski-da:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


scripts/src/generate-daml-codegen.ts
```diff
@@ -0,0 +1,94 @@
...
+    const darsDir = path.join(repoRoot, '.localnet/dars')
+
+    if (!existsSync(darsDir)) {
+        execSync('yarn script:fetch:localnet', {
+            cwd: repoRoot,
+            stdio: 'inherit',
+        })
+    }
+
+    Promise.all(
```
pawelstepien-da, line 78: I think we need `await` here

package.json
```diff
@@ -16,8 +16,8 @@
         "generate:signing": "yarn workspace @canton-network/core-rpc-generator build && yarn open-rpc-generator generate -c ./rpc-generator-config-signing.json",
         "generate:grpc": "tsx ./scripts/src/generate-protobufs.ts",
         "generate:tokenstandard": "tsx ./scripts/src/generate-token-standard.ts",
-        "generate:test-token": "tsx ./scripts/src/generate-test-token.ts",
-        "generate:all": "yarn generate:dapp && yarn generate:dapp-remote && yarn generate:user && yarn generate:signing && yarn generate:tokenstandard && yarn generate:test-token && yarn run prettier . --write > /dev/null 2>&1",
+        "generate:codegen": "tsx ./scripts/src/generate-daml-codegen.ts",
+        "generate:all": "yarn generate:dapp && yarn generate:dapp-remote && yarn generate:user && yarn generate:signing && yarn generate:tokenstandard && yarn generate:codegen && yarn run prettier . --write > /dev/null 2>&1",
```
pawelstepien-da, line 20: I didn't check if it's changed further in stack, but remember that we replaced yarn with pnpm

core/splice-codegen/README.md
```diff
@@ -0,0 +1,166 @@
...
+
+## Installation
+
+```sh
+yarn add @canton-network/core-splice-codegen
+```
+
+## Exports
+
+Top-level exports (from `src/index.ts`):
```
pawelstepien-da, line 18: I would go through this readme and check which sections actually provide value to reader. For example here the same info can be derived by quick look at index. You could either add some descriptions what is TestToken and what is OTCTrade, or just omit it. Build outputs section also provides exactly the same info that dedicated property in package.json - I would skip that.

mjuchli-da (approved):

PR #2235, opened 2026-08-05
refactor: rebase multi-sync example

mateuszpiatkowski-da:
<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

Closes #2241

PR #2251, opened 2026-08-07
feat(example-test-token-v1-registry): enable passing synchronizerId for allocation & transfer

mateuszpiatkowski-da:

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

### DIFF
--- a/examples/test-token-v1-registry/src/api/allocation-instruction/getAllocationFactory.ts
+++ b/examples/test-token-v1-registry/src/api/allocation-instruction/getAllocationFactory.ts
@@ -38,9 +38,8 @@ export const getAllocationFactory: TExpressOpenApiRequestHandler<
             })
             return
         }
-    }
-
-    if (fetchedFactories[0]) {
+        // no multi-sync mode
+    } else if (fetchedFactories[0]) {
         res.json({
             factoryId: fetchedFactories[0].contractId,
             choiceContext: emptyChoiceContext,
--- a/examples/test-token-v1-registry/src/api/allocation-instruction/getAllocationFactory.ts
+++ b/examples/test-token-v1-registry/src/api/allocation-instruction/getAllocationFactory.ts
@@ -7,6 +7,7 @@ import { TestToken } from '@canton-network/core-splice-codegen'
 import { APIError, emptyChoiceContext } from '../common'
 import { OffLedger } from '@canton-network/core-token-standard'
 import { TExpressOpenApiRequestHandler } from 'openapi-ts-router/express'
+import { synchronizerId } from '../../common/synchronizer'
 
 /**
  * Resolves or creates an allocation factory for initiating allocation workflows.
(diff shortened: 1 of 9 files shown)


examples/test-token-v1-registry/src/api/allocation-instruction/getAllocationFactory.ts
```diff
@@ -18,17 +19,30 @@ export const getAllocationFactory: TExpressOpenApiRequestHandler<
...
+        templateIds: [TestToken.DAR.TestTokenV1.TokenRules.templateId],
+    })
+
+    // multi-sync mode
+    if (synchronizerId.allocationInstruction) {
+        const syncFactory = fetchedFactories.find(
+            (factory) =>
+                factory.synchronizerId === synchronizerId.allocationInstruction
+        )
+        if (syncFactory) {
```
pawelstepien-da, line 34: If `syncFactory` is undefined, because none of factories matched `synchronizerId.allocationInstruction`, but fetchedFactories has at least one item, then we return `fetchedFactories[0]`.

I think a better flow in "multi-sync mode" would be to skip returning `fetchedFactories[0]` if no factory matches sync id, and go straight to "...and create one otherwise" part.

examples/test-token-v1-registry/src/index.ts
```diff
@@ -6,38 +6,64 @@ import allocationAPIRouter from './api/allocation/index.js'
...
 import { TestToken } from '@canton-network/core-splice-codegen'
 import sdk from './common/sdk'
+import { Server } from 'http'
+import { assignSynchronizerIds } from './common/synchronizer.js'
 
-const app = express()
+let server: Server
 
-await initOperatorParty()
+export const startRegistry = async (
```
pawelstepien-da, line 18: Nothing but tests calls `startRegistry`. Is that intentional in scope of that branch?

examples/test-token-v1-registry/src/index.test.ts
```diff
@@ -60,14 +81,35 @@ describe('entry file', () => {
         vi.clearAllMocks()
     })
 
+    it("shouldn't do anything", async () => {
+        const { stopRegistry } = await import('.')
```
pawelstepien-da, line 85: Now that importing index doesn't trigger any action, and start/stop has to be manually called, maybe it would be less confusing to have one static import at the beginning of the file?

examples/test-token-v1-registry/src/common/synchronizer.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+export const synchronizerId = {
```
pawelstepien-da, line 4: I'm suspicious of that approach to have a shared module level object with sync ids that is mutated after calling `startRegistry` if `options?.synchronizerIds` is provided, while stopRegistry doesn't reset it to empty state.

Example problematic scenario:
-`startRegistry` with options?.synchronizerIds -> `synchronizerId` modified, getAllocationFactory would check this syncs id
-`stopRegistry`
-`startRegistry` without options?.synchronizerIds -> `synchronizerId` from previous startRegistry still present, getAllocationFactory would check sync ids from previous `startRegistry`.

I would at least clear it in `stopRegistry`.
Or you could consider a different approach where syncIds live only inside registry instance, instead of a common module level object. Seems safer and easier to modify, i.e. if you wanted an option to have multiple instances of registry at the same time for some reason. But this is far-fetched scenario which I don't know we will ever need, so for me it's only important that you assure uninteded synchronizer ids don't appear when we didn't want them, so cleanup in stopRegistry probably would suffice.

examples/test-token-v1-registry/src/index.ts
```diff
@@ -6,38 +6,64 @@ import allocationAPIRouter from './api/allocation/index.js'
...
+        }
+    }>
+) => {
+    const app = express()
 
-/**
- * @customize The registry shouldn't be responsible for vetting daml files. We're doing this for development purposes only. Feel free to remove this when constructing your own token.
- */
-if (process.env.NODE_ENV === 'development') await TestToken.utils.vetDar(sdk)
+    await initOperatorParty(options?.operator)
```
pawelstepien-da, line 29: Not sure if it's a problem, but if you `startRegistry`, `stopRegistry` and `startRegistry` again, it will try to allocate same party twice.

examples/test-token-v1-registry/src/index.ts
mateuszpiatkowski-da: yes, this is preparing the grounds for https://github.com/canton-network/wallet/pull/2235

mjuchli-da (dismissed):

mjuchli-da (approved):

Issue #1867, opened 2026-05-22
Emit event on rejection


mjuchli-da:
Emit a `TxChangedFailedEvent` so that the dapp can signal it to the user.

PR #2253, opened 2026-08-07
test: more reliable e2e

### LINKED ISSUE #2226: Improve E2E testing utils
The testing utilities should be:
- expressive (state what happens and which conditions are required)
- finely granular (more detailed assertions that allow for more precise spotting of issues)
- non-flaky
- as fast as possible (fail fast, if flakiness permits).

pawelstepien-da:
Improvements to Wallet Gateway E2E tests aimed at reducing flakiness and making debugging easier.

Notable changes to tests:
-Added more assertions after actions, to ensure that they already took effect before proceeding to avoid race conditions or to fail early.
-Merged popup and separate page WG utils into one util with `isPopup` param.
-Reduced pollings that would make failures make tests take long time waiting for positive outcome that would never come.
-Wrapped test/utils step with test.step, added descriptions to steps and expects for more readable logs.
image

Notable changes to app logic, that were inducing flakiness, mostly race conditions not/rarely observed in regular usage, because of human speed:
-mock-oauth2 - added random as `icp` claim, which makes each JWT unique. Before they could be the same, if 2 token were generated in same second for same user and network.
-WG select network screen - submitting is locked until both `listNetworks` and `listIdps` requests finish. Before it was possible to click login with only `listNetworks` finished, while `listIdps` was still pending which caused UI error.
-WG UI root level component `user-app` now delays rendering page component for URL, until session verification completes, rendering "Loading..." instead. Before it could cause problems like page executing `connectCallback` just to be redirected to /login a moment later, or interrupting the session check (i.e. because a page list rendered too fast and test clicked "Add" button, causing a redirtect) resulted `getSessionId` going to `catch` which return sessionId null, which resulted in call to `attemptRemoveSession` - flaky and hard to debug.


examples/ping/tests/dfns.spec.ts
```diff
@@ -96,9 +67,13 @@ test.describe('Dfns external signing', () => {
...
     test('rejects a transaction in the wallet UI', async () => {
-        await wg.rejectTransaction(() => clickCreatePingContract(dappPage), {
-            waitForClose: true,
-        })
+        const { commandId } = await wg.rejectTransaction(
+            () => clickCreatePingContract(dappPage),
+            { waitForClose: true }
+        )
+
+        // TODO check dapp event, once removing tx starts emitting one
```
pawelstepien-da, line 75: https://github.com/canton-network/wallet/issues/1867

mjuchli-da (commented):
I just skimmed it and it looks very nice! (I'll give it a more detailed pass tomorrow).
That said, there are incredibly many occurrences of `await expect(lambda).toBeVisible({ timeout: 15000 })`. Would it maybe make sense to introduce a helper function `await expectWithTimeout(lambda)`?

alexmatson-da (dismissed):
thanks!

pawelstepien-da:
@mjuchli-da 
>That said, there are incredibly many occurrences of await expect(lambda).toBeVisible({ timeout: 15000 }). Would it maybe make sense to introduce a helper function await expectWithTimeout(lambda)?

Instead of expect wrapper util, I set default timeout per assertion in playwright config.
15s might be a bit generous, but it only prolongs tests that would fail by a couple of seconds compared to more optimistic timeout, while reducing flakiness if we underestimated a timeout in different environments.

mjuchli-da (approved):

PR #2168, opened 2026-07-18
feat: upgrade splice

mateuszpiatkowski-da:
Auto-generated by `yarn script:upgrade:splice`

PR #2266, opened 2026-08-10
fix: delete stale generated clients during codegen and wire up Canton 3.5.8

### LINKED ISSUE #1380: Fix codegen from canton openapi and asyncapi files
When there is a new canton version and we run `yarn script:generate:openapi`, this creates new clients with he new version  under `core/ledger-client-types/src/generated-clients` for the standard open api codegen, ledger provider types codegen, and the asyncapi, but this does not delete the old versioned files or update the imports to point to the new versions.

nicosampler:
## What

Closes #1380.

Bumping the Canton version in `scripts/src/lib/version-config.json` and re-running the codegen silently does nothing: the generated clients carry the version in their filename (e.g. `openapi-3.5.8.ts`), so a bump writes a new set of files next to the old one, nothing deletes the stale set, and the hand-written imports in `core/ledger-client-types/src/index.ts` keep pointing at the old version. This already happened: #2168 bumped devnet to 3.5.8, but `index.ts` stayed on 3.5.1 and `openapi-3.5.8-paths.ts` was never committed.

## What we changed

1. **Wired up Canton 3.5.8.** `index.ts` now imports 3.5.8, the missing `openapi-3.5.8-paths.ts` is generated, and the 3.5.1 clients are deleted.
2. **The codegen now prunes stale clients.** Both scripts delete generated clients whose version is absent from `version-config.json`, keeping the versions of every network (not just the one passed via `--network`, which would make a devnet run delete mainnet's client). Deleting a file that `index.ts` still imports is intentional: it turns a silent no-op into a compile error listing the files to repoint. The prune runs after generation, so a failed run leaves the tree untouched.
3. **`core/asyncapi-client` builds its channel map with `Object.fromEntries`.** 3.5.8 adds a WebSocket channel that 3.4.12 lacks, and the previous union-keyed write (`acc[currentVersion]`) only compiled while both versions had identical channels. Runtime behavior is unchanged.
4. **Deleted `asyncapi-3.5.1.d.ts` and its `.d.ts.map`,** stray tsc output committed in #1681.

## Verification

`yarn build:all` and `yarn test:all` pass. The prune was exercised end to end by pointing devnet at 3.5.1: it deleted the 3.5.8 client, printed the file to repoint, and the `ledger-client-types` build failed as intended.


mjuchli-da (approved):

rukmini-basu-da:
LGTM

PR #2268, opened 2026-08-11
feat(core-signing-taurus-protect): add Taurus-PROTECT signing provider

nbonvin:
Route Canton commands through a Taurus-PROTECT Canton gateway
 (tg-canton-gatewayd): a submit provider — the gateway prepares, signs
 (ECDSA P-256) and submits each CIP-103 command under Taurus governance,
 so the wallet never handles a raw signature; it forwards commands and
 reconciles status.

 New package @canton-network/core-signing-taurus-protect:
 - GatewayClient: JSON-RPC transport (connect, listAccounts,
   prepareExecute, getTransactionStatus) with lazy session and one-shot
   reconnect, a monotonic per-command status cache with requestId
   re-seed, a pre-flight 1 MiB body guard, and an always-set commandId
   as the end-to-end idempotency key.
 - routingOnlyCommands: on the encoded path the gateway validates the
   full argument tree only to discard it for the PTX, and real
   token-standard arguments fail that validation, so commands are
   reduced to the routing fields (templateId '#'-prefixed, argument
   trees blanked).
 - TaurusProtectSigningDriver: SigningDriverInterface over the client;
   the gateway lifecycle rides in metadata.gatewayStatus ('executed'
   maps to 'signed', unknown values to 'failed').

 wallet-gateway-remote wiring:
 - TransactionService: sign forwards the command once and re-polls
   after; execute reconciles and never posts to the ledger. A command
   is complete only on 'failed', or on 'executed' once the updateId is
   present — the gateway reports 'executed' before the updateId is
   observable. The ledger updateId stands in for the signature, and the
   executed payload/txChanged event is exactly
   {updateId, completionOffset: 0}.
 - TaurusProtectWalletAllocator: imports parties provisioned in
   Taurus-PROTECT (no topology transaction, no hash signing), guarding
   ambiguous prefixes; the vault picker lists parties by prefix.
 - Wallet sync skips Taurus-PROTECT wallets in the sync-needed check
   and the repair loop: parties are hosted remotely, so absence from
   the local rights snapshot says nothing.
 - Registered when TAURUS_PROTECT_GATEWAY_URL and
   TAURUS_PROTECT_GATEWAY_TOKEN are set.

 Single-tenant by design: one machine token serves all wallet users.

Signed-off-by: NBO <519181+nbonvin@users.noreply.github.com>


mjuchli-da (commented):
Thank you for the contribution. We are excited to list Taurus as a signing driver.
We have just switched over from `yarn` to `pnpm`. Would you mind rebasing the PR again? Afterward, we will do a detailed review.

mjuchli-da:
@pawelstepien-da will review shortly

pawelstepien-da:
@nbonvin could you instruct me how can I get token and api URL for env variables for manual testing? Is there a testing sandbox I could use?

pawelstepien-da:
A signing provider manages signing keys. Wallet Gateway hosts the party on its configured Canton participant, prepares transactions there, obtains signatures from the provider, and submits those signatures through the same participant.

This PR does not implement that signing provider architecture. It neither allocates the party nor signs transactions. Instead, it introduces a separate remote submission architecture that bypasses Wallet Gateway’s participant and modifies wallet synchronization to conceal the resulting inconsistency between local state and Canton.

The PR also does not meet the expectations in AI_POLICY.md. AI-assisted contributions must be understood and self-reviewed before submission so that review costs are not transferred to maintainers. The fundamental architectural mismatch and compensating special cases throughout the Wallet Gateway indicate that this responsibility was not met.

nbonvin:
@pawelstepien-da this is the architecture that was discussed privately with DA. We need to have this gateway, as Taurus-PROTECT is more than a simple signing machine. 

Can you please specify which part of the AI_POLICY.md is not met ?

PR #2269, opened 2026-08-12
fix: sanitize error

pawelstepien-da:
Error handling fixes:

1. Added global error handler middleware. It catches all errors that propagated through previous middlewares and sanitizes response. Before it would reach express handler which could return HTML error, that could expose sensitive internals.
```
curl -s -i -X POST http://localhost:3030/api/v0/user \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"listWallets","params":[]}'
```
Was:
```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Error: User is not connected<br> &nbsp; &nbsp;at getEthProviderError (<absolute-path>/node_modules/.pnpm/@metamask+rpc-errors@7.0.3_supports-color@10.2.2/node_modules/@metamask/rpc-errors/src/errors.ts:271:10)<br> &nbsp; &nbsp;at Object.unauthorized (<absolute-path>/node_modules/.pnpm/@metamask+rpc-errors@7.0.3_supports-color@10.2.2/node_modules/@metamask/rpc-errors/src/errors.ts:181:12)<br> &nbsp; &nbsp;at assertConnected (<absolute-path>/core/wallet-auth/src/auth-utils.ts:14:30)<br> &nbsp; &nbsp;at StoreSql.assertConnected (<absolute-path>/core/wallet-store-sql/dist/index.js:266:10)<br> &nbsp; &nbsp;at StoreSql.getSession (<absolute-path>/core/wallet-store-sql/dist/index.js:403:23)<br> &nbsp; &nbsp;at &lt;anonymous&gt; (<absolute-path>/wallet-gateway/remote/src/middleware/sessionHandler.ts:43:18)<br> &nbsp; &nbsp;at Layer.handleRequest (<absolute-path>/node_modules/.pnpm/router@2.2.0_supports-color@10.2.2/node_modules/router/lib/layer.js:152:17)<br> &nbsp; &nbsp;at trimPrefix (<absolute-path>/node_modules/.pnpm/router@2.2.0_supports-color@10.2.2/node_modules/router/index.js:342:13)<br> &nbsp; &nbsp;at <absolute-path>/node_modules/.pnpm/router@2.2.0_supports-color@10.2.2/node_modules/router/index.js:297:9<br> &nbsp; &nbsp;at param (<absolute-path>/node_modules/.pnpm/router@2.2.0_supports-color@10.2.2/node_modules/router/index.js:600:14)</pre>
</body>
</html>
```

Now:
`{"jsonrpc":"2.0","id":1,"error":{"code":4100,"message":"No active session found"}}`

2. sessionHandler now checks that there is accessToken attached to request and returns 401 early if not. Before it failed at store level.

3. Made JSON-RPC errors preserve their custom message. It was getting lost in `handleRpcError`.
```
curl -s -X POST http://localhost:3030/api/v0/dapp \
  -H 'Content-Type: application/json' \
  -d '{"jsonrpc":"2.0","id":3,"method":"thisMethodDoesNotExist","params":[]}'
  ```
Was:
`{"jsonrpc":"2.0","id":null,"error":{"code":-32601}}`

Now:
`{"jsonrpc":"2.0","id":null,"error":{"code":-32601,"message":"Method thisMethodDoesNotExist not found"}}`

4. Changed sessionHandler and jwtAuth to return auth failures in the JSON-RPC error shape. They previously returned `{"error": "<message>"}`, which fails ErrorResponse validation in HttpTransport, so clients discarded the real reason and fell back to a generic message derived from the HTTP status.

5. Stopped attaching the thrown error to error.data. handleRpcError put the raw error object into the response, which could leak sensitive internals. Full error still gets logged.
```
let response: ErrorResponse = {
    error: {
        ...rpcErrors.internal(),
        message: genericMessage,
        data: error,        // removed
    },
}
```
Verified by unit test `forwards the Error message but not the error object itself`

Additional fixes not related to issue:

1. Fixed middleware not applying properly to endpoints, if dapp or user api was configured to have path that doesn't start with `/api`.
 
2. Renamed `userPath ` to `userApiUrl` in wallet-gateway-configuration endpoint, because it's actually full url not just path and it's consistent with dappApiUrl.

### CODE REVIEW

File: wallet-gateway/remote/src/middleware/errorHandler.ts [hunk 1]
```diff
@@ -0,0 +1,78 @@
...
+                    error: {
+                        code: rpcErrors.internal().code,
+                        message: 'Something went wrong',
+                    },
+                })
+            )
+            return
+        }
+
+        res.status(500).json({ error: 'Internal Server Error' })
```
mjuchli-da, line 76: Is it not possible to tie this to the request somehow? 
Or are we relying on the fact that logs should be written and tied to the request id at the place where the error happens?

wallet-gateway/remote/src/middleware/errorHandler.ts
```diff
@@ -0,0 +1,78 @@
...
+    const { status, statusCode } = err as {
+        status?: unknown
+        statusCode?: unknown
+    }
+
+    return status === 413 || statusCode === 413
+}
+
+// Catches unhandled errors and prevents internal details like stack trace from reaching end user
+export function errorHandler(
```
mjuchli-da, line 27: Why is this separate from `handleRpcError` in `jsonRpcHandler.ts`?

pawelstepien-da: Because `jsonRpcHandler` wouldn't know about errors thrown anywhere else but controller. Issue this PR originated from reports a runtime error from sessionHandler middleware. When an error is thrown (or next(err) is called) along the journey of request through middlewares, express looks for closest next error middleware (a middleware that has four args: err, req, res, next). We didn't have one, so the error went straight to default express error handler and it responded with a verbose JS error in body.

I think it's a standard pattern to have one general error handler at the end of chain that would catch unexpected errors.

wallet-gateway/remote/src/middleware/errorHandler.ts
pawelstepien-da: I don't think we can fully rely on logs from all potential places error can come from as a way to correlate error with request. I am inclined to add request info next to the error in `logger.error` at the top of that handler. But I'm reluctant to just add whole request body and even more so headers to the log, because I'm worried about persisting sensitive information in logs. I'm thinking maybe a subset of request info like:
* http method
* http url path
* json rpc method (if present)
* json rpc id (if present)
* maybe boolean hasAuthContext, without leaking authContext details like `accessToken`

What do you think? It could guide us to api and method that caused the unhandled error, but unfortunately not exactly what was in the payload.

wallet-gateway/remote/src/middleware/errorHandler.ts
mjuchli-da: Ok makes sense to me, thanks. I think still think it would be good for @alexmatson-da to have a look here, given that he is the author of `jsonRpcHandler`

wallet-gateway/remote/src/middleware/errorHandler.ts
alexmatson-da: ultimately we would want trace ids for correlating errors to specific requests

alexmatson-da (approved):
thanks!

mjuchli-da (approved):

PR #2270, opened 2026-08-12
fix: repair multi-sync feature

jarekr-da:


scripts/src/start-localnet.ts
```diff
@@ -30,6 +48,13 @@ function ensureComposeOverride() {
             `        canton.participants.app-provider.ledger-api.command-service.max-commands-in-flight = ${CANTON_MAX_COMMANDS_IN_FLIGHT}`,
             `        canton.participants.app-user.ledger-api.command-service.max-commands-in-flight = ${CANTON_MAX_COMMANDS_IN_FLIGHT}`,
             `        canton.participants.sv.ledger-api.command-service.max-commands-in-flight = ${CANTON_MAX_COMMANDS_IN_FLIGHT}`,
+            ...(multiSync && cantonSupportsMultiSyncFeatureFlag
+                ? [
+                      '  multi-sync-startup:',
+                      '    volumes:',
+                      `      - ${CUSTOM_APP_SYNCHRONIZER_SC}:/app/app-synchronizer.sc`,
```
mateuszpiatkowski-da, line 59: `CUSTOM_APP_SYNCHRONIZER_SC` ends with app-synchronizer.sc, so the whole path would be `${path.join(rootDir), 'canton/multi-sync/app-synchronizer.sc'}/app/app-syncrhonizer.sc` which looks like a bug for me. Please confirm that this is intended

jarekr-da: isn't like a VOLUME mount: (with a colon) - that shows what in which path is `canton/multi-sync/app-synchronizer.sc` visible for docker
`canton/multi-sync/app-synchronizer.sc:/app/app-synchronizer.sc`?

mateuszpiatkowski-da: oh right! sorry bout that :D

PR #2295, opened 2026-08-18
feat: optional signing providers

### LINKED ISSUE #2259: Make all signing drivers optional via environment variables
See parent. The `wallet-kernel` driver is always "activated".

### LINKED ISSUE #2260: Make signing store optional
If `wallet-kernel` driver is not activated, we should allow to skip the signing store config.

pawelstepien-da:
.


wallet-gateway/remote/src/index.ts
```diff
@@ -110,9 +110,12 @@ if (hasDb) {
 }
 
 const hasSigningDb = process.argv.slice(2).includes('signing-db')
+// TODO check if I can get rid db and signing-db arguments.
```
pawelstepien-da, line 113: @mjuchli-da Do you think we can remove arguments 'db' and 'signing-db'? If not, then I will just remove TODO comment.

docs/wallet-gateway/signing-providers/index.md
```diff
@@ -4,13 +4,27 @@ The Wallet Gateway supports multiple signing providers that handle cryptographic
...
 
+Each signing provider can be explicitly disabled by setting its corresponding
+environment variable to `true`:
+
+- `WALLET_KERNEL_SIGNING_DISABLED`
+- `PARTICIPANT_SIGNING_DISABLED`
+- `FIREBLOCKS_SIGNING_DISABLED`
+- `BLOCKDAEMON_SIGNING_DISABLED`
+- `DFNS_SIGNING_DISABLED`
+- `SECUROSYS_SIGNING_DISABLED`
```
alexmatson-da, line 15: why environment variables rather than set from the config file? I think we should aim to use the config whenever possible since it is self-documenting (with JSON schema). The only time we cant put something in the config file is if it has to be a secret (like an API key) 

I suggest we add a new root-level config block for `providers` so users can do something like 

```js
{
  providers: {
    wallet_kernel: {
      enable: false
    },
    participant: {
      enable: false
   },
   fireblocks: {
      enable: true,
      api_path: "https://api.fireblocks.io/v1",
   },
   blockdaemon: {
     enable: true,
     capi2: "canton:testnet"
   }
}
```

pawelstepien-da: I think I wanted to have everything that controls whether signing driver is registered in one place.
I agree that it's better have non-sensitive signing provider configuration in WG config while keeping sensitive in env vars. I will rewrite it.

wallet-gateway/remote/src/index.ts
mjuchli-da: i don't think we can

core/signing-blockdaemon/README.md
```diff
@@ -68,15 +68,25 @@ The driver accepts a `BlockdaemonConfig` object:
...
-When running the Wallet Gateway (Remote), the Blockdaemon signing driver is configured using the following environment variables:
+When running the Wallet Gateway (Remote), configure non-secret settings in the
+Gateway config:
 
-- `BLOCKDAEMON_API_URL`: The base URL for the Blockdaemon API. Defaults to `http://localhost:5080/api/cwp/canton` if not set.
-- `BLOCKDAEMON_API_KEY`: The API key for authenticating with Blockdaemon.
+- `signingProviders.blockdaemon.baseUrl`: The base URL for the Blockdaemon API. Defaults to `http://localhost:5080/api/cwp/canton`.
+- `signingProviders.blockdaemon.caip2`: The CAIP-2 network identifier. Defaults to `canton:testnet`.
+
+Set `BLOCKDAEMON_API_KEY` environment variable separately.
```
mjuchli-da:
```suggestion
Set the following environment variable separately:
- `BLOCKDAEMON_API_URL`: The base URL for the Blockdaemon API. Defaults to `http://localhost:5080/api/cwp/canton` if not set.
- `BLOCKDAEMON_API_KEY`: The API key for authenticating with Blockdaemon.
```

core/signing-dfns/README.md
```diff
@@ -86,23 +86,27 @@ Dfns uses two-layer authentication: a long-lived service account JWT plus per-re
...
-- `DFNS_ORG_ID`: Your Dfns organization ID.
-- `DFNS_BASE_URL`: The base URL for the Dfns API. Defaults to `https://api.dfns.io` if not set.
-- `DFNS_CRED_ID`: The default credential ID for Dfns API authentication.
-- `DFNS_PRIVATE_KEY`: The private key for signing Dfns API requests.
-- `DFNS_AUTH_TOKEN`: The authentication token for the Dfns API.
+- `signingProviders.dfns.orgId`: Your Dfns organization ID.
+- `signingProviders.dfns.baseUrl`: The base URL for the Dfns API. Defaults to `https://api.dfns.io`.
+- `signingProviders.dfns.credId`: The default credential ID for Dfns API authentication.
+
+Set `DFNS_PRIVATE_KEY` and `DFNS_AUTH_TOKEN` in environment variables.
```
mjuchli-da:
```suggestion
Set the environment variables that contain secret/private values separately.
- `DFNS_PRIVATE_KEY`: The private key for signing Dfns API requests.
- `DFNS_AUTH_TOKEN`: The authentication token for the Dfns API.
```

wallet-gateway/remote/src/config/Config.ts
```diff
@@ -115,13 +115,122 @@ const bootstrapFromEnv = bootstrapConfigSchema.extend({
     ),
 })
 
+const providerEnableSchema = z.object({
```
mjuchli-da, line 118: I actually think we might not need this: what if we just assume that if the config holds a record for a signing provider, then the intention is to enable it. That means, by default, everything is disabled, but if the property `fireblocks` with all its fields exists, then we plan to enable the driver.
This should also work fine for `walletKernel` and `participant` which do not hold sub-fields – here we can use the `~` tilde symbol as per yaml standard.

The reason why I think we should at least check if this is doable is because it might make the subsequent plug&play of signing drivers more user friendly.

mjuchli-da (commented):
Thanks! It looks good but let us investigate if we can simplify the configuration as outlined in the comment below.

wallet-gateway/remote/src/config/Config.ts
pawelstepien-da: We could just assume that lack of provider's object in config means it won't register, but it is a breaking change that would require changing config after updating to Gateway version that will have this feature.

Explicit opt-out is one of two ways of preventing breaking changes I implemented, and the second is in case of missing config properties, like `signingProviders.blockdaemon.baseUrl`, to fallback on environment variables that were used before.

I see 3 options depending on how we want to balance no breaking changes vs simpler config and easier plug&play:
1. Keep as is. No breaking changes, more complexity.
2. Remove `enable` field, but keep checking environment variables as fallback for values that are now supposed to be in WG config. i.e if you have all env vars needed to register blockdaemon but no `signingProviders.blockdaemon` in config, it still registers. Only `wallet-kernel` and `participant` have breaking changes, simpler config for users, but still a bit convoluted for us to keep supporting deprecated env vars.
3. Only existence of `signingProviders.x` entry allows registration of signing driver. Simplest config and code maintenance, but absolutely requires updating config after upgrading to WG version with that feature, or no signing provider will register.

mjuchli-da: I suggest going with option 3. With the browser extension deployment, we will anyways need to opt-in a different default setup (e.g. no postgres but extension storage instead).

mjuchli-da: We agreed that if `signingProviders` is missing, we will supply a default config:
```
"signingProviders": {
        "participant": {},
        "gateway": {},
        "fireblocks": {},
        "blockdaemon": {},
        "all the rest": {},        
    }
```

pawelstepien-da:
Signing providers are now opt-in through WG config.signingProviders[provider-name]
Non-sensitive config fields are part of config, while secrets are still in env vars. Env vars names are unchanged by default, but now can be edited through corresponding config properties. 
Lack of config.signingProviders uses previous legacy logic of every signing provider registered by default (if required env vars are set)

PR #2314, opened 2026-08-18
fix: pin wallet-gateway-remote's *-store-sql deps to exact versions on publish

### LINKED ISSUE #2281: Pin exact dependencies for core-wallet-store-sql
The package.json for Wallet Gateway remote specifies a loose caret dependency for  `@canton-network/core-wallet-store-sql`. This is problematic if you want to run an older version of the gateway, because NPM resolves the newest version of store-sql and runs future migrations that may not work on the older version of the gateway. 

example: running `npx @canton-network/wallet-gateway-remote@1.5.0 -c /<REPO_ROOT>/wallet-gateway/test/config.json` will break because it executes migration 15 which adds the `origin` field to the sesions table, breaking logins. 

To fix, ensure that we publish the gateway with exact dependencies for the sql package.

ayushsingh82:
## Summary

`wallet-gateway/remote/package.json` declared `@canton-network/core-wallet-store-sql` and `@canton-network/core-signing-store-sql` (its other SQL-backed store dependency, same class of risk) as `workspace:^`. pnpm rewrites that to a caret range on publish, so `npm install` later can resolve a *newer* store-sql than the gateway version was actually built and tested against — and that newer store-sql can ship DB migrations the older gateway doesn't expect.

Concretely (per the issue): running `npx @canton-network/wallet-gateway-remote@1.5.0 -c <config>` pulls in a store-sql migration that adds an `origin` column to the sessions table, breaking logins on that older gateway version.

Changed both to `workspace:*`, which pnpm rewrites to the exact current version on publish instead of a range, so a given gateway release is always paired with the exact store-sql version it shipped with.

`pnpm-lock.yaml` updated accordingly (specifier only, no version changes since these are workspace links).

Fixes #2281

## Test plan

- [x] Confirmed via `pnpm-lock.yaml` that the specifier updated correctly for both packages, no unrelated changes
- [x] Ran `pnpm pack` on `wallet-gateway/remote` and inspected the resulting tarball's `package.json`: both deps are pinned to an exact version (e.g. `"1.11.1"`, no `^`), while other untouched `workspace:^` deps like `core-wallet-store` still resolve to `"^1.11.1"` as expected


alexmatson-da (approved):
thanks!

PR #2316, opened 2026-08-18
feat: json-rpc proxying for browser extension

### LINKED ISSUE #2296: Implement JSON-RPC messaging between dApp <-> content script <-> background script
Should include validation

alexmatson-da:
Fixes #2296 

This PR enables the extension to send RPC requests/responses between the dapp and the extension. This is backed by a dummy implementation for `connect` and `status` methods so that a dApp can successfully connect:

https://github.com/user-attachments/assets/650bb09a-0041-4c2d-9572-cc534694e48f

This PR also adds the user controller codegen, plus introduces [LogTape](https://logtape.org/intro) to handle logging... we may want to adopt this library in all our libraries, but for now I am trialing it in the extension.


wallet-gateway/extension/entrypoints/background/index.ts
```diff
@@ -1,6 +1,26 @@
 // Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
+import { configure } from '@logtape/logtape'
```
fayi-da, line 4: What happened to pino? :(

alexmatson-da: apparently you have to "[browserify](https://getpino.io/#/docs/browser)" it to work... so i was like naaahh. Btw logtape doesnt actually write anything itself, you have to pair with a sink -- technically we still just using `console` to output the logs from logtape

fayi-da (approved):

Issue #609, opened 2025-10-06
Generalize CSS styling


mjuchli-da:
This works should result in the styling of the Wallet GW becoming reusable etc.

PR #2323, opened 2026-08-19
Allow supplying a custom stylesheet for the dApp SDK's wallet picker popup

### LINKED ISSUE #1151: Allow dApp SDK implementers to supply their own stylesheet
This ticket is likely related to #609 but should be treated as higher priority.

It should be possible to supply a stylesheet (`.css`) which will have an effect on the `Discovery` popup that comes as part of the dApp SDK.

ayushsingh82:
## Summary

The wallet picker popup (opened via \`pickWallet\`, the default \`walletPicker\` for \`DappSDK\`) renders as a Lit custom element inside an isolated popup window, themed entirely through \`--wg-theme-*\` CSS custom properties (\`themes/default.css\`). Since it's a separate browsing context, a dApp implementer had no way to affect that theming -- setting those properties on their own page has no effect on a completely separate window/document.

Added an optional \`stylesheet\` to \`PopupOptions\` (\`core/wallet-ui-components/src/windows/popup.ts\`), injected as CSS text into the popup document's own \`<head>\`, outside the component's shadow root. Custom properties inherit through the shadow boundary by spec, so a \`:root { --wg-theme-primary-color: ... }\` rule there re-themes the component without needing to know its internal selectors.

Threaded through the existing call chain:
- \`popup.open(component, { stylesheet })\`
- \`pickWallet(entries, { stylesheet })\` (new \`PickWalletOptions\`)
- \`new DappSDK({ walletPickerStylesheet })\` (ignored if a custom \`walletPicker\` is supplied, since that bypasses the default popup entirely)

Related to #609. Fixes #1151

## Test plan

- [x] \`tsc --noEmit\` in \`core/wallet-ui-components\` and \`sdk/dapp-sdk\`: same pre-existing error count before/after (32 and 62 respectively -- both from unbuilt workspace deps, unrelated to this change), confirmed via before/after \`git stash\` comparison
- [x] \`eslint\` / \`prettier --check\` clean on all 5 changed files
- [x] Existing \`popup.test.ts\` / \`wallet-picker.test.ts\` suites (10 tests) pass unchanged
- [x] Added 2 new tests to \`popup.test.ts\`: one fetches the generated popup blob and confirms the custom stylesheet text actually lands in the document, the other confirms a literal \`</style\` in caller-supplied CSS can't break out of the \`<style>\` tag it's injected into
- [x] Added a \`PopupWithCustomStylesheet\` story to \`wallet-picker.stories.ts\` demonstrating a dark custom theme

### DIFF
--- a/core/wallet-ui-components/src/windows/popup.test.ts
+++ b/core/wallet-ui-components/src/windows/popup.test.ts
@@ -85,4 +85,45 @@ describe('popup', () => {
         expect(openSpy).toHaveBeenCalledTimes(1)
         expect(mockWin.location.href).toBe('https://second.example')
     })
+
+    it('includes a caller-supplied stylesheet in the popup document head', async () => {
+        const mockWin = createMockPopupWindow()
+        vi.spyOn(window, 'open').mockReturnValue(mockWin as unknown as Window)
+
+        class TestPopupElementWithStylesheet extends HTMLElement {
+            static styles = '.test { color: red; }'
+        }
+        customElements.define(
+            'test-popup-element-with-stylesheet',
+            TestPopupElementWithStylesheet
+        )
+
+        popup.open(TestPopupElementWithStylesheet, {
+            stylesheet: ':root { --wg-theme-primary-color: #123456; }',
+        })
+
+        const html = await (await fetch(mockWin.location.href)).text()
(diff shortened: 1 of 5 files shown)


mjuchli-da (commented):
LGTM! 
@mateuszpiatkowski-da to provide a 2nd review

PR #2326, opened 2026-08-19
chore: assert amulet.admin is present before building commands that need it

### LINKED ISSUE #1797: Wallet SDK V1 asserting registry data
we need to add assertion in certain areas surrounding create commands, case specifically is:
https://github.com/canton-network/wallet-gateway/blob/b3ca0b5bfe336dfd8bca466c621b33ecae58d49c/sdk/wallet-sdk/src/wallet/namespace/amulet/preapproval.ts#L59

if amulet.admin is undefined, then a correct command will be created, but without the expectedDSO field, which can even be submitted, but wont work as a valid CC preapproval.

ayushsingh82:
## Summary

`fetchAmulet()` (`sdk/wallet-sdk/src/wallet/namespace/amulet/namespace.ts`) returns `AssetBody` with `admin: PartyId` typed as always-present, but `parseAssets` (`sdk/wallet-sdk/src/wallet/namespace/utils/url.ts`) does no runtime validation of the raw registry response it's built from -- `admin` can actually be `undefined` at runtime despite what the type claims.

Three call sites currently pass `amulet.admin` straight into ledger commands with no guard:
- `preapproval.ts`'s create command (`expectedDso: amulet.admin`)
- `traffic.ts`'s `buyMemberTraffic`
- `namespace.ts`'s `tap`

If a registry response omits `admin`, this silently builds a command (e.g. a transfer preapproval with an undefined `expectedDso`) that can even be submitted but won't work correctly, instead of failing loudly at the point where the bad data enters the SDK.

Added one guard inside `fetchAmulet()`, using the SDK's existing `ctx.error.throw(...)` convention, which protects all three call sites at once.

Fixes #1797

## Test plan

- [x] `tsc --noEmit` in `sdk/wallet-sdk`: 115 errors before and after (exact match, confirmed via `git stash` comparison) -- all pre-existing from unbuilt workspace deps, zero new errors from this change
- [ ] Could not run the existing `amulet.test.ts`/`preapproval.test.ts`/`traffic.test.ts` suite -- it transitively needs `core-token-standard`, whose build requires Daml-generated codegen artifacts (`damljs/token-standard-models`) not available in the environment I worked in


sdk/wallet-sdk/src/wallet/namespace/amulet/namespace.ts
```diff
@@ -222,14 +222,26 @@ interface FeaturedAppNamespace {
...
+                      amuletCtx.registry.href,
+                  ])
+              )[0]
+            : amuletCtx.registry
+
+    // The type claims `admin` is always present, but it's parsed from an
+    // external registry response with no runtime validation -- a registry
+    // that omits it would otherwise silently produce commands (e.g. transfer
+    // preapprovals) with an undefined expectedDso.
+    if (!asset.admin) {
```
rukmini-basu-da: Let's move this validation to parseAssets function instead since this is the function that converts the registry data to an `AssetBody` and would be nice to add a unit test for this.

rukmini-basu-da (changes requested):

sdk/wallet-sdk/src/wallet/namespace/utils/url.ts
```diff
@@ -35,8 +35,21 @@ export function parseAssets(
     ctx: SDKContext,
     assets: Awaited<ReturnType<TokenStandardService['registriesToAssets']>>
 ) {
```
rukmini-basu-da: nitpick: add an explicit return type

```suggestion
) : AssetBody[] {
```

ayushsingh82: @rukmini-basu-da Done — added the explicit `AssetBody[]` return type. Pushed in fc24c08a.

rukmini-basu-da (approved):

PR #2328, opened 2026-08-19
feat: core-signing-wxt-store

### LINKED ISSUE #2298: Create a WXT-based signing store

rukmini-basu-da:


core/signing-store-wxt/package.json
```diff
@@ -0,0 +1,56 @@
+{
+    "name": "@canton-network/core-signing-store-wxt",
+    "version": "0.0.1",
```
alexmatson-da: can start at 1.0.0

core/signing-store-wxt/src/schemas.ts
```diff
@@ -0,0 +1,203 @@
...
+    privateKey: string | null
+    metadata: string | null
+    createdAt: string
+    updatedAt: string
+}
+
+export const signingKeyItem = (keyId: string) =>
+    storage.defineItem<SigningKeyRecord>(`local:signingKeyItem:${keyId}`, {})
+
+export const signingKeyIndexItem = () =>
```
alexmatson-da: how will we handle encryption in the future? I'm wondering if its better to store the whole DB in one object so we can encrypt a single item at once rather than multiple items... or would it be fine to encrypt each item individually?

core/signing-store-wxt/README.md
```diff
@@ -0,0 +1 @@
+# signing-store
```
mjuchli-da, line 1: Missing!

Package needs to be introduced here (see other readmes) and technicalities should in this case be well explained here.

core/signing-store-wxt/src/schemas.ts
```diff
@@ -0,0 +1,203 @@
...
+    name: string
+    publicKey: string
+    privateKey: string | null
+    metadata: string | null
+    createdAt: string
+    updatedAt: string
+}
+
+export const signingKeyItem = (keyId: string) =>
+    storage.defineItem<SigningKeyRecord>(`local:signingKeyItem:${keyId}`, {})
```
mjuchli-da, line 27: I suggest to move these kinds of getters and setters directly into the implementation of the store.
Furthermore I suggest creating a single wrapper for the getter and a single wrapper for the setter.

E.g.
```ts
get<T>(key: K)
set<T>(key: K, val : V)
```
, where `K` is a const/enum to increase type-safety. (Fallback can be an arg too)

core/signing-store-wxt/src/schemas.ts
```diff
@@ -0,0 +1,203 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
```
mjuchli-da, line 1: This file should imho only hold mappings from/to

core/signing-store-wxt/src/wxt-store.ts
mjuchli-da, line 1: rename to `store-wxt.ts` for consistency

core/signing-store-wxt/src/wxt-store.test.ts
mjuchli-da, line 1: Rename for consistency

core/signing-store-wxt/src/schemas.ts
rukmini-basu-da: The `storage.defineItem` returns an object that provides `.getValue()`, `.setValue()`, and `.removeValue()` and inherently `.getValue()` should return a `Promise<K>`, so I'm not sure if I see the benefit in adding an additional wrapper

core/signing-store-wxt/src/schemas.ts
```diff
@@ -0,0 +1,203 @@
...
+export const publicKeyIndexItem = (publicKey: string) =>
+    storage.defineItem<string | null>(
+        `local:signingKeyByPublicKey:${publicKey}`,
+        { fallback: null }
+    )
+
+export const nameIndexItem = (name: string) =>
+    storage.defineItem<string | null>(`local:signingKeyByName:${name}`, {
+        fallback: null,
+    })
```
alexmatson-da: I think it is a bit unnecessary to have all these separate storage items for retrieval... lets just have one item that stores `SigningKey[]` and use array methods (e.g. `.find(...)`) to retrieve keys by index/name/pubkey/etc

I doubt any single instance of the extension will hold more than 1-2 dozen keypairs, so performance shouldnt be an issue.

(Same for transactions)

rukmini-basu-da: Updated!

core/signing-store-wxt/src/schemas.ts
mjuchli-da: The wrapper exactly abstracts away `storage.defineItem`, which is then handled in the suggested "helper" methods. This way, the store methods work against a common interface, rather than the explicit wxt interface. It's nitpicking, but I think it's a tad cleaner.

rukmini-basu-da: as per convo added item/items functions

alexmatson-da (approved):
nice! thanks

PR #2333, opened 2026-08-20
feat: allow to alter signing provider for wallets

### LINKED ISSUE #2277: Allow to alter Signing Provider for Wallets
Currently, the `wallet-sync-service` handles the mapping of hosted parties found on the connected validator (network) with the signing key residing in the signing providers.

In some cases, such as key delegations, this mapping does not take place, leaving the wallet in a stale state (signing provider = participant; deactivated = true).

For reasons such as these, we want to allow users to alter the `sigining_provider_id` property of their `wallet`. 
However, given that a signing provider holds multiple signing materials, we also want to give the option to select the `public_key` (signing provider `getKeys` lists identifier and public key). Upon selecting the key material, we should then recompute the `namespace` property of the given wallet according to the selected `public_key` (even though the `partyId` might carry a different namespace!).

This change requires:
- User API endpoint to update wallet
- User API endpoint to list key material for given signing provider
- User UI to edit wallet with input fields for signing provider and key material

mateuszpiatkowski-da:
.


api-specs/openrpc-user-api.json
```diff
@@ -1065,13 +1065,13 @@
             }
         },
         {
-            "name": "listSigningProviderVaults",
-            "description": "Lists available vaults of a signing provider",
+            "name": "listSigningProviderKeys",
```
mjuchli-da, line 1068: why does that replace `listSigningProviderVaults` instead of it being an addition?

api-specs/openrpc-user-api.json
```diff
@@ -1088,21 +1088,125 @@
...
+                        },
+                        {
+                            "$ref": "#/components/schemas/Null"
+                        }
+                    ]
+                }
+            }
+        },
+        {
+            "name": "changeSigningProvider",
```
mjuchli-da, line 1162: should we use a more generalized `updateWallet` for this?

core/wallet-store-inmemory/src/store-internal.ts
```diff
@@ -132,6 +133,17 @@ export class StoreInternal implements Store, AuthAware<StoreInternal> {
         })
     }
 
+    async getWallet(constraint: WalletUniqueConstraint) {
```
mjuchli-da: Much needed introduction of `WalletUniqueConstraint`, thanks!

core/wallet-ui-components/src/routing.ts
```diff
@@ -5,6 +5,7 @@ export const ALLOWED_ROUTES = [
     '/api-keys/add',
     '/api-keys',
     '/login',
+    '/parties/edit',
```
mjuchli-da, line 8: Can be a follow-up but I suggest changing routes that involve `/parties` with `/wallets`

mock-oauth2/src/index.ts
```diff
@@ -143,7 +143,7 @@ async function main() {
         token.payload.sub = clientId // Mocked subject given the clientId
         token.payload.email =
             process.env.BLOCKDAEMON_API_EMAIL ||
-            'phillip.olesen@digitalasset.com'
+            'mateusz.piatkowski@digitalasset.com'
```
mjuchli-da, line 146: easter🥚

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/blockdaemon-wallet-allocator.ts
```diff
@@ -28,9 +28,24 @@ export class BlockdaemonWalletAllocator implements WalletAllocator {
         private store: Store,
         private logger: Logger,
         private partyAllocator: PartyAllocationService,
-        private signingDriver: SigningDriverInterface
+        protected signingDriver: SigningDriverInterface
     ) {}
 
+    private handleSigningError<T extends object>(result: SigningError | T): T {
```
mjuchli-da: The name is misleading. I initially thought it's related to signing, but it's related to a signing driver api call.

mjuchli-da: Also the method is repetitive; perhaps something to extract to a central location

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/fireblocks-wallet-allocator.ts
```diff
@@ -201,10 +208,4 @@ export class FireblocksWalletAllocator implements WalletAllocator {
 
         return this.store.updateWallet(walletUpdate)
     }
-
-    async getVaults(userId: UserId): Promise<{ vaults: string[] }> {
```
mjuchli-da, line 205: Ah I start to understand that vaults and keys are the same here.

mjuchli-da: Thus vaults listing was redundant; correct me if i'm wrong

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/kernel-wallet-allocator.ts
```diff
@@ -4,31 +4,38 @@
...
     constructor(
         private store: Store,
-        private logger: Logger,
+        private _logger: Logger,
         private partyAllocator: PartyAllocationService,
-        private signingDriver: SigningDriverInterface
+        protected signingDriver: SigningDriverInterface
     ) {}
 
+    private handleSigningError<T extends object>(result: SigningError | T): T {
```
mjuchli-da: duplicate

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/dfns-wallet-allocator.ts
```diff
@@ -33,9 +24,24 @@ export class DfnsWalletAllocator implements WalletAllocator {
         private store: Store,
         private logger: Logger,
         private partyAllocator: PartyAllocationService,
-        private signingDriver: SigningDriverInterface
+        protected signingDriver: SigningDriverInterface
     ) {}
 
+    private handleSigningError<T extends object>(result: SigningError | T): T {
```
mjuchli-da: duplicate

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/securosys-wallet-allocator.ts
```diff
@@ -4,33 +4,39 @@
...
 export class SecurosysWalletAllocator implements WalletAllocator {
     constructor(
         private store: Store,
         private logger: Logger,
         private partyAllocator: PartyAllocationService,
-        private signingDriver: SigningDriverInterface
+        protected signingDriver: SigningDriverInterface
     ) {}
 
+    private handleSigningError<T extends object>(result: SigningError | T): T {
```
mjuchli-da: duplicate

mjuchli-da (commented):
You nailed it conceptually! Left a few tweaks

core/wallet-ui-components/src/routing.ts
pawelstepien-da: Wasn't this change Wallets -> Parties in what is visible in UI a product decision?

wallet-gateway/remote/src/ledger/wallet-allocation/signing-providers/fireblocks-wallet-allocator.ts
pawelstepien-da: Exactly, Fireblocks vaults are equal to keys. Fireblocks had it earlier, because unlike other drivers it doesn't allow creating keys, you have to select pre-existing vault when creating a party.

wallet-gateway/remote/src/ledger/wallet-sync-service.ts
```diff
@@ -429,6 +427,7 @@ export class WalletSyncService {
                     signingProviderId:
                         resolvedSigningProvider.signingProviderId,
                     disabled: !isMatched,
+                    userId: '',
```
pawelstepien-da: Why empty userId? You can get current user ID by `this.authContext.userId`, but store.addWallet doesn't need it anyway, because it also extracts userId from authContext.

core/wallet-store-sql/src/store-sql.ts
```diff
@@ -136,6 +137,16 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
         })
     }
 
+    async getWallet(constraint: WalletUniqueConstraint) {
```
pawelstepien-da: I'm not sure about manually passing userId to store. Other methods take userId from authContext (`assertConnected`). User Alice shouldn't be able to get Wallet that belongs to Bob and it's good to have this constraint on store level. Unless you need to be able to get Wallet of any user from some service, I would adhere to convention that store gets userId from authContext, not from a param.

api-specs/openrpc-user-api.json
```diff
@@ -1088,21 +1088,125 @@
...
+            "params": [
+                {
+                    "name": "params",
+                    "schema": {
+                        "title": "GetWalletParams",
+                        "type": "object",
+                        "additionalProperties": false,
+                        "properties": {
+                            "userId": {
+                                "title": "userId",
```
pawelstepien-da: This allows a user to log in as user Bob to Network1 and get Wallet of user Alice for Network2. Even if UI constrains it, one could execute a request manually.

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -1215,15 +1218,65 @@ export const userController = (
...
             }
-            return walletAllocationService.getVaults(
-                assertConnected(authContext),
-                params.signingProviderId as SigningProvider
+            return keys
+        },
+        changeSigningProvider: async (
+            params: ChangeSigningProviderParams
+        ): Promise<null> => {
+            const { signingProviderId, partyId, publicKey } = params
```
pawelstepien-da, line 1241: I'm not sure if we should trust that request payload contains singingProviderId that exists and a publicKey that belongs to that signing provider. Perhaps we should try to do `listSigningProviderKeys` (or extract a subset of that method this one) and compare request params against result before changing wallet record?

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -1215,15 +1218,65 @@ export const userController = (
...
             )
-            if (!drivers[params.signingProviderId as SigningProvider]) {
+            const keys = await walletAllocationService.getKeys(
+                assertConnected(authContext),
+                params.signingProviderId as SigningProvider
+            )
+            if (
+                !drivers[params.signingProviderId as SigningProvider] ||
+                !keys
+            ) {
```
pawelstepien-da: I would switch order - first check that driver is registered, act on it (getKeys) afterwards.

wallet-gateway/remote/src/web/frontend/parties/edit/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    protected get form() {
+        return html`
+            <wg-wallet-edit-form
+                .signingProviders=${this.allowedSigningProviders}
+                .publicKeys=${this.publicKeys}
+                .selectedSigningProvider=${this.selectedSigningProvider}
+                .partyId=${this.walletConstraint.partyId}
+                .selectedPublicKeyId=${this.selectedPublicKey}
```
pawelstepien-da: Does that work? Inside edit-form we compare wallet.publicKey against key.id. I assumed those are not (or not always) equal.

wallet-gateway/remote/src/web/frontend/parties/index.ts
```diff
@@ -153,7 +156,9 @@ export class UserUiParties extends BaseElement {
                             <wg-wallet-card
                                 .wallet=${wallet}
                                 verified
+                                .editable=${wallet.signingProviderId !== SigningProvider.PARTICIPANT}
```
pawelstepien-da, line 160: I see a problematic case related to wallet-sync. If sync finds a party but can't find a key in signing providers, it makes its signing provider "participant" and disabled: true. It could be that the party that needs editing of it's signing provider and keys can't access edit form.
I don't have definitive answer how this should be handled. My first idea is that disabled allows you to edit the party even if participant.
I feel we could come up with a better solution than such workaround based on placeholder signing provider we chose for wallet-sync when it can't figure out signing provider. Maybe we should allow changing from participant to other signing providers, but not the other way around, but it would be problematic with how wallet sync figures out that a party has signing provider participant if party namespace matches participant node namespace, before going through signing drivers keys, plus we don't know yet how it works on canton topology side with key delegation in such case. 

There is so many questions and implications related to wallet-sync that I think it needs a separate follow-up ticket.

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -596,10 +594,15 @@ export class WalletGateway {
     private async waitForPopupToCloseAfterAction(
         popupPage: Page
     ): Promise<void> {
-        // For dApp-triggered approvals the popup is opened with
-        // `closeafteraction`, so success is signalled by the popup closing.
+        // dApp-triggered approvals often close the popup, but newer flows can
```
pawelstepien-da: I preferred previous comment. 
It's not like it happens "often" or "rarely", nor there are "newer" and "older" flows. It is deterministic based on whether approve screen is triggereed via dapp-sdk, or by going to WG transaction list, you will get query param closeafteraction or not.

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -276,11 +276,14 @@ export class WalletGateway {
                     .getByRole('checkbox', { name: 'Set as primary wallet' })
                     .check()
             }
-            await (
-                await this.page()
-            )
-                .getByRole('button', { name: 'Create' })
-                .click()
+            const submitButton = (await this.page()).getByRole('button', {
+                name: /^(Add|Create)$/,
```
pawelstepien-da: There is more than one button text?

wallet-gateway/remote/src/web/frontend/parties/index.ts
```diff
@@ -232,6 +243,12 @@ export class UserUiParties extends BaseElement {
             })
     }
 
+    private _onWalletEdit(e: WalletCardEditEvent) {
+        setLocationHref(
+            `/parties/edit?partyId=${e.wallet.partyId}&networkId=${e.wallet.networkId}&userId=${e.wallet.userId}`
```
pawelstepien-da: Could you use `toRelPath` also here? It could work on with `dev` but fail with `build` without that.

wallet-gateway/remote/src/web/frontend/parties/edit/index.ts
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private async onEditParty(event: WalletEditEvent) {
+        if (!this.loaded) return
+        this.submitting = true
+
+        try {
+            await this.userClient.request({
+                method: 'changeSigningProvider',
+                params: event,
```
pawelstepien-da: I wouldn't pass DOM Event directly to API, better extract specific properties from it.

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -1215,15 +1218,65 @@ export const userController = (
...
+        },
+        changeSigningProvider: async (
+            params: ChangeSigningProviderParams
+        ): Promise<null> => {
+            const { signingProviderId, partyId, publicKey } = params
+            const network = await store.getCurrentNetwork()
+            if (network === undefined) {
+                throw new Error('No network session found')
+            }
+            if (!network.adminAuth) {
```
pawelstepien-da, line 1255: This method doesn't really need adminAuth, but I understand that we need it for PartyAllocationService constructor, even tho we only need it to create a fingerprint which doesn't require adminAuth.
I created a separate tech debt issue to address those unnecessary adminAuth dependencies https://github.com/canton-network/wallet/issues/2360

api-specs/openrpc-user-api.json
mateuszpiatkowski-da: AFAIK this method returns a set of keys assigned to a signing provider. "Vaults" is a reserved keyword for Fireblocks, but it doesn't make sense to keep this naming convention when extending the functionality to other SP, hence the change in naming :)

api-specs/openrpc-user-api.json
mateuszpiatkowski-da: I'm a little reluctant about this - `updateWallet` would suggest the ability to edit any/all properties of a specified wallet, whereas here we're only interested in updating the portion of the data associated with the SP. If you want I could create a generic method called like that and enable the user to update any property and use it to update the SP data where required

mock-oauth2/src/index.ts
mateuszpiatkowski-da: omg I don't remember changing this 🙈  I'll revert this (wouldn't like to override our master signature 😬)

wallet-gateway/remote/src/ledger/wallet-allocation/wallet-allocation-service.ts
```diff
@@ -45,6 +42,24 @@ export class WalletAllocationService {
...
+                return this.blockdaemonAllocator
+            case SigningProvider.DFNS:
+                return this.dfnsAllocator
+            case SigningProvider.FIREBLOCKS:
+                return this.fireblocksAllocator
+            case SigningProvider.SECUROSYS:
+                return this.securosysAllocator
+            case SigningProvider.WALLET_KERNEL:
+                return this.kernelAllocator
+            case SigningProvider.PARTICIPANT:
```
pawelstepien-da, line 75: Now that we have BitGo driver merged, you will need to sync your branch with main and add case for it here and make sure getKeys is available in it's allocator.

wallet-gateway/remote/src/web/frontend/parties/index.ts
mateuszpiatkowski-da: Why do we assume that the SP is the participant in the first place? Shouldn't that be a problem for the user to display? Something like a greyed-out component that is disabled entirely with a tooltip describing the problem? I think that assuming something like that may lead to a series of bugs/problems in the future (FYI/rant :P)

I think a follow-up task is a sound idea.

pawelstepien-da: That was a decision we made long ago in early days of wallet-sync development, to set wallet signingProvider as participant with disabled flag set to true, if we can't figure out signing provider from party namespace. We didn't think it would matter. That participant was just to fill required column and disabled flag was supposed to handle every implication. Now that we have a feature that allows altering signing provider and keys, using participant as a fallback is annoying, so perhaps it would be better have something like signingProvider "undetermined" instead.

api-specs/openrpc-user-api.json
mjuchli-da: Makes sense yes.

api-specs/openrpc-user-api.json
mjuchli-da: We can always widen the scope, I agree with you.

mjuchli-da (dismissed):

core/wallet-ui-components/src/components/wallet-edit-form.ts
```diff
@@ -0,0 +1,201 @@
...
+    protected get isDisallowedSigningProvider(): boolean {
+        return (
+            !this.selectedSigningProvider.length ||
+            this.disabledSigningProviders.includes(this.selectedSigningProvider)
+        )
+    }
+
+    override connectedCallback(): void {
+        super.connectedCallback()
+        this.submitDisabled = this.isDisallowedSigningProvider
```
pawelstepien-da: It doesn't allow you to change they key, if you don't change signing provider before. It resolves to true and disables submit button. Perhaps a race condition.

core/wallet-ui-components/src/components/wallet-card.ts
```diff
@@ -314,7 +314,7 @@ export class WgWalletCard extends BaseElement {
     }
 
     private renderEditButton() {
-        return this.editable
+        return this.editable || this.wallet?.disabled
```
pawelstepien-da, line 317: I would also check reason for wallet being disabled. If it's because sync couldnt determine which signing provider it is, then editable. Other case is when wallet was originally participant, but user no longer has rights to the party and then I believe we shouldn't allow editing.
```
return this.editable || (this.wallet?.disabled && this.wallet?.reason === WALLET_DISABLED_REASON.NO_SIGNING_PROVIDER_MATCHED)
```

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -276,11 +276,14 @@ export class WalletGateway {
                     .getByRole('checkbox', { name: 'Set as primary wallet' })
                     .check()
             }
-            await (
-                await this.page()
-            )
-                .getByRole('button', { name: 'Create' })
-                .click()
+            const submitButton = (await this.page()).getByRole('button', {
```
pawelstepien-da: I'm pretty sure that if you revert all changes to e2e tests they will work again :)

core/wallet-ui-components/src/components/wallet-card.ts
mjuchli-da: I consider this optional for this PR: we can make a follow-up where we introduce some intelligence there.

pawelstepien-da: Fine by me

wallet-gateway/remote/src/web/frontend/parties/add/index.ts
```diff
@@ -152,35 +60,18 @@ export class UserUiAddParty extends BaseElement {
...
-            </div>
-
-            <div class="form-wrap">
-                <wg-wallet-create-form
-                    .signingProviders=${this.signingProviders}
-                    .networkIds=${this.networkIds}
-                    .vaultSigningProviders=${UserUiAddParty.vaultSigningProviders}
-                    .vaults=${this.vaults}
-                    ?vaultsLoading=${this.vaultsLoading}
-                    .submitLabel=${'Create party'}
```
pawelstepien-da, line 176: Removal of this line makes e2e tests fail, because it changes text of submit button and tests can't see it.

mjuchli-da (approved):

PR #2336, opened 2026-08-20
fix: properly setup user-rpc-client for browser-ext

### LINKED ISSUE #2308: authenticate with oauth

### LINKED ISSUE #2330: investigate @webext-core/proxy-service
https://www.npmjs.com/package/@webext-core/proxy-service

should allow us to directly call the controller methods from inside the content script, and drop the custom wrapper we have around `runtime.Port` messaging. 

todo after initial delivery

alexmatson-da:
enables the extension UI to call the user API!

image

Also switches to `web-ext/proxy-service` to let any extension component (UI, content script, etc) talk to the background script. 

Fixes #2308 
Fixes #2330


fayi-da (approved):
🫡

PR #2339, opened 2026-08-20
Create separate SynchronizerId type, mirroring PartyId

### LINKED ISSUE #522: Create separate type `SynchronizerId`
https://github.com/hyperledger-labs/splice-wallet-kernel/blob/dc3e74344e40856748b380ec7596729e88a057a7/sdk/wallet-sdk/src/topologyController.ts#L305

ayushsingh82:
## Summary

- `synchronizerId` is passed around as a bare `string` in several places even where a sibling `partyId` parameter already uses the `PartyId` type from `core-types`.
- Added a `SynchronizerId` zod schema/type in `core/types/src/index.ts`, following the exact same pattern as `PartyId` (schema + `z.infer` type + example constant).
- Applied it at the sites that already sit right next to a `PartyId`-typed value:
  - `SDKContext.defaultSynchronizerId` (`sdk/wallet-sdk/src/wallet/init/types/context.ts`)
  - `CreatePartyOptions.synchronizerId` (`sdk/wallet-sdk/src/wallet/namespace/party/external/types.ts`)
  - `SignedPartyCreationService.allocate()`'s `synchronizerId` param (`.../party/external/signed.ts`, which already imports `PartyId`)
  - `MergeDelegationNamespace.setup()`'s `synchronizerId` param (`.../token/utxos/mergeDelegation.ts`, same file already imports `PartyId`)

I deliberately left the ~200 other bare `synchronizerId: string` occurrences elsewhere in the repo as-is — the large majority are in auto-generated OpenAPI/AsyncAPI/protobuf client code (`core/ledger-client-types/src/generated-clients/**`, `core/ledger-proto/src/_proto/**`) that would just get overwritten on the next codegen run. This mirrors how `PartyId` itself isn't retrofitted everywhere either (`grep -rn "partyId: string"` still finds ~27 files outside the generated clients).

Note: like `PartyId`, `SynchronizerId` isn't a branded type (`z.infer` of a `z.string()` schema resolves structurally to `string`), so this is a type-level-only, zero-runtime-behavior change everywhere it's applied.

Fixes #522

## Test plan

- [x] Added unit tests for the `SynchronizerId` schema (valid id, missing `::` separator, too-short) in `core/types/src/index.test.ts` — 14/14 passing in that suite.
- [x] `npx vitest run --project node src/wallet/namespace/party/party.test.ts` (wallet-sdk) — 18/18 passing.
- [x] `eslint` and `prettier --check` clean on all 6 changed files.
- [x] `tsc --noEmit` clean on every touched file (pre-existing unrelated module-resolution errors elsewhere in `wallet-sdk` come from other packages needing a full monorepo build, not from this change).

### DIFF
--- a/core/types/src/index.ts
+++ b/core/types/src/index.ts
@@ -26,8 +26,6 @@ export const PartyId = z
 
 export type PartyId = z.infer<typeof PartyId>
 
-export const SYNCHRONIZER_ID_EXAMPLE = 'sync::122012312312312312123'
-
 export const SynchronizerId = z.string().includes('::').min(10)
 
 export type SynchronizerId = z.infer<typeof SynchronizerId>
--- a/core/types/src/index.ts
+++ b/core/types/src/index.ts
@@ -26,6 +26,12 @@ export const PartyId = z
 
 export type PartyId = z.infer<typeof PartyId>
 
+export const SYNCHRONIZER_ID_EXAMPLE = 'sync::122012312312312312123'
+
+export const SynchronizerId = z.string().includes('::').min(10)
+
+export type SynchronizerId = z.infer<typeof SynchronizerId>
+
 export const HttpUrl = z
     .url({
         message: 'Must be a valid HTTP or HTTPS URL',
--- a/core/types/src/index.test.ts
+++ b/core/types/src/index.test.ts
@@ -7,8 +7,9 @@ import {
     isSpliceMessage,
     isSpliceMessageEvent,
     SynchronizerId,
-    SYNCHRONIZER_ID_EXAMPLE,
 } from './index'
+
+const SYNCHRONIZER_ID_EXAMPLE = 'sync::122012312312312312123'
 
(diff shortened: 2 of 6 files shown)


core/types/src/index.ts
```diff
@@ -26,6 +26,12 @@ export const PartyId = z
 
 export type PartyId = z.infer<typeof PartyId>
 
+export const SYNCHRONIZER_ID_EXAMPLE = 'sync::122012312312312312123'
```
mjuchli-da: Example should not be part of the types library and only be used in tests

mjuchli-da (changes requested):
Minor correction needed

core/types/src/index.ts
ayushsingh82: @mjuchli-da Done — moved `SYNCHRONIZER_ID_EXAMPLE` out of the types library and into `index.test.ts` as a local test-only constant. Pushed in efd2d194.

PR #2345, opened 2026-08-21
fix: scope validation

### LINKED ISSUE #2334: Allow token scope claims to be superset of network.auth.scope

pawelstepien-da:
Removed validation of `accessToken.scope` and/or `accessToken.scp` against `network.auth.scope` in `userApi.addSession`.
As OIDC standard permits auth server to return token with different `scope`/`scp` claims than requested, it is not possible to reliably validate them against `network.auth.scope`. `iss` and `aud` claims comparison against `network.auth` remain main mechanism for ensuring token is meant for a specific network when calling `addSession`.

### DIFF
--- a/wallet-gateway/remote/src/user-api/token-network-matching.ts
+++ b/wallet-gateway/remote/src/user-api/token-network-matching.ts
@@ -1,7 +1,7 @@
 // Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
 // SPDX-License-Identifier: Apache-2.0
 
-import { Idp, Auth } from '@canton-network/core-wallet-auth'
+import { Idp } from '@canton-network/core-wallet-auth'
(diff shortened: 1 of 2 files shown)


wallet-gateway/remote/src/user-api/token-network-matching.ts
```diff
@@ -17,31 +17,6 @@ function normalizeAudienceClaim(value: JWTPayload['aud']): string[] {
     return []
 }
 
-function normalizeScopeClaim(
```
mjuchli-da, line 20: I agree with the removal but where was this stuff used?

mjuchli-da (approved):

wallet-gateway/remote/src/user-api/token-network-matching.ts
pawelstepien-da: In user API `addSession` only, to validate that accessToken claims match networkId passed to this method.

PR #2358, opened 2026-08-25
feat: add sign and execute for extension

### LINKED ISSUE #2303: user API: implement `sign`, and `execute`

fayi-da:
Mostly a copy of remote (slightly stripped down). Will be able to test after the tx and approval stuff is implemented

Closes #2303

### DIFF
--- a/wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
+++ b/wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
@@ -1,0 +1,187 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { LedgerClient, type Types } from '@canton-network/core-ledger-client'
+import {
+    isRpcError,
+    SigningProvider,
+    type SigningDriverInterface,
+} from '@canton-network/core-signing-lib'
+import type { AuthContext } from '@canton-network/core-wallet-auth'
+import type {
+    Store,
+    Transaction,
+    Wallet,
+} from '@canton-network/core-wallet-store'
+import type { Logger } from 'pino'
+import type {
+    ExecuteParams,
+    ExecuteResult,
+    SignParams,
+    SignResultSigned,
+} from './rpc-gen/typings.js'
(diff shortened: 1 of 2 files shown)


wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
```diff
@@ -0,0 +1,187 @@
...
+} from '@canton-network/core-wallet-store'
+import type { Logger } from 'pino'
+import type {
+    ExecuteParams,
+    ExecuteResult,
+    SignParams,
+    SignResultSigned,
+} from './rpc-gen/typings.js'
+
+export class TransactionService {
```
mjuchli-da, line 24: Let's extract the service from the remote WG into a core lib and reuse it. 
In this instance, we can override the `sign` method to limit the signing provider.

mjuchli-da (dismissed):

mjuchli-da (dismissed):

wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
mjuchli-da: We agreed to do it in a follow-up.

mjuchli-da (approved):

mjuchli-da (approved):

wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
```diff
@@ -0,0 +1,187 @@
...
+                `Unsupported signing provider: ${wallet.signingProviderId}`
+            )
+        }
+
+        return this.signWithWalletKernel(authContext.userId, wallet, signParams)
+    }
+
+    public async execute(
+        userId: string,
+        wallet: Wallet,
```
mateuszpiatkowski-da, line 50: I think that `wallet` already has `userId` so there is no need to pass it as an extra arg

mateuszpiatkowski-da (approved):

wallet-gateway/extension/entrypoints/background/user/transaction-service.ts
fayi-da: We're refactoring this whole service later

PR #2364, opened 2026-08-25
perf(example-portfolio): speed up the portfolio e2e job

### LINKED ISSUE #2363: portfolio-e2e is the slowest job in CI (21 min)
`portfolio-e2e` takes 21 minutes and runs twice per PR, once for devnet and once
for mainnet. It sets the wall clock of the whole run: the next slowest job is the
build, at 11 minutes.

Numbers below are marked **[CI]** when they come from a CI run (workflow run
32298657718, job `portfolio-e2e (devnet)`, read off the Playwright HTML report it
uploads) and **[local]** when they come from running the suite on a dev machine.

## Where the time goes [CI]

- **15 min running the tests.** The 18 tests run one at a time, and most of that
  is waiting on the ledger, not using CPU.
- **4 min starting LocalNet**, of which 2.5 are waiting for splice to become
  healthy.

Inside those 15 minutes, roughly:

- 200s waiting for the validator automation to accept preapprovals (~50s each,
  paid 4 times).
- 106s waiting for the ledger to confirm after each approval.
- 310s driving the wallet gateway's web UI to create parties and pick a primary
  wallet, none of which is what these tests assert.

## What is worth doing

In order of value per effort:

1. **Serve the built dApp instead of the Vite dev server.** In dev the browser
   pulls hundreds of unbundled modules per page. **[local]** 19% off the suite,
   measured on the transfers spec against the same LocalNet.
2. **Run the tests on parallel workers.** Since they mostly wait, this overlaps
   the waiting. **[local]** two workers took a full run from 11.6 to 6.5 min with
   per-step latencies unchanged; four workers was worse than two (8.8 min, more
   flakes) because Canton alone took 3 cores. Blocked today by shared state: the
   gateway scopes wallets per user and every test connects as the same one.
3. **Have the test helpers create wallets through the wallet gateway's User API
   instead of driving its web UI**, which costs a page load per call. **[local]**
   scaffolding on the transfers spec went from 79s to 37s.

Not worth chasing: the two thirds of the time that are Canton and splice doing
real work. That is only addressed by overlapping the waits, i.e. point 2.

## Caveat on the local numbers

The dev machine has 8 cores; CI runners have 4, and Canton plus splice already
take about 2 of them before any test starts. Expect the parallel gain to be
smaller on CI, and validate there before raising the worker count.

nicosampler:
Closes #2363.

`portfolio-e2e` takes 21 minutes and runs twice per PR, so it sets the wall clock
for the whole run. Three independent commits:

b77d8573 serves the dApp from the production build instead of the Vite dev server.
In dev the browser pulls hundreds of unbundled modules per page, which shows up on
every screen the tests open; from the build, each one loads in about half the time.
The build already happened in CI, it just was not in the artifact the e2e jobs
download.

43436991 runs the tests on two workers. They mostly wait on the ledger rather than
use CPU, so overlapping them is nearly free: a full local run went from 11.6 to 6.5
minutes with per-step latencies unchanged.

What kept them serial was shared state: the gateway scopes wallets, and which one is
primary, per user, and every test connected as the same user. Each worker now
connects as its own, derived from the Playwright parallel index. That id also travels
to the participant as the token subject, and Canton rejects subjects it does not
know, so a new global setup creates those users first.

fefe8911 changes how the tests set up their wallets. They used to create parties and
pick the primary wallet by clicking through the gateway's web UI, which meant
reloading its parties page 78 times across the suite, about 1.5s each, for something
no test actually asserts. The helpers now call the gateway's User API directly, the
same JSON-RPC methods that UI uses under the hood. Approving transactions still goes
through the UI, since that is the integration under test. On the transfers spec,
setup went from 79s to 37s.

The numbers come from a dev machine with 8 cores. CI runners have 4, and Canton plus
splice take about 2 before any test starts, so expect less there. Four workers was
already worse than two locally.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,7 @@ jobs:
                       core/**/dist/
                       sdk/**/dist/
                       wallet-gateway/**/dist/
+                      examples/portfolio/dist/
                       damljs
                       .nx/cache
                       .nx/workspace-data
--- a/core/wallet-test-utils/package.json
+++ b/core/wallet-test-utils/package.json
@@ -16,7 +16,11 @@
         }
     },
     "dependencies": {
+        "@canton-network/core-ledger-client": "workspace:^",
+        "@canton-network/core-rpc-transport": "workspace:^",
         "@canton-network/core-types": "workspace:^",
+        "@canton-network/core-wallet-auth": "workspace:^",
+        "@canton-network/core-wallet-user-rpc-client": "workspace:^",
         "@canton-network/wallet-sdk": "workspace:^",
         "@playwright/test": "^1.61.1",
         "pino": "^10.3.1",
--- a/core/wallet-test-utils/src/gateway-user-api.ts
+++ b/core/wallet-test-utils/src/gateway-user-api.ts
@@ -1,0 +1,133 @@
+// Copyright (c) 2025-2026 Digital Asset (Switzerland) GmbH and/or its affiliates. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
+
+import { HttpTransport } from '@canton-network/core-rpc-transport'
+import UserApiClient from '@canton-network/core-wallet-user-rpc-client'
+
+/**
+ * Client for the wallet gateway User API.
+ *
+ * Lets a test create wallets by calling the wallet directly, instead of driving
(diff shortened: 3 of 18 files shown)


rukmini-basu-da:
Thanks for your contribution, @nicosampler! Could you please fix the merge conflicts?

PR #2372, opened 2026-08-26
feat: make hashing scheme configurable

rukmini-basu-da:


wallet-gateway/remote/src/env.ts
```diff
@@ -38,7 +44,8 @@ export class Env {
         Env.get('BITGO_API_URL', { fallback })
     static BITGO_ENTERPRISE_ID = () => Env.get('BITGO_ENTERPRISE_ID')
     static BITGO_COIN = () => Env.get('BITGO_COIN')
-
+    static HASHING_SCHEME_VERSION = (fallback: HASHING_SCHEME_VERSION) =>
```
pawelstepien-da: I think wallet gateway json config would be a better place for this than env var

wallet-gateway/remote/src/config/Config.ts
```diff
@@ -115,6 +115,21 @@ const bootstrapFromEnv = bootstrapConfigSchema.extend({
     ),
 })
 
+const hashingSchemeSchema = z
+    .object({
+        version: z
+            .enum([
+                'HASHING_SCHEME_VERSION_UNSPECIFIED',
```
pawelstepien-da: I see it in ledger api specs, but I think it just makes preparing and executing txes fail. 
Could you try it out in runtime and if it fails, then maybe we should just stick with V2 and V3?

rukmini-basu-da: makes sense to remove since we default to v3, updated to remove unspecified

pawelstepien-da (approved):

PR #2379, opened 2026-08-30
feat: default wallet picker to in-page modal

### LINKED ISSUE #1678: Review and improve wallet-picker
The wallet picker was recently updated to support wallet connect. The code is somewhat sloppy and need to be cleaned up. Some ideas include:

- Clean up dead and unused css code
- Refactor inline styles and use css tokens/vars where appropriate after refactor
- connectedCallback() adds an inline message handler for `wc-uri`, but disconnectedCallback() does not remove it. Might cause issues
-  wcUri and wcQrDataUrl are never cleared when retrying or returning to the list, so stale QR/URI datamight show up in any later attempts. Verify and clear fields appropriately if true

singhyash05:
## Why

dApps should open the wallet picker as an in-page modal by default. Popup stays available with `setWalletPicker(pickWallet)`. Continues #2237.

## What we did

Took the modal from draft shape into something we could ship as the SDK default, then rewired everything that assumed popup.

- Built the production modal UI (~940 lines) plus styles, and Chromium unit tests (~735 lines) covering open, select, dismiss, error, and retry
- Refactored dapp-sdk picker plumbing: default is `pickWalletModal`; popup is opt-in; connected/error/retry/back split by modal vs popup; WalletConnect QR mounts in the modal
- Reworked Playwright helpers (`openWalletPicker`, modal host/row locators, WG connect) so e2e drives the new path instead of waiting on a picker popup
- Wrote `walletpicker-modal.spec.ts` for open / backdrop+back / live remote connect
- Rewrote `walletpicker.spec.ts` for modal error UX (in-modal alert, not popup "Try Again"). Hit a real lifecycle bug: after modal connect the WG login window closes, so WG popup logout no longer clears dApp Status. Switched that step to dApp Disconnect and fixed ping `App` / `useStatus` so Status clears on disconnect
- Aligned extension `ping-page.ts` with the modal default

## Ledger port (`2975`)

Localnet exposes JSON Ledger API on `2975`, not `5003`. Updated `wallet-gateway/test/config.json` and `Config.test.ts` so bootstrap/connect e2e hits the real localnet port (same as `example-config`).

## Testing (what we actually ran)

- Clean / build / unit on touched packages (`core-wallet-ui-components`, `dapp-sdk`, `core-wallet-test-utils`, `wallet-gateway-remote`)
- Full monorepo: `pnpm clean:all` → `build:all --skip-nx-cache` → `test:all`
- E2e against live localnet + `start:all`. First connect runs failed after remote `clean` wiped sqlite (empty `networks`). Reseeded DB, confirmed 6 networks, re-ran until green:

```bash
CI=1 pnpm nx playwright:e2e @canton-network/example-ping -- \
  tests/walletpicker-modal.spec.ts tests/walletpicker.spec.ts --project=chromium
```

Result: 4/4 passed.

## Test plan

- [x] Touched-package units + `test:all`
- [x] Modal + picker e2e (chromium), 4/4
- [ ] Optional: full ping chromium e2e
- [ ] Manual: default modal; `setWalletPicker(pickWallet)` restores popup

### DIFF
--- a/core/wallet-test-utils/src/wallet-gateway.ts
+++ b/core/wallet-test-utils/src/wallet-gateway.ts
@@ -2,7 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { expect, Locator, Page, test } from '@playwright/test'
-import { openWalletPicker } from './wallet-picker.js'
+import {
+    openWalletPicker,
+    WALLET_PICKER_MODAL_HOST,
+    walletPickerModalRowByTitle,
+    type WalletPickerSurface,
+} from './wallet-picker.js'
 
 export interface NetworkFormInput {
     id: string
@@ -121,14 +126,13 @@ export class WalletGateway {
                 'the dApp should offer a way to connect a wallet'
             ).toBeVisible()
 
-            const pickerPopup = await openWalletPicker(
-                dapp.dappPage,
-                connectButton
+            const picker = await openWalletPicker(dapp.dappPage, connectButton)
+
+            await this.selectFromWalletPicker(picker, args.customURL)
+
+            const popup = await this.waitForConnectFormPopup(
+                picker.kind === 'popup' ? picker.page : undefined
             )
-
-            await this.selectFromWalletPicker(pickerPopup, args.customURL)
-
-            const popup = await this.waitForConnectFormPopup(pickerPopup)
             const selectNetwork = popup.getByLabel('Select a network')
             await expect(
                 selectNetwork,
@@ -482,23 +486,46 @@ export class WalletGateway {
     }
 
     private async selectFromWalletPicker(
-        popup: Page,
+        picker: WalletPickerSurface,
         customURL?: string
     ): Promise<void> {
-        if (customURL !== undefined) {
-            const customUrlInput = popup.locator('.custom-url-input')
-            await customUrlInput.waitFor({ state: 'visible', timeout: 3000 })
-            await customUrlInput.fill(customURL)
-            await popup.locator('.btn-add').click()
+        const { page, kind } = picker
+
+        if (kind === 'modal') {
+            if (customURL !== undefined) {
+                await walletPickerModalRowByTitle(page, 'Remote Wallet').click()
+                const host = page.locator(WALLET_PICKER_MODAL_HOST)
+                const urlInput = host.getByLabel('Remote Wallet URL')
+                await urlInput.waitFor({ state: 'visible', timeout: 5_000 })
+                await urlInput.fill(customURL)
+                await host.locator('.gateway-connect-button').click()
+                return
+            }
+
+            const firstWallet = page
+                .locator(WALLET_PICKER_MODAL_HOST)
+                .getByRole('button')
+                .filter({ hasNotText: /^Remote Wallet/ })
+                .first()
+            await firstWallet.waitFor({ state: 'visible', timeout: 5_000 })
+            await firstWallet.click()
             return
         }
 
-        const walletCard = popup.locator('.wallet-card').first()
+        if (customURL !== undefined) {
+            const customUrlInput = page.locator('.custom-url-input')
+            await customUrlInput.waitFor({ state: 'visible', timeout: 3000 })
+            await customUrlInput.fill(customURL)
+            await page.locator('.btn-add').click()
+            return
+        }
+
+        const walletCard = page.locator('.wallet-card').first()
         await walletCard.waitFor({ state: 'visible', timeout: 3000 })
         await walletCard.click()
     }
 
-    private async waitForConnectFormPopup(initialPopup: Page): Promise<Page> {
+    private async waitForConnectFormPopup(initialPopup?: Page): Promise<Page> {
         const hasConnectForm = async (page: Page): Promise<boolean> => {
             try {
                 await page
@@ -518,13 +545,10 @@ export class WalletGateway {
             }
         }
 
-        // Pre-fix behavior: picker page itself transitions to connect form.
-        if (await hasConnectForm(initialPopup)) {
+        if (initialPopup && (await hasConnectForm(initialPopup))) {
             return initialPopup
         }
 
-        // New behavior: picker may close and the wallet connect form appears in
-        // a fresh popup. Poll the tracked popup first to avoid races.
         for (let i = 0; i < 20; i++) {
             const popup = this._popup
             if (popup && !popup.isClosed() && (await hasConnectForm(popup))) {
--- a/core/wallet-test-utils/src/wallet-picker.ts
+++ b/core/wallet-test-utils/src/wallet-picker.ts
@@ -3,11 +3,66 @@
 
 import type { Locator, Page } from '@playwright/test'
 
+export const WALLET_PICKER_MODAL_HOST = '[data-swk-wallet-picker-modal]'
+
+export type WalletPickerSurface = {
+    kind: 'modal' | 'popup'
+    page: Page
+    dappPage: Page
+}
+
 export async function openWalletPicker(
     dappPage: Page,
     connectButton: Locator
-): Promise<Page> {
-    const pickerPopup = dappPage.waitForEvent('popup')
+): Promise<WalletPickerSurface> {
+    const modalHost = dappPage.locator(WALLET_PICKER_MODAL_HOST)
+
+    const popupRace = dappPage
+        .waitForEvent('popup', { timeout: 15_000 })
+        .then((page): WalletPickerSurface => ({
+            kind: 'popup',
+            page,
+            dappPage,
+        }))
+        .catch(() => null)
+
+    const modalRace = modalHost
+        .waitFor({ state: 'attached', timeout: 15_000 })
+        .then((): WalletPickerSurface => ({
+            kind: 'modal',
+            page: dappPage,
+            dappPage,
+        }))
+        .catch(() => null)
+
     await connectButton.click()
-    return pickerPopup
+
+    const surface = await Promise.race([popupRace, modalRace])
+    if (!surface) {
+        throw new Error(
+            'wallet picker did not open as an in-page modal or a popup window'
+        )
+    }
+
+    if (surface.kind === 'modal') {
+        await modalHost
+            .locator('[role="dialog"]')
+            .waitFor({ state: 'visible', timeout: 5_000 })
+    }
+
+    return surface
+}
+
+export function walletPickerModalHost(dappPage: Page): Locator {
+    return dappPage.locator(WALLET_PICKER_MODAL_HOST)
+}
+
+export function walletPickerModalRowByTitle(
+    dappPage: Page,
+    title: string
+): Locator {
+    const escaped = title.replace(/[.*+?^${}()|[\]\\]/g, '\\$&')
+    return walletPickerModalHost(dappPage).getByRole('button', {
+        name: new RegExp(`^${escaped}(\\s+${escaped})?$`),
+    })
 }
--- a/core/wallet-ui-components/src/styles/modal.ts
+++ b/core/wallet-ui-components/src/styles/modal.ts
@@ -41,3 +41,704 @@ export const modalStyles = css`
         }
     }
 `
+
+const LIGHT_TOKENS = `
+    --accent: #111111;
+    --accent-hover: #000000;
+    --accent-contrast: #ffffff;
+    --accent-soft: rgba(17, 17, 17, 0.06);
+    --border: #ececf3;
+    --border-hover: #d3d3e0;
+    --surface: #ffffff;
+    --surface-2: #f4f4f7;
+    --surface-hover: #ebebf0;
+    --text: #373737;
+    --text-muted: #6b7280;
+    --icon-muted: #999999;
+    --scrollbar: #d4d4dc;
+    --success: #4e9e73;
+    --success-soft: rgba(78, 158, 115, 0.14);
+    --danger: #dc2626;
+    --danger-text: #b91c1c;
+    --danger-soft: rgba(220, 38, 38, 0.08);
+    --danger-soft-hover: rgba(220, 38, 38, 0.1);
+    --backdrop: rgba(71, 88, 107, 0.24);
+    --fade-shadow: rgba(15, 23, 42, 0.18);
+    --shadow-modal:
+        0 12px 32px -18px rgba(15, 23, 42, 0.12),
+        0 2px 8px -6px rgba(15, 23, 42, 0.06);
+`
+
+const DARK_TOKENS = `
+    --accent: #f4f4f6;
+    --accent-hover: #e2e2e6;
+    --accent-contrast: #16171b;
+    --accent-soft: rgba(255, 255, 255, 0.08);
+    --border: #2b2d34;
+    --border-hover: #3b3e47;
+    --surface: #17181c;
+    --surface-2: #202228;
+    --surface-hover: #2a2c33;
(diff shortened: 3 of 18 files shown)


mjuchli-da (commented):
Hi @singhyash05 

Thanks for the PR, the new modal looks very nice!

I have a few high-level things:
- Could you add both "connect" buttons to the example-ping dapp?
- The order of the items listed is different from the popup picker. I believe in the popup we have the order: 1. default remotes (defaultAdapters) 2. additional by dapp (additionalAdapters) 3. announced (installed) 4. suggested (not-installed, if enableSuggestedWallets)
- WalletConnect entry opens a blank tab
- WalletConnect URL/QR is not showing
- CI does not pass

We (@alexmatson-da and I) will provide a more detailed review shortly.

wallet-gateway/test/config.json
```diff
@@ -76,7 +76,7 @@
                     "clientSecret": "service-account-secret"
                 },
                 "ledgerApi": {
-                    "baseUrl": "http://127.0.0.1:5003"
+                    "baseUrl": "http://127.0.0.1:2975"
```
alexmatson-da, line 79: I don't see a reason for these ports to change: these networks are configured to use a Canton node started outside of localnet (`pnpm start:canton`), and has a default ledger API port of 5003.

There is already a WG network defined for localnet instances in the config, that should be the one selected on the login screen (when using `pnpm start:localnet`)

core/wallet-test-utils/src/wallet-picker.ts
```diff
@@ -3,11 +3,66 @@
 
 import type { Locator, Page } from '@playwright/test'
 
+export const WALLET_PICKER_MODAL_HOST = '[data-swk-wallet-picker-modal]'
+
+export type WalletPickerSurface = {
+    kind: 'modal' | 'popup'
```
alexmatson-da, line 9: do we really want to support both `modal` and `popup` style pickers? IMO I don't really see a reason to continue to provide the popup as an option, I'd be in favor of going full modal-only

cc @mjuchli-da @joel-da for thoughts

sdk/dapp-sdk/src/sdk.ts
```diff
@@ -408,7 +454,42 @@ export class DappSDK {
...
+                                new CustomEvent<WalletPickerEntry>('attempt', {
+                                    detail: retrySelection,
+                                })
+                            )
+                        } catch (retryError) {
+                            cleanup()
+                            reject(retryError)
+                        }
+                        return
+                    }
```
alexmatson-da, line 492: I wonder if using an `AbortController` [here](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) can help clean up the code and make it more concise/idiomatic. I dont have much first-hand experience with this API, so I leave it to you to see if its worth the gain.

My basic idea/sketch of what it might look like:

```typescript
const abortController = new AbortController()

// when we create the wallet picker, we give it an optional abortController
export async function pickWalletModal(
    entries: WalletPickerEntry[],
    abortController?: AbortController
): Promise<WalletPickerResult> {
    active?.destroy()
    active = new WalletPickerModalController(entries, abortController)
    return active.awaitSelection()
}

// inside the wallet picker implementation, we call abort when the user clicks the back button
awaitBack(): Promise<void> {
    if (this.backRequested) {
        this.backRequested = false
        
        // Invoke the abort signal here
        this.abortController.abort()
    }
    return new Promise<void>((resolve) => {
        this.backPending = { resolve }
    })
}

// ----------------------
// later, when we call discovery.connect, we pass in the associated *signal* from the same instance (also optional)
discovery.connect(targetId, { signal: abortController.signal })

// inside the implementation, tear down 
async connect(providerId?: ProviderId | undefined, signal?: AbortSignal | undefined): Promise<void> {
    let targetId = providerId
    // ...

    signal.onabort = () => {
      this.disconnect()
      // ... other cleanup
    }
}
```

may not be totally accurate, but thats the general idea

singhyash05:
Thanks @mjuchli-da @alexmatson-da addressed the review feedback on `c1f9391` (rebased on latest main).

### Marc
- **Both connect buttons** on example-ping: `connect to Wallet` (modal default) + `connect (popup)` via `setWalletPicker(pickWallet)`
- **List order** matches popup / discovery: defaultAdapters → additionalAdapters → announced → saved custom remotes → Remote Wallet → suggested
- **WalletConnect blank tab**: `openPopupForUri` defaults to `false`
- **WC URL/QR in modal**: adapter calls `setWalletPickerModalWalletConnectUri`
- **CI / e2e**: local picker e2e green with `pnpm start:canton` (ledger `:5003`) — 4/4 on `walletpicker-modal` + `walletpicker` chromium

### Alex
- **Ports**: reverted non-LocalNet networks to `5003`; LocalNet stays `2975` (as discussed)
- **Firefox badge**: only label matching platform; otherwise `Get` (no “Get for chrome” on Firefox)
- **Custom WG localStorage**: persist + deletable recent entries (unit coverage)
- **Status clear**: ping `useStatus` / `useConnect` clear Status on disconnect / `statusChanged` (WG logout path). Note: after modal connect the login popup closes, so e2e uses dApp Disconnect; please re-check WG Logout via Open wallet if you want that exact path
- **Modal-only**: kept popup as opt-in for now (matches Marc’s dual-button request). Happy to drop popup in a follow-up if you + @joel-da agree
- **AbortController**: left as follow-up — current `Promise.race` + back waiter works; can refactor if you want that cleanup in this PR

Happy to take another pass once CI reports.

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -533,27 +557,42 @@ export class WalletGateway {
...
 
-        // Pre-fix behavior: picker page itself transitions to connect form.
-        if (await hasConnectForm(initialPopup)) {
-            return initialPopup
+        if (initialPopup && !initialPopup.isClosed()) {
+            if (await hasConnectForm(initialPopup)) {
+                return initialPopup
+            }
+            // Popup may still be loading the connect form.
+            for (let i = 0; i < 40; i++) {
```
mateuszpiatkowski-da: This seems off for me. Could you elaborate why 40 is the limit here and guarantee that it is enough for the connect form to show up?

Additionally, is there a reason why we don't simply use [playwright API](https://playwright.dev/docs/api/class-locatorassertions#locator-assertions-to-be-visible) to supply the utils with the timeout?

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -533,27 +557,42 @@ export class WalletGateway {
...
+            }
         }
 
-        // New behavior: picker may close and the wallet connect form appears in
-        // a fresh popup. Poll the tracked popup first to avoid races.
-        for (let i = 0; i < 20; i++) {
-            const popup = this._popup
-            if (popup && !popup.isClosed() && (await hasConnectForm(popup))) {
-                return popup
+        for (let i = 0; i < 40; i++) {
```
mateuszpiatkowski-da: same as https://github.com/canton-network/wallet/pull/2379/changes#r4072673476

core/wallet-test-utils/src/wallet-picker.ts
mateuszpiatkowski-da: instead of forcing the user to pass this prop we can instead introduce a boolean like `isModal = false` that will tell us whether we deal with a modal or a popup (which is by default)

core/wallet-test-utils/src/wallet-picker.ts
```diff
@@ -3,11 +3,66 @@
...
+    if (surface.kind === 'modal') {
+        await modalHost
+            .locator('[role="dialog"]')
+            .waitFor({ state: 'visible', timeout: 5_000 })
+    }
+
+    return surface
+}
+
+export function walletPickerModalHost(dappPage: Page): Locator {
```
mateuszpiatkowski-da: I don't see a value in such function (it's enough to export WALLET_PICKER_MODAL_HOST IMO)

core/wallet-test-utils/src/wallet-picker.ts
```diff
@@ -3,11 +3,66 @@
...
+    return dappPage.locator(WALLET_PICKER_MODAL_HOST)
+}
+
+export function walletPickerModalRowByTitle(
```
mateuszpiatkowski-da: this seems like an overkill, because we're basically using an overly complicated function in one place to get to a button. Please simplify the logic and use it directly in https://github.com/singhyash05/wallet/blob/c1f93915d9d53a22993ffa826506674c976336d0/core/wallet-test-utils/src/wallet-gateway.ts#L508, unless you would like to reuse it in a foreseeable future (in such case please share more information)

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    --backdrop: rgba(0, 0, 0, 0.5);
+    --fade-shadow: rgba(0, 0, 0, 0.5);
+    --shadow-modal:
+        0 12px 32px -18px rgba(0, 0, 0, 0.4),
+        0 2px 8px -6px rgba(0, 0, 0, 0.24);
+`
+
+export const walletPickerModalCss = `
+:host {
+    all: initial;
```
mateuszpiatkowski-da: Could you share some insight into why we need to do this?

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+.discovery-modal-backdrop {
+${LIGHT_TOKENS}
+    position: fixed;
+    inset: 0;
+    background: var(--backdrop);
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    padding: 16px;
+    z-index: 2147483000;
```
mateuszpiatkowski-da: That's a weird value for a z-index. The problem is that it's really easy to mess up the values if no order is introduced. Could you please change the value so that it reflects some proper significance while eliminating randomness? 

```suggestion
    z-index: var(--z-index-modal);
```
Where
```
  --z-index-modal: 10
```

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
         }
     }
 `
+
+const LIGHT_TOKENS = `
```
mateuszpiatkowski-da: There's no reason to encapsulate the variables like that. I'd appreciate it if you could move the css vars to a root component and make them public, so that all other components would have access to them

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    --accent: #111111;
+    --accent-hover: #000000;
+    --accent-contrast: #ffffff;
+    --accent-soft: rgba(17, 17, 17, 0.06);
```
mateuszpiatkowski-da: is there a reason as to why we're using both hex and rgba notations? AFAIK hex colors also have support for alpha values so that way we could stick to just one format

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    cursor: pointer;
+    color: var(--icon-muted);
+    width: 32px;
+    height: 32px;
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    border-radius: 50%;
+    transition:
+        background 0.15s ease,
```
mateuszpiatkowski-da: I would refrain from transitioning background since it's not an optimized way to animate stuff in CSS and stick to `transform` and `opacity` only. In our case we wish to animate from transparent background to a full color. Please make a modification to use opacity instead

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    --danger-soft: rgba(242, 109, 109, 0.12);
+    --danger-soft-hover: rgba(242, 109, 109, 0.16);
+    --backdrop: rgba(0, 0, 0, 0.5);
+    --fade-shadow: rgba(0, 0, 0, 0.5);
+    --shadow-modal:
+        0 12px 32px -18px rgba(0, 0, 0, 0.4),
+        0 2px 8px -6px rgba(0, 0, 0, 0.24);
+`
+
+export const walletPickerModalCss = `
```
mateuszpiatkowski-da: please use `css` from lit everywhere, it's difficult to review this without having an IDE support

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    border: 1px solid var(--border);
+    border-radius: 20px;
+    box-shadow: var(--shadow-modal);
+    width: min(87.4vw, 342px);
+    overflow: hidden;
+    animation: swkSlideUp 0.28s cubic-bezier(0.16, 1, 0.3, 1);
+}
+
+.discovery-modal-inner {
+    width: min(87.4vw, 342px);
```
mateuszpiatkowski-da: these values seem random... Have you inspected the component with different viewports?

EDIT: Now I see that this doesn't make sense as for lower viewports we use a media query: 

Image

Therefore, we never use the vw or vh values

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    max-height: min(86vh, 640px);
+    display: flex;
+    flex-direction: column;
+}
+
+@keyframes swkSlideUp {
+    from { transform: translateY(16px) scale(0.98); opacity: 0; }
```
mateuszpiatkowski-da: why use scale here?

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    border-radius: 4px;
+    border: 2px solid var(--surface);
+}
+
+.discovery-modal-view {
+    animation: swkViewFade 0.25s ease;
+}
+
+@keyframes swkViewFade {
+    from { opacity: 0; transform: translateY(4px); }
```
mateuszpiatkowski-da: why use translateY here?

core/wallet-ui-components/src/styles/modal.ts
```diff
@@ -41,3 +41,704 @@ export const modalStyles = css`
...
+    color: var(--danger-text);
+    font-size: 13px;
+    line-height: 1.4;
+}
+
+.wallet-picker-container {
+    display: flex;
+    flex-direction: column;
+    gap: 8px;
+    /* Show ~4.5 wallet rows (62px tall + 8px gap), then scroll. Only this
```
mateuszpiatkowski-da: Ideally I'd put css vars here in order to ensure that the values for gaps/heights are the same as in the max-height

examples/ping/src/hooks/useStatus.ts
```diff
@@ -30,29 +36,25 @@ export function useStatus(): {
...
+        if (!isConnected) {
+            return
+        }
+
+        const onStatusChanged = (s: sdk.dappAPI.StatusEvent) => {
+            setStatusEvent(s.connection?.isConnected ? s : undefined)
+        }
+
+        sdk.onStatusChanged(onStatusChanged).catch(() => {
+            // Client not ready — ignore; next connect will re-run this effect.
```
mateuszpiatkowski-da: What does this mean?

PR #2380, opened 2026-08-30
test: mock postMessage in popup unit tests

singhyash05:
## Why

`popup.open(url)` polls with `win.postMessage` every 500ms. The popup unit-test mock never defined `postMessage`, so Vitest can exit red with unhandled `TypeError: win.postMessage is not a function` even when all assertions pass (flake, worse under serial / slow runs).

## Change

One line in `popup.test.ts`: add `postMessage: vi.fn()` to `createMockPopupWindow()`.

## Test

```bash
pnpm nx test @canton-network/core-wallet-ui-components --skip-nx-cache
```


mjuchli-da (approved):
Thanks

PR #2381, opened 2026-08-31
fix: async approval add polling

rukmini-basu-da:
Adds a polling mechanism and a new status `awaiting-signature`:
image

1.pending: prepared command not sent anywhere
2.awaiting-signature: sent to signing provider but haven't received signature yet (this happens after we hit approve the first time)
3. signed: found signature on signing provider (so as soon as we go to this activity on the ui, it starts querying the signing driver for this tx if it's in awaiting-signature and then as soon as it finds it we update the sqllite status)
4. executed: happens after we hit approve the second timeand happy path if the execute against the ledger is successful
5. failed: as is, can happen if something goes wrong at any step but would also include if something fails on the execute step


core/wallet-store-inmemory/src/store-internal.ts
```diff
@@ -462,7 +456,8 @@ export class StoreInternal implements Store, AuthAware<StoreInternal> {
     async listAllPendingTransactions(): Promise<Array<Transaction>> {
```
pawelstepien-da, line 475: This is used by signing worker and I think it could break something there in condition that compares `refreshedTx.status !== 'pending'`, or at least result in unnecessary processing txes that are not `pending-signature`. That module will probably need to adjust to new status.

wallet-gateway/remote/src/user-api/controller.ts
```diff
@@ -1240,67 +1193,97 @@ export const userController = (
...
+                    status: tx.status,
+                    ...(tx.externalTxId && { externalTxId: tx.externalTxId }),
+                    ...(tx.failureReason && {
+                        failureReason: tx.failureReason,
+                    }),
+                }
+            }
+
+            //double check if I should pass in partyId and then do getWallets.find()
+            const wallet = await store.getPrimaryWallet()
```
pawelstepien-da: Reliance on primary party was there before, just in a different place. It would be better to know what wallet or at least signing provider initiated signing based on transaction row alone. I will think about it.

core/wallet-store-sql/src/store-sql.ts
```diff
@@ -697,6 +673,7 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
             }),
             ...(signedAt !== undefined && { signedAt }),
             ...(externalTxId !== undefined && { externalTxId }),
+            ...(failureReason !== undefined && { failureReason }),
```
pawelstepien-da, line 708: We need migration for that new field. Also we should migrate statuses. I think we can infer that a tx that was pending, if it has "external_tx_id", it will now be "pending-signing"

api-specs/openrpc-user-api.json
```diff
@@ -884,6 +874,33 @@
                 }
             }
         },
+        {
+            "name": "getTransactionStatus",
+            "description": "Polls the signing provider for a transaction awaiting external signature and udpated the store status. ",
```
alexmatson-da: (typo)

wallet-gateway/remote/src/web/frontend/approve/index.ts
```diff
@@ -104,45 +126,87 @@ export class ApproveUi extends BaseElement {
...
+            this.syncPolling()
+        }
+    }
+
+    private async handleApproveAction() {
+        if (this.status === 'signed') {
+            return this.handleExecute()
         }
+        return this.handleSign()
     }
```
alexmatson-da, line 209: UX-wise, I'm not sure i see a need to keep the two-click interaction... wdyt about calling `handleSign` immediately in the connectedCallback, so the request is sent in the background when the popup is first opened. Then the "Approve" button is deactivated during polling until the TX status is `signed`

core/wallet-store-sql/src/store-sql.ts
```diff
@@ -714,43 +691,58 @@ export class StoreSql implements BaseStore, AuthAware<StoreSql> {
...
+    ): Promise<boolean> {
         const userId = this.assertConnected()
         const network = await this.getCurrentNetwork()
         const existing = await this.getTransaction(transactionId)
         if (!existing) {
             throw new Error(`Transaction not found with id: ${transactionId}`)
         }
+        if (opts?.expectedStatus && existing.status !== opts.expectedStatus) {
+            return false
+        }
```
alexmatson-da, line 754: whats the idea behind expectedStatus / how will it be used?

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -260,6 +261,129 @@ export class TransactionService {
...
+            provider === SigningProvider.SECUROSYS
+                ? { txId: externalTxId }
+                : { userId, txId: externalTxId }
+
+        return driver.getTransaction(args).then(handleSigningError)
+    }
+
+    private async applySigningResult(
+        tx: Transaction,
+        wallet: Wallet,
```
alexmatson-da: `wallet` seems to be unused here

wallet-gateway/remote/src/user-api/controller.ts
alexmatson-da: I would not assume the primary wallet is used here, and would explicitly pick the wallet by partyId. The `sign` and `execute` methods in the controller also work this way

core/wallet-store-sql/src/store-sql.ts
rukmini-basu-da: it's currently used in the `applySigningResult`
```
            const applied = await this.store.setTransactionSigned(
                tx.id,
                now,
                signingResult.txId,
                { expectedStatus: tx.status }
            )

            if (!applied) {
                const current = await this.store.getTransaction(tx.id)
                return { status: current!.status }
            }


```

and I think this was when I was trying tackle the scenario where someone hits approve twice on the execute where the version approve click has already gone through the sign/execute flow but then another one is in flight so this would prevent the `setTransactionSigned` row from matching (since the status would be `executed`) and then the UI shows the transaction as completed

wallet-gateway/remote/src/user-api/controller.ts
rukmini-basu-da: sounds good, updated this by passing in the partyId to the transaction status params

core/wallet-store-inmemory/src/store-internal.ts
rukmini-basu-da: I've modified the signingWorker to now look at `refreshedTx.status !== 'pending' && refreshedTx.status !== 'awaiting-signature'` and modified the `signAndExecute` (updated the doc at the top to explain the lifecycle/what happens on each tick)

wallet-gateway/remote/src/web/frontend/approve/index.ts
rukmini-basu-da: As discussed, we would want the users to view the transaction (on the wg) before sending to signing provider because we don't want to risk having a signed blob that could be executed

core/wallet-store-sql/src/store-sql.ts
rukmini-basu-da: sounds good, added this to the migration scripts!

api-specs/openrpc-user-api.json
```diff
@@ -1219,6 +1246,33 @@
                 "format": "uuid",
                 "description": "The internal transaction identifier."
             },
+            "GetTransactionStatusResult": {
+                "type": "object",
```
alexmatson-da, line 1251: needs a `title` here

core/wallet-test-utils/src/wallet-gateway.ts
```diff
@@ -404,12 +409,17 @@ export class WalletGateway {
             await approveButton.click()
 
             if (opts?.isExternalSigning) {
+                //TODO: race condition where the poll already updated to signed. figure out a better assertion
+                // await this.expectActivityWithStatus(
+                //     commandId,
+                //     'awaiting-signature'
+                // )
```
alexmatson-da, line 416: maybe you can have it wait until its no longer `pending`?

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -210,30 +203,44 @@ export class TransactionService {
...
-        if (
-            !('signature' in signResult) ||
-            signResult.signature === undefined
-        ) {
-            throw new Error(
-                'Service account signing did not return a signature'
-            )
+            if (refreshed.status === 'awaiting-signature') {
+                return {
+                    status: 'pending',
```
alexmatson-da, line 224: why status `pending` and not `awaiting-signature`?

wallet-gateway/remote/src/ledger/transaction-service.test.ts
```diff
@@ -803,33 +913,43 @@ describe('TransactionService', () => {
 
         describe('external signing providers', () => {
             it.each([
-                SigningProvider.WALLET_KERNEL,
+                // SigningProvider.WALLET_KERNEL,
```
alexmatson-da: why disable?

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -248,24 +255,158 @@ export class TransactionService {
...
+            ...(failureReason && { failureReason }),
+        })
+
+        // dApp reports pending for anything not yet signed
+        // awaiting-signature can be a gateway UI internal distinction for polling
+        this.notifier.emit('txChanged', {
+            ...tx,
+            status: status === 'awaiting-signature' ? 'pending' : status,
+            externalTxId: signingResult.txId,
+        })
```
alexmatson-da, line 401: I dont quite see a reason to not send `awaiting-signature` to the dapp as well, but maybe add a TODO + tech debt issue for now

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -287,30 +428,16 @@ export class TransactionService {
...
-                .getTransaction({
-                    userId,
-                    txId: tx.externalTxId,
-                })
-                .then(handleSigningError)
-        } else {
-            signingResult = await driver
-                .signTransaction(signTransactionParams)
-                .then(handleSigningError)
-        }
```
alexmatson-da, line 311: why was this removed?

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -321,66 +448,26 @@ export class TransactionService {
             debug: { signingResult, tx },
         })
 
+        await this.applySigningResult(tx, signingResult)
```
alexmatson-da: do we need to handle the return value? I can't tell if the `if` statement should check `signingResult.status` or `appliedSigningResult.status` (or if it matters)?

rukmini-basu-da: good call, this guard should help if another request tries to sign or execute the same transaction concurrently (which ideally should be prevented by `loadPreparedTransactionForSigning`), but it's worth having to make sure the applied result and driver result agree on the status

wallet-gateway/remote/src/ledger/transaction-service.test.ts
rukmini-basu-da: fixed!

wallet-gateway/remote/src/ledger/transaction-service.ts
rukmini-basu-da: yeah, this is a bit weird because this value goes on as `SignResult` and the api spec has `SignResultPending` (where the status is a const: pending), so this would fail the schema validation. Kinda related to https://github.com/canton-network/wallet/pull/2381#discussion_r3981449843


I eventually wanted to add the `SigningResultAwaitingSignature` but figured that could happen in a follow up pr, but maybe it makes sense to add now. wdyt?

wallet-gateway/remote/src/ledger/transaction-service.ts
rukmini-basu-da: Moved the path to polling the provider for the signature to `refreshTransaction`. And also, `signWithDriver` is only called from `sign` now and `sign` calls `loadPreparedTransactionForSigning` first so any tx with an `externalTxId` throws before `signWithDriver` runs so the if(tx.externalTxId) would never be reachable

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -404,102 +497,204 @@ export class TransactionService {
...
+            this.logger.error(
+                { err: err, transactionId: transaction.id },
+                `Ledger rejected the submission`
+            )
 
-        logDynamically(this.logger, 'Externally signed execution result', {
-            info: { transactionId: transaction.id },
-            debug: { result, transaction, executeParams, userId },
-        })
+            await this.store.setTransactionStatus(transaction.id, 'failed', {
```
pawelstepien-da, line 697: I'm not sure we should always set it to failed in case of any error. 
For example if there was temporary loss of connection with particpant at the moment /executeAndWait happens, a signed transaction that would execute otherwise will be impossible to execute forever because of status failed. Perhaps we should at least narrow it down to canton errors? We have `isJsCantonError` method somewhere. WDYT?

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -248,24 +255,158 @@ export class TransactionService {
...
+        }
+
+        const status =
+            signingResult.status === 'pending' ? 'awaiting-signature' : 'failed'
+        const failureReason =
+            status === 'failed'
+                ? `Signing provider returned status: ${signingResult.status}`
+                : undefined
+
+        await this.store.setTransactionStatus(tx.id, status, {
```
pawelstepien-da: Should we pass `expectedStatus` also here? Like with `setTransactionSigned`. Just to be sure no concurrent requests mess things up.

wallet-gateway/remote/src/web/frontend/approve/index.ts
```diff
@@ -206,6 +309,13 @@ export class ApproveUi extends BaseElement {
...
                     : ''
             }
+            ${
+                this.failureReason
+                    ? html`<div class="alert alert-warning" role="alert">
+                          ${this.failureReason}
+                      </div>`
+                    : ''
+            }
             <wg-transaction-detail
```
pawelstepien-da, line 335: Inside that component, in case of status awaiting-signature, the buttons don't render. I think we should be able to delete transaction in that status. WDYT?

```
private renderActionButtons() {
        if (this.status !== 'pending' && this.status !== 'signed') {
            return nothing
        }
```

wallet-gateway/remote/src/web/frontend/approve/index.ts
```diff
@@ -215,11 +325,13 @@ export class ApproveUi extends BaseElement {
                 .createdAt=${this.createdAt}
                 .signedAt=${this.signedAt}
                 .origin=${this.origin}
+                .externalTxId=${this.externalTxId}
                 .backHref=${toRelHref(ACTIVITIES_PAGE_REDIRECT)}
                 .isApproving=${this.isApproving}
                 .isDeleting=${this.isDeleting}
+                .isSigning=${this.isSigning}
```
pawelstepien-da, line 348: This property doesn't seem to do anything inside that component.

api-specs/openrpc-user-api.json
```diff
@@ -772,12 +772,7 @@
                                 "type": "string"
                             }
                         },
-                        "required": [
-                            "transactionId",
-                            "signature",
```
pawelstepien-da, line 777: Should we just remove `signature` and `signedBy`, instead of making it optional? It seems to not be used anymore.

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -248,24 +255,158 @@ export class TransactionService {
...
+    ): Promise<Exclude<GetTransactionResult, SigningError>> {
+        const provider = wallet.signingProviderId as SigningProvider
+        const signingProvider = this.signingDrivers[provider]
+        if (!signingProvider) {
+            throw new Error(`No driver found for provider ${provider}`)
+        }
+
+        const controllerId =
+            provider === SigningProvider.BLOCKDAEMON
+                ? authContext.email
```
pawelstepien-da, line 338: I wonder if we should throw in case email is missing for blockdaemon, like we do in sign
```
                if (!authContext.email) {
                    throw new Error(
                        'Email is required for Blockdaemon wallet allocation'
                    )
                }
```

wallet-gateway/remote/src/ledger/transaction-service.ts
pawelstepien-da: This SignResult is about status of signTransaction request in signing provider, not wallet gateway transaction. It basically means that transaction was created, is not yet signed, not yet failed, nor rejected. Wallet Gateway prepares transaction and adds it in db (pending), after singing provider was signTransaction was called it becomes pending-signature. Additionally signing drivers interface has separate failed and rejected, while GW only has failed. So WG tx status, and signing provider signTransaction status are separate enums and it's correct to check for pending there. If that's not clear, I'm open to renaming pending -> pending-signature in drivers interface, but this would be need to update many places.

wallet-gateway/remote/src/web/frontend/approve/index.ts
rukmini-basu-da: as discussed offline not required

wallet-gateway/remote/src/ledger/transaction-service.ts
rukmini-basu-da: Added an extended retry to the `ledgerPostWithRetry` so it retries for 1 minute (instead of 15 seconds) for the case of intermittent network failures

pawelstepien-da (approved):

PR #2392, opened 2026-09-02
fix: verify self signed token against client secret in each request

pawelstepien-da:


wallet-gateway/remote/src/auth/jwt-auth-service.ts
```diff
@@ -1,11 +1,113 @@
...
+    logger: Logger
+): Promise<AuthContext | undefined> {
+    const tokenAudiences = normalizeAudienceClaim(decoded.aud)
+    if (tokenAudiences.length === 0) {
+        logger.warn('JWT does not contain an audience claim')
+        return undefined
+    }
+
+    const networks = await store.listNetworks()
+    const candidates = networks.filter(
```
mjuchli-da: Instead of considering multiple candidates (which may again impose safety concerns), we may just adhere to the standard, which is to look up the network by `jwt.issuer`.

Actually, this should exactly follow (reuse) the same behaviour as in `jwtAuthService.verifyToken`:
```ts
            const idp = idps.find((i) => i.issuer === iss)
            const networks = await store.listNetworks()
            const networksForIdp = networks.filter(
                (n) => n.identityProviderId === idp.id
            )
            const expectedAudiences = networksForIdp
                .map((n) => n.auth.audience)
                .filter((aud): aud is string => aud !== undefined && aud !== '')
...
```

wallet-gateway/remote/src/auth/jwt-auth-service.ts
```diff
@@ -1,11 +1,113 @@
...
+            'No self-signed networks match the JWT audience'
+        )
+        return undefined
+    }
+
+    for (const network of candidates) {
+        try {
+            const { payload } = await jwtVerify(
+                jwt,
+                new TextEncoder().encode(network.auth.clientSecret),
```
mjuchli-da: This is the main difference compared to IDP tokens that come with a jwks config set.

wallet-gateway/remote/src/auth/jwt-auth-service.ts
mjuchli-da: Only in the case of self-signed, we would then also filter the **network** by **issuer** (not just the idp.issuer) [i think]

mjuchli-da (changes requested):
As discussed via meeting

wallet-gateway/remote/src/auth/jwt-auth-service.ts
pawelstepien-da: When using user-api selfSignedToken, `idp.issuer` is used as `iss` claim, not `network.auth.issuer`, so this wouldn't allow differentiating between networks using same IDP.

pawelstepien-da: To always be able to know which specific network self-signed token was minted for, I made user-api selfSignedAccessToken add networkId to claims as `kid`, which stands for id of a key that was used to sign the token, and in case of self-signed, network and signing key (clientSecret) is 1:1.

rukmini-basu-da (approved):

mjuchli-da (approved):
LGTM!

PR #2393, opened 2026-09-02
fix: refactor transaction-service

alexmatson-da:
This PR refactors the transaction-service to use a common signing function relying on the driver interface


core/signing-fireblocks/src/index.ts
```diff
@@ -86,16 +86,27 @@ export default class FireblocksSigningDriver implements SigningDriverInterface {
                 // TODO: validate transaction here
 
                 try {
+                    const txHashHex = Buffer.from(
+                        params.txHash,
+                        'base64'
+                    ).toString('hex')
+
```
alexmatson-da: fireblocks deals with hex encoded values where the WG always deals in base64... this moves the encoding wrangling directly into the signing driver to abstract it away from WG concerns (which is exactly why the signing driver abstraction exists in the first place ;D)

pawelstepien-da: Very good call!
Besides transaction-service, this base64 <> hex juggling for fireblocks `driver.signTransaction()` is also in party-allocation-service, for topology transaction signature. Also in that driver `getTransaction` should also return base64 signature instead of hex. This should fix e2e test for fireblocks.

Additionally `getKeys` also returns hex format for `publicKey`, and it's converted to base64 when creating a fireblocks party in `fireblocks-allocator.ts`. This could also be abstracted inside the driver.

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -315,190 +320,15 @@ export class TransactionService {
...
-
-    private async signWithFireblocks(
+    private async signWithDriver(
+        driver: ReturnType<SigningDriverInterface['controller']>,
         userId: UserId,
         wallet: Wallet,
-        signParams: SignParams
+        tx: Transaction,
+        signTransactionParams: Parameters<
+            ReturnType<SigningDriverInterface['controller']>['signTransaction']
```
pawelstepien-da: Is it possible to import the type directly?

core/signing-fireblocks/src/index.ts
pawelstepien-da: Also `driver.getTransactions` would still return hex signature, but we don't use this method in WG.

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -87,34 +107,54 @@ export class TransactionService {
                         'Email is required for Blockdaemon wallet allocation'
                     )
                 }
-                return this.signWithBlockdaemon(
+                return this.signWithDriver(
+                    driver,
                     authContext.email,
```
pawelstepien-da, line 126: Currently (in this PR) that email isn't used inside blockdaemon driver, but rather `userId` param from constructor. Before in `signWithBlockdaemon` we were creating driver controller instance (with email as userId). Now it uses the one passed as first arg, which has the actual userId instead of email. Against real blockdaemon instance it fails if email is not matched with bd user owning API key.

I will update mocked BD API for e2e to catch this issue in the future https://github.com/canton-network/wallet/issues/2394

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -513,248 +343,18 @@ export class TransactionService {
...
-                .then(handleSigningError)
-        }
-
-        const now = new Date()
-
-        logDynamically(this.logger, 'Securosys signing result', {
-            info: { transactionId: tx.id, status: signingResult.status },
-            debug: { signingResult, tx },
-        })
+        // throw new Error(JSON.stringify(signingResult))
```
pawelstepien-da: leftover

wallet-gateway/remote/src/ledger/transaction-service.ts
```diff
@@ -513,248 +343,18 @@ export class TransactionService {
...
-                    },
-                })
+                .signTransaction(signTransactionParams)
                 .then(handleSigningError)
         }
 
         const now = new Date()
 
-        logDynamically(this.logger, 'Dfns signing result', {
+        logDynamically(this.logger, 'Driver signing result', {
```
mjuchli-da, line 315: We should indicate with every log which driver is used; otherwise, debugging can be really hard.

wallet-gateway/remote/src/ledger/transaction-service.ts
alexmatson-da: nice catch! luckily I think this is the reason ping-e2e failed, so at least we have _some_ test coverage

alexmatson-da: nvm, looks like the failure was the base64 thing (as you mentioned in the other comment ;D)

core/signing-fireblocks/src/index.ts
alexmatson-da: made the driver consistently return base64 for all sigs & public key values

mjuchli-da (approved):

PR #2398, opened 2026-09-03
feat: handle locks using relative expiresAfter

### LINKED ISSUE #2388: Expired locked holdings should be available for transfers
> Locked holdings with expired locks ([code](https://github.com/canton-network/splice/blob/3418ac7be124d75d58f79801fefbd1e7480ccb99/token-standard/splice-api-token-holding-v1/daml/Splice/Api/Token/HoldingV1.daml#L28-L35)) must be treated as unlocked holdings, i.e., they should be included in the available total supply and they should be used to fund transfers and other actions that require input holdings. Doing so ensures that fully vested holdings can always be used as funding as soon as they vest.

pascalhonegger-da:
Closes #2388

### DIFF
--- a/core/token-standard-service/src/token-standard-service.ts
+++ b/core/token-standard-service/src/token-standard-service.ts
@@ -34,7 +34,6 @@ import {
     PrettyContract,
     renderTransaction,
     ViewValue,
-    Holding as TxParseHolding,
     PrettyTransactions,
     Transaction,
     TransferObject,
@@ -1723,7 +1722,7 @@ export class TokenStandardService {
     }
 
     static isHoldingLocked(
-        holding: Holding | TxParseHolding,
+        holding: HoldingView,
         currentTime: Date = new Date()
     ): boolean {
         const lock = holding.lock
--- a/core/token-standard-service/src/token-standard-service.ts
+++ b/core/token-standard-service/src/token-standard-service.ts
@@ -125,17 +125,9 @@ export class CoreService {
             )
         }
 
-        const unlockedSenderHoldings = senderHoldings.filter((utxo) => {
-            //filter out locked holdings
-            const lock = utxo.interfaceViewValue.lock
-            if (!lock) return true
-
-            const expiresAt = lock.expiresAt
-            if (!expiresAt) return false
-
-            const expiresAtDate = new Date(expiresAt)
-            return expiresAtDate <= now
-        })
+        const unlockedSenderHoldings = senderHoldings.filter(
+            (utxo) => !TokenStandardService.isHoldingLocked(utxo, now)
+        )
 
         if (unlockedSenderHoldings.length > 100) {
             this.logger.warn(`Sender has more than 100 unlocked utxos.`)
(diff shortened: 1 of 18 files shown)


core/tx-parser/src/types.ts
```diff
@@ -59,7 +60,9 @@ export interface Holding {
 export interface HoldingLock {
     holders: string[]
     expiresAt?: string
-    expiresAfter?: string
+    expiresAfter?: {
+        microseconds: string
+    }
```
pascalhonegger-da: I was very surprised that this did not break any type checks, given the type in daml I believe this is more correct

core/tx-parser/src/parser.ts
```diff
@@ -839,6 +839,7 @@ export class TransactionParser {
                     amount: holdingView.amount,
                     instrumentId: holdingView.instrumentId,
                     contractId: exercisedEvent.contractId,
+                    contractCreatedAt: holdingView.contractCreatedAt,
```
fayi-da: Hmmm 🤔

mjuchli-da: what's the verdict here?

fayi-da: Puilled the branch to do some digging, i think we should be getting the timestamp from the event, not the holding 🤔

core/tx-parser/src/types.ts
```diff
@@ -49,6 +49,7 @@ export interface TokenStandardEvent {
 // Same definition as HoldingView in Daml
 export interface Holding {
     contractId: string
+    contractCreatedAt: string
```
fayi-da: This doesn't exist on the holdingView daml interface

https://github.com/canton-network/wallet/blob/de09ef1def61fef97b58de80597432c09dda7415/damljs/token-standard-models/daml/Splice/Api/Token/HoldingV1.daml#L49-L64

pascalhonegger-da: Neither does `contractId` which is why the type is redefined. I was also surprised of the redefined type here, we should discuss on where we get the timestamp from - but if we don't add it to the holding we'll have to extend a few helper classes with a new argument which will impact the public API

mjuchli-da: Isn't this just wrong and we should define these properties as part of `PrettyContract` (same file) which maps metadata (added via the lapi)? `Holding` interfaces and the like should map the Daml.

pascalhonegger-da: I've created some commits which explore removing the custom `Holding` interface altogether and directly referencing the generated types, still not quite finished but if you want to provide feedback on the direction this is going I'm happy to work on it Wednesday

fayi-da: Brave for the refactor but I actually like this :heart_hands: . The only downside is the API change for the parser and any wallet-sdk methods(e.g ` token.holdings`) that previously returned `Holding` now returns `PrettyContract<Holding>`

pascalhonegger-da: I did some cleanup and tried to minimize the breaking API changes, but actually there are some cases where the current type has a required `contractId` which is never set 😱 

The current diff really tries to keep the pain for the end users to a minimum, the breaking API changes are `isHoldingLocked` (needs `PrettyContract<HoldingView>` instead of `Holding`) and the result type for `UtxoNamespace.list` (returns `HoldingView` instead of `Holding` - never returned `contractId` even though specified in TypeScript!)

core/token-standard-service/src/token-standard-service.ts
```diff
@@ -1723,16 +1714,46 @@ export class TokenStandardService {
...
+        if (lock.expiresAfter) {
+            const createdAt = new Date(
+                holding.activeContract.createdEvent.createdAt
+            )
+
+            // 1 microsecond = 0.001 milliseconds
+            const msToAdd = parseInt(lock.expiresAfter.microseconds) / 1000
+
+            expiresAtRelative = new Date(createdAt.getTime() + msToAdd)
+        }
```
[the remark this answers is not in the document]
pascalhonegger-da: Good point — sub-ms precision loss is real, but it’s a general JS Date limitation and not unique to this API. For expiresAfter, from what I understand expected usage is on the order of seconds/minutes, so preserving sub-ms precision doesn’t provide practical value. I’d favor keeping this simple and predictable for now.

core/tx-parser/src/types.ts
```diff
@@ -46,22 +46,10 @@ export interface TokenStandardEvent {
...
+export type Holding = HoldingView & { contractId: string }
 
-export interface HoldingLock {
-    holders: string[]
-    expiresAt?: string
-    expiresAfter?: string
-    context?: string
-}
+/** @deprecated use Lock from @canton-network/core-token-standard directly */
+export type HoldingLock = Required<Holding['lock']>
```
rukmini-basu-da: you can remove this (typically we only add the deprecated to the sdks rather than the core packages)

pascalhonegger-da: Done 🚀

rukmini-basu-da (approved):

PR #2408, opened 2026-09-07
fix: multisync - review fixes

jarekr-da:


canton/multi-sync/app-synchronizer.sc
mateuszpiatkowski-da, line 6:
```suggestion
// survives a `pnpm script:fetch:localnet` (the .localnet bundle is gitignored).
```

core/splice-codegen/rollup.config.js
mateuszpiatkowski-da, line 73: Do we foresee a possibility to have multiple versions included inside `damljs` directory (e.g. for backwards compatibility)? If so, then we'll need a stricter filtering than `startsWith` since it will pick only the first occurrence and apply it without any further means of control.

core/token-standard-service/src/token-standard-service.ts
mateuszpiatkowski-da: good catch! 🥳

docs/wallet-integration-guide/examples/scripts/17-multi-sync/_amulet_ops.ts
mateuszpiatkowski-da, line 1: I'm not sure I understand this diff... comparing to what is on main this seems identical, yet for some reason GH recognizes these files as different/added. Do you know what is the reason behind that? Have you modified these files?

Same goes for the rest of the files in the dir

mateuszpiatkowski-da: Oh I know now. You need to rebase on top of main since you have conflicts :D

mateuszpiatkowski-da (commented):
@jarekr-da I have reviewed a portion of the PR since most of it is a repetition of what is inside main rn. I'll review the rest of it once it's rebased

PR #2416, opened 2026-09-09
feat: synchronizer must be set explicitly

jarekr-da:
Goal - remove logic that chooses synchronizerId based on name (like `global`).

SynchronizerId is now explicitly selected in sdkContext.

if not provided the code will check if there is only one synchronizer available and use it
otherwise it will fail.

This is alignment with https://docs.google.com/document/d/1Dtr_X-AOg_evnHkZynSNBUOQD4z5hzqmN48gw9tWvAg/edit?usp=sharing

> Key Guideline: Use automatic reassignments, but explicitly choose the target synchronizer for each workflow step. 

Notice:  many operations have its own synchronizerId argument (this is unchanged)

multiple tests changed (trivial changes to adapt to this logic)
we still need to select synchronizerId by name in setup of tests


core/ledger-client/src/ledger-client.ts
```diff
@@ -521,32 +528,35 @@ export class LedgerClient {
...
+            throw new Error(
+                `Participant is connected to ${synchronizers.length} synchronizers ` +
+                    `(${synchronizers.map((s) => s.synchronizerId).join(', ')}); ` +
+                    'an explicit synchronizerId is required.'
             )
         }
-        this.synchronizerId = synchronizerId
-        return synchronizerId
+
+        this.synchronizerId = synchronizers[0].synchronizerId
```
jarekr-da: this logic is problematic if we dynamically connect - disconnect synchronizers
we might end with a random  synchonizerId cached

the current logic had similar problem

but I do not see much better solution as of now (maybe we should not cache synchronizer at all (if not specified)

PR #2418, opened 2026-09-09
chore: migrate to PNPM 12

pascalhonegger-da:
- **chore: update to latest node 24 patch**
- **chore: update to PNPM 12**

PNPM 12 should give even faster installs thanks to their rust backend, but there are no significant user facing differences as far as I can tell.


alexmatson-da (approved):
Thanks!

PR #2426, opened 2026-09-10
feat: introduce core-wallet-services

alexmatson-da:
This PR introduces the `core-wallet-services` package for code reused between the extension & remote backends. It brings `transaction-service` into the package


core/wallet-services/src/utils.ts
```diff
@@ -0,0 +1,78 @@
...
+    commandId?: string
+    commands?: Commands
+    actAs?: string[]
+    readAs?: string[]
+    disclosedContracts?: DisclosedContracts
+    packageIdSelectionPreference?: PackageIdSelectionPreference
+}
+
+export function ledgerPrepareParams(
+    userId: string,
```
mateuszpiatkowski-da, line 24: nitpick - consider changing to single record for verbosity (calling a function with object is more clear and readable than using multiple args, as user would see keys next to values)

core/wallet-services/package.json
```diff
@@ -0,0 +1,62 @@
+{
+    "name": "@canton-network/core-wallet-services",
```
mateuszpiatkowski-da, line 2:
```suggestion
    "name": "@canton-network/core-wallet-gateway-services",
```

I think that the renaming is clearer as it points the purpose of the package (wallet-services sounds ambiguous)

mateuszpiatkowski-da (approved):
LGTM, left a few nitpicks, nothing groundbreaking :)

core/wallet-services/package.json
alexmatson-da: I think I prefer the looser naming, `wallet-gateway` is more of a product-facing name and would rather not put it in the package name if unnecessary

core/wallet-services/src/utils.ts
alexmatson-da: agreed

pawelstepien-da (dismissed):
Changes look good! 2 questions:
1. Was such big lock update intentional?
2. Would it make sense to add wallet-services to start:all so we have hot-reload?

pawelstepien-da (approved):

PR #2427, opened 2026-09-10
fix(tx-visualizer): extract amount nested under Mint choice argument

davidgayman-da:
RegistrarHoldingOperationsDelegation_Mint and similar Mint choices nest the amount under chosenValue.mint.amount rather than a top-level chosenValue.amount or chosenValue.allocation.transferLeg.amount, so the Wallet Gateway mint approval popup showed N/A for the amount.

Adds a fallback lookup path for chosenValue.mint.amount and a regression test covering the Mint choice argument shape.


wallet-gateway/remote/Dockerfile
wallacekelly-da, line 1: 🤔 Is this Dockerfile needed? If so, perhaps it should be in a separate PR since it's not related to extracting amount.

davidgayman-da: Yes that's a good idea. Moving to a separate branch.

core/tx-visualizer/src/index.ts
```diff
@@ -237,6 +259,10 @@ function extractChoiceIdAndAmount(obj: any) {
                 ),
                 'amount'
             )
+        ) ??
+        findFirstNumericField(
+            getFieldValue(exercise?.chosenValue, 'mint'),
```
rukmini-basu-da, line 264: Hardcoded path per choice doesn't scale (every new choice adds another `??` branch) so we would want to create a more generic solution to transaction parsing

rukmini-basu-da:
Transaction visualization improvements will be part of https://github.com/canton-foundation/canton-dev-fund/issues/365). Notably, standard token transfer transactions return no amount and the `getFirstNodeOfType` is also an issue as it relies on the nodes array order rather than the execution order. We're suggesting removing the amount from the transaction view until parsing becomes more reliable

mjuchli-da:
Thanks for the contribution. As mentioned by @rukmini-basu-da , we feel it's best to remove the `amount` field entirely until a more stable concept for parsing tx's has been introduced. I'm therefore closing this PR and we will open a follow-up PR for the removal.

PR #2443, opened 2026-09-11
fix(wallet-sdk): fix array based plugin registration

pascalhonegger-da:
- **chore: add cors to test-token-v1-registry**
- **Remove unnecessary casts**
- **fix: bad plugin name if registering plugins[]**
- **chore: add utilities endpoint to test registry**


examples/portfolio/src/hooks/query-options.ts
```diff
@@ -216,7 +217,9 @@ export const preapprovalStatusQueryOptions = ({
                 })
             )
 
-            return await sdk.utilities.preapprovalTransfer.fetchQuick({
+            return await sdk[
+                WalletSDKUtilitiesPluginName
```
rukmini-basu-da: as discussed offline, it would be good to keep `sdk.utilities.preapprovalTransfer.fetchQuick` to be more consistent with how we access other namespaces. an idea would be to add something like this to the `plugin.ts`

```
export abstract class SDKPlugin<
    ExtendedNamespaceItems extends keyof ExtendedFullSDKInterface = never,
    Name extends string = string,
> {
    /**
     *
     * @deprecated use this.ctx.logger instead
     */
    protected readonly logger: ReturnType<SDKLogger['child']>
    protected readonly ctx: SDKPluginContext<ExtendedNamespaceItems>

    constructor(
        public readonly name: Name,
        protected readonly _ctx: SDKPluginContext<ExtendedNamespaceItems>
    )
```

pascalhonegger-da: Needed some more minor adjustments, the current version was not type safe at all and allowed all string keys to be used on the sdk 😨 

It's now type safe, but requires the use of `extends SDKPlugin<'myPlugin'>` for using array based plugin registration, I don't believe it's possible to infer this type from the `super('myPlugin')` call unfortunately.

rukmini-basu-da (approved):
thanks!

PR #2473, opened 2026-09-18
test: vitest 5 and minor dependency upgrades

pascalhonegger-da:
- **chore: upgrade to vitest 5**
- **chore: upgrade playwright to 1.63.0**
- **chore: upgrade eslint and plugins to latest**


alexmatson-da (approved):
nice!

PR #2475, opened 2026-09-18
feat: self-issued idp and auth schemas and username input at login

### LINKED ISSUE #2457: Extend WG auth for `self_issued`
Image

Extend `idp.type` (and implicitly `network.auth.method`) with `self_issued`. 
If this method is defined on the selected network, we present the above alternative login method via username.

pawelstepien-da:


core/wallet-auth/src/config/schema.test.ts
```diff
@@ -14,6 +14,14 @@ describe('schemas', () => {
 
         expect(idpSchema.safeParse(validIdpSelfSigned).success).toBe(true)
 
+        const validIdpSelfIssued = {
+            id: 'test1',
+            type: 'self_issued',
+            issuer: 'party-jwt',
```
mjuchli-da: This seems off. Issuer is determined upon token creation (depending on the party in use), not on build time (idp config).

core/wallet-auth/src/config/schema.ts
```diff
@@ -53,26 +59,34 @@ export const authSchema = z.discriminatedUnion('method', [
...
 export const idpSchema = z.discriminatedUnion('type', [
     z.object({
         id: z.string(),
         type: z.literal('self_signed'),
         issuer: z.string(),
     }),
+    z.object({
+        id: z.string(),
+        type: z.literal('self_issued'),
+        issuer: z.string(),
```
mjuchli-da: See comment above.

core/wallet-store-sql/src/schema.test.ts
```diff
@@ -41,6 +41,12 @@ const selfSignedIdp: Idp = {
     issuer: 'unsafe-auth',
 }
 
+const selfIssuedIdp: Idp = {
+    id: 'idp-self-issued',
+    type: 'self_issued',
+    issuer: 'party-jwt',
```
mjuchli-da: See other comments.

docs/wallet-gateway/configuration/schema.md
```diff
@@ -400,6 +400,23 @@ npx @canton-network/wallet-gateway-remote@latest --config-schema
...
+                                "type": "object",
+                                "properties": {
+                                    "id": {
+                                        "type": "string"
+                                    },
+                                    "type": {
+                                        "type": "string",
+                                        "const": "self_issued"
+                                    },
+                                    "issuer": {
```
mjuchli-da: See other comments.

wallet-gateway/test/config.json
```diff
@@ -46,6 +46,11 @@
                 "id": "idp-self-signed",
                 "type": "self_signed",
                 "issuer": "unsafe-auth"
+            },
+            {
+                "id": "idp-self-issued",
+                "type": "self_issued",
+                "issuer": ""
```
mjuchli-da: See other comments.

mjuchli-da (commented):
The only concern is the `issuer` property, which I thought we didn't need in the idp config. Otherwise LGTM.

core/wallet-auth/src/config/schema.ts
```diff
@@ -17,6 +17,10 @@ const authorizationCodeAuthSchema = z
 
 const clientCredentialsAuthSchema = z.object({
     method: z.literal('client_credentials'),
+    identityProviderId: z.string().optional().meta({
+        description:
+            "Overrides the network's identity provider for client credentials token acquisition.",
+    }),
```
alexmatson-da, line 23: I'm a bit confused about the purpose of this. What situations would this field be used?

pawelstepien-da: When using IDP.type = 'self_issued' we don't have IDP.configUrl. If adminAuth of correlated network has method = 'client_credentials', it requires configUrl to create an accessToken. That configUrl is taken from the IDP. This override allows using a different IDP for regular auth (self_issued) and another IDP for adminAuth (oauth).

api-specs/openrpc-user-api.json
```diff
@@ -1447,7 +1573,54 @@
...
+                    "audience": {
+                        "title": "audience",
+                        "type": "string"
+                    }
+                },
+                "required": ["method", "audience", "scope"]
+            },
+            "Auth": {
+                "title": "Auth",
+                "description": "Authentication configuration for a network",
```
mjuchli-da: Nice. In the future, the WG can also make use of this for it's auth barrier.

core/wallet-store-sql/src/migrations/017-make-idp-issuer-nullable.ts
```diff
@@ -0,0 +1,90 @@
...
+            .alterTable('idps')
+            .alterColumn('issuer', (column) => column.dropNotNull())
+            .execute()
+    } else {
+        // `networks.identity_provider_id` references `idps` with ON DELETE CASCADE.
+        // Temporarily disable enforcement while replacing the referenced table,
+        // so dropping it does not delete the dependent network rows.
+        await sql`PRAGMA foreign_keys = OFF`.execute(db)
+        try {
+            await db.transaction().execute(async (trx) => {
```
mjuchli-da: Why do we need to create a new table, copy and the rename?
Don't we just want to make the column nullable? https://dba.stackexchange.com/questions/159793/how-to-set-a-not-null-column-into-null-in-postgresql

mjuchli-da (approved):
LGTM! One remark regarding the migration. Also, I'm not sure if we make things too complicated with overriding values here https://github.com/canton-network/wallet/pull/2475#discussion_r4080100192

core/wallet-store-sql/src/migrations/017-make-idp-issuer-nullable.ts
pawelstepien-da: postgres is in if block above, else is sqlite. I just found out that capability to drop NOT NULL in sqlite was recently added

https://www.sqlite.org/lang_altertable.html#:~:text=The%20ability%20to%20set%20or%20drop%20NOT%20NULL%20constraints%20from%20a%20column%20using%20the%20ALTER%20TABLE%20ALTER%20COLUMN%20syntax%20was%20added%20in%20SQLite%203.53.0%20(2026%2D04%2D09).

> The ability to set or drop NOT NULL constraints from a column using the ALTER TABLE ALTER COLUMN syntax was added in SQLite 3.53.0 (2026-04-09).

But I don't think we should rely on such new SQLite features. Makes me think whether we should declare something like minimum sqlite and postgres version for wallet gateway.

mjuchli-da: Ah I've misread the if/else. Makes sense to keep the copying for now until the feature matures.

